PR #8, opened 2025-04-03
fix cell output for insert code cell tool

eleonorecharles:


echarles:
@eleonorecharles Will be good to also bump version and release in pypi + dockerhub

Issue #10, opened 2025-04-10
Review of jupyter-mcp-server-extended


eleonorecharles:
Hi @itisaevalex,

We've been testing the recent changes and additions of [jupyter-mcp-extended](https://github.com/itisaevalex/jupyter-mcp-extended). Here are our main takeaways:

What works well
- Some of the fixes and new tools are great improvements — they bring valuable new capabilities and polish.

What concerns us
- The number of available tools has grown significantly (from 2 to 18), which in our experience makes the AI's behavior less predictable and sometimes harder to control.
- A key limitation we've observed is that the add_cell tool does not automatically trigger a code cell execution. As a result, the AI doesn't get immediate feedback from the code outputs after using a tool. This weakens the feedback loop and makes it harder for the AI to learn from or react to the latest context.

echarles:
Closing. Please reopen if you'd like to further discuss.

PR #22, opened 2025-06-08
Add new endpoints for managing existing cells

jamesmcm:
Adds the following MCP tool endpoints:
- overwrite_cell_source(cell_index: int, cell_source: str) -> str
- execute_cell(cell_index: int) -> list[str]
- read_all_cells() -> list[dict[str, Union[str, int, list[str]] ]]
- read_cell(cell_index: int) -> dict[str, Union[str, int, list[ str]]]
- get_notebook_info() -> dict[str, Union[str, int, dict[str, int]]]
- delete_cell(cell_index: int)

This server is awesome btw I was able to watch the agent as it went one-by-one modifying the cells in the notebook!

I think the only thing still missing is clearing outputs.

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -4,6 +4,7 @@
 
 import logging
 import os
+from typing import Union
 
 from jupyter_kernel_client import KernelClient
 from jupyter_nbmodel_client import (
@@ -15,11 +16,11 @@ from mcp.server.fastmcp import FastMCP
 mcp = FastMCP("jupyter")
 
 
-NOTEBOOK_PATH = os.getenv("NOTEBOOK_PATH", "notebook.ipynb")
+NOTEBOOK_PATH: str = os.getenv("NOTEBOOK_PATH", "notebook.ipynb")
 
-SERVER_URL = os.getenv("SERVER_URL", "http://localhost:8888")
+SERVER_URL: str = os.getenv("SERVER_URL", "http://localhost:8888")
 
-TOKEN = os.getenv("TOKEN", "MY_TOKEN")
+TOKEN: str = os.getenv("TOKEN", "MY_TOKEN")
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


echarles:
Thank you so much @jamesmcm for this contribution. Give us a few days to review and comment, we are pretty packed atm... I guess to additions have been working for you, right?

jamesmcm:
No problem, yeah exactly I used it with with Cursor Chat and it worked perfectly.

We'd probably want another endpoint to insert a new cell in a specific position too though.

echarles:
> We'd probably want another endpoint to insert a new cell in a specific position too though.

Yep, that one is much wanted - Feel free to push that one as part of this PR.

echarles:
Something to consider is the impact on the LLM of that growing number of tools, see comment here https://github.com/datalayer/jupyter-mcp-server/issues/10#issue-2984966011

@jamesmcm Have you seen a change in the user experience when adding those tools?

jamesmcm:
It was fine for me, but I disabled the add_cell tools in Cursor as I knew I didn't need them (I had a specific task of filling out an existing notebook).

While having too many tools is definitely an issue for the LLMs to manage, I think it'd be better for users to handle this in their client (e.g. disabling the tools they don't need in Cursor), and have the MCP server focus on providing a consistent LLM-friendly API across all the tools (e.g. naming, description, argument and return types). Users should already be doing this between MCP servers anyway (only turning on the ones you need).

It'd be nice if clients also had an option for the LLM to first review the task and then disable tools it thinks it won't need, so it wouldn't need to be done manually for every task.

echarles:
> e.g. disabling the tools they don't need in Cursor

Can work, that is pushing a bit more work to the user... and also specific to Cursor vs other clients.

> It'd be nice if clients also had an option for the LLM to first review the task and then disable tools it thinks it won't need,

Again, highly depend on the client.

I wonder if we could have on the Jupyter MCP Server side a base set of tools, and argument on start to add optional ones.

eleonorecharles:
> I wonder if we could have on the Jupyter MCP Server side a base set of tools, and argument on start to add optional ones.

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/278#issuecomment-2944634453 is suggesting a similar approach: specifying the tools to use at the server level.

eleonorecharles:
Thanks a lot @jamesmcm for this PR!
I have reviewed the changes, they look good to me. We’ll just need to rename after merging `cell_content` to `cell_source` to stay consistent across all tools — and I agree that `cell_source`, as you used, makes more sense.
Before merging, do you want to add the `insert_execute_code_cell` and `insert_markdown_cell` tools with an index ?

jamesmcm:
Thanks, yeah sure, and we can look at the disabling tools issue, verifying the cell type for writing in `overwrite_cell_source`, and fixing the CI builds in other PRs.

eleonorecharles:
Great, thank you so much @jamesmcm! I will do few more tests today and push the docker image.

PR #37, opened 2025-08-04
Improve docker build

romainx:
Hello,

I made some simple improvements to the docker build in order to improve images footprints and security.

- Fix [Hadolint](https://github.com/hadolint/hadolint) findings to produce lighter images:
  - [`DL3042`](https://github.com/hadolint/hadolint/wiki/DL3042): Avoid to use pip cache when installing Python packages.
  - [`DL3059`](https://github.com/hadolint/hadolint/wiki/DL3059): Stack all pip installation layers into a single layer.
- Group `COPY` instructions: Stack single file copy layers into a single layer.
- Add a `.dockerignore`: Since the `Dockerfile` is located at the root of the repository (which is not recommended), all repository contents are sent to the daemon as part of the build context. It is important to reduce the size of the build context for performance reason, but more importantly for security reason to avoid inadvertently sending files containing sensitive information to the daemon or adding them to the image using the `ADD` or `COPY` commands.

Hope it helps.


echarles (approved):
LGTM Thx @romainx

PR #47, opened 2025-08-08
Refactoring of the read cell / cell info model

romainx:
Hello,

I made a careful (I hope 😄 ) refactoring of the extract of the cell info data that is returned by the tools `read_cell` and `read_all_cells`. I did not change the interfaces (tests are passing before and after the change) but I created a model `CellInfo` model that can instantiate itself from index and notebook cell info. I have also enhanced the tests to include `read_all_cells` in their logic.

This permits to get more structured data from the external interface (model validation) and to regroup logic (and avoid code duplication) related to data extraction.

Cheers 🍰

### DIFF
--- a/jupyter_mcp_server/models.py
+++ b/jupyter_mcp_server/models.py
@@ -1,8 +1,9 @@
 # Copyright (c) 2023-2024 Datalayer, Inc.
 #
 # BSD 3-Clause License
-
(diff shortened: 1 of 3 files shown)


eleonorecharles:
Great! Thank you @romainx.

Issue #89, opened 2025-09-30
A Amazing Discovery: Adding the `execute_ipython` Tool


ChengJiale150:
## 📋 Abstract

This issue proposes adding the `execute_ipython` tool to Jupyter MCP Server, empowering LLMs with the ability to interact directly with IPython kernels. By leveraging IPython's built-in shell command support, magic commands, and Python code execution capabilities, this tool will significantly expand the functional boundaries of Jupyter MCP tools, enabling true freedom to explore and master the Jupyter server environment.

### Inspiration from the Shell Tool

Current AI programming practices (such as Claude Code, Cursor, etc.) demonstrate that the `Shell` tool is an exceptionally powerful tool, with its strength lying in:

- **🔧 Direct System Access**: Execute system commands and operate on the file system directly
- **📦 Package Management**: Install and manage Python packages and system dependencies
- **🔍 Environment Exploration**: View system information, environment variables, file structure, etc.
- **⚡ Flexibility**: Can accomplish nearly any task achievable via command line

However, in the Jupyter environment, we have an even more powerful interactive environment than Shell—**IPython**, which not only encompasses all Shell functionalities but also provides rich features specifically designed for scientific computing and data analysis.

### Core Concept

Inspired by the Shell tool, we propose building the `execute_ipython` tool to enable LLMs to interact directly with IPython kernels. This is an exceptionally powerful tool that will:

- ✅ Support **IPython Magic Commands**
- ✅ Support **Shell Commands** (via `!` prefix)
- ✅ Support **Direct Python Code Execution**
- ✅ Enable **Interaction with Remote Jupyter Servers**
- ✅ Provide **Instant Feedback and Debugging Capabilities**

#### 1️⃣ Magic Commands

IPython's magic commands are one of its most powerful features, providing rich functionality enhancements.

**Typical Use Cases:**

##### Performance Analysis
```python
# Measure code execution time
%timeit sum(range(1000))

# Detailed performance profiling
%prun -s cumulative my_function()

# Line-by-line performance profiling
%lprun -f my_function my_function()
```

##### Code and Environment Management
```python
# List all variables
%who

# List variables of specific type
%who_ls str

# View detailed variable information
%whos

# View all history records related to pandas
%history -g "pd."
```

##### Code Loading and Execution
```python
# Load code from file
%load script.py

# Run external Python script
%run analysis.py

# Run script and enter debug mode
%run -d script.py

# Auto-reload modules
%load_ext autoreload
[...]

PR #90, opened 2025-09-30
Feature/execute ipython

ChengJiale150:
See in #89

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-    branches: ["main","feature/execute-ipython"]
+    branches: ["main"]
   pull_request:
 
 defaults:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-    branches: ["main"]
+    branches: ["main","feature/execute-ipython"]
   pull_request:
 
 defaults:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,9 @@ name: Test
 
 on:
   push:
-    branches: ["main","feature/execute-ipython"]
+    branches: ["main"]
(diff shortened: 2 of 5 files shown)


ChengJiale150:
Currently, it's implemented by adding a temporary Code Cell at the end of the Notebook, executing it to get the output, and then deleting it. This is not elegant, and it will be improved in the future

echarles:
> Currently, it's implemented by adding a temporary Code Cell at the end of the Notebook, executing it to get the output, and then deleting it.

I think this step can be skipped using the kernel-client api directly.

ChengJiale150:
> > Currently, it's implemented by adding a temporary Code Cell at the end of the Notebook, executing it to get the output, and then deleting it.
> 
> I think this step can be skipped using the kernel-client api directly.

Sounds Good to me. fix it using kernel api in `notebook.execute_cell`

echarles:
On https://github.com/datalayer/jupyter-kernel-client you can use simply the following to run code (not need for asyncio, tasks...)

```
import os

from platform import node
from jupyter_kernel_client import KernelClient

with KernelClient(server_url="http://localhost:8888", token="MY_TOKEN") as kernel:
    code = """import os
from platform import node
print(f"Hey {os.environ.get('USER', 'John Smith')} from {node()}.")
"""
    reply = kernel.execute(code)
    print(reply)
    assert reply["execution_count"] == 1
    assert reply["outputs"] == [
        {
            "output_type": "stream",
            "name": "stdout",
            "text": f"Hey {os.environ.get('USER', 'John Smith')} from {node()}.\n",
        }
    ]
    assert reply["status"] == "ok"
```

ChengJiale150:
How can I control  the timeout without asyncio? The `timeout` in `kernel.execute` do not represent for the timeout which I want.

echarles:
OK, I see. timeout for code execution is sometimes tricky.

Happy to merge as-is. Later, it would be great to factor better timeout control in the kernel-client library.

PR #110, opened 2025-10-12
Fix: Unify remote file paths for cross-platform compatibility

ChengJiale150:
This change addresses an bug where file paths could be inconsistent when connecting from a Windows client to a Jupyter Lab on a Linux host server on remote connection.

The primary modification is in the 'use_notebook_tool.py' file, where 'path.parent' is now explicitly converted to a POSIX-standard path using '.as_posix()'. This ensures that the directory paths sent to the Jupyter server use forward slashes ('/'), regardless of the client's operating system.

### DIFF
--- a/jupyter_mcp_server/__version__.py
+++ b/jupyter_mcp_server/__version__.py
@@ -4,4 +4,4 @@
 
 """Jupyter MCP Server."""
 
-__version__ = "0.15.3"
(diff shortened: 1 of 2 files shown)


echarles:
@ChengJiale150 jupyter-mcp-server 0.15.4 released in pypi

Issue #132, opened 2025-10-18
Rename `execute_ipython` to `execute_code`


echarles:
The `execute_ipython` is not specific to python, so we should rename it into a more genreric `execute_code` (or equivalent).

ChengJiale150:
@echarles Already in my latest roadmap in #127

echarles:
Awesome, let's leave this open for visibility and close when implemented.

echarles:
Closing as implemented in #134

Issue #127, opened 2025-10-17
🛣️ My Jupyter MCP Server Tools Improvement Roadmap


ChengJiale150:
## 📋 Overview

Currently, the Jupyter MCP Server tool implementation is basically complete, covering core functionalities of Server Management, Notebook Management, and Cell Operations. However, to truly enhance user experience and enable LLMs to better utilize these tools, we need to make in-depth improvements in the following three main areas:

1. **Tool Description and Input Parameter Optimization** - Making it easier for LLMs to understand tool purposes and usage methods
2. **Tool Output and Error Message Enhancement** - Providing more user-friendly and structured output information
3. **Potential Bug Fixes** - Identifying and fixing existing boundary issues and exceptions

---

### ✅ Completed Work

- ✨ Complete core tool functionality (16 tools)
- 🔄 Support for multi-notebook management
- 📊 Support for multimodal output (text, images, etc.)

### ⚠️ Areas Needing Improvement

- 📝 **Insufficient Tool Descriptions**: Current tool descriptions are relatively brief, making it difficult for LLMs to understand applicable scenarios and parameter meanings
- 🔧 **Incomplete Parameter Documentation**: Input parameter constraints, optional values, boundary conditions, and other information are not clearly documented
- 📤 **Inconsistent Output Formats**: Return values from different tools may not be LLM-friendly enough
- 🚨 **Unfriendly Error Handling**: Error messages are not specific enough to help LLMs quickly locate problems
- 🐛 **Insufficient Boundary Case Handling**: Some exceptional situations (such as network errors, timeouts, resource shortages, etc.) lack comprehensive handling mechanisms

---

### 📌 Phase 1: Server Management Tools Enhancement

**Target Tools**:
- `list_files` - File system listing
- `list_kernels` - Kernel listing
- `assign_kernel_to_notebook` - Kernel assignment

#### Improvement Content

- [x] Supplement detailed descriptions, usage scenarios, and precautions for each tool
- [x] Add pagination support to `list_files` tool to avoid returning too many tokens at once
- [x] Add result filtering options to `list_files` tool
- [x] remove `assign_kernel_to_notebook` 
---

### 📌 Phase 2: Notebook Management Tools Enhancement

**Target Tools**:
- `use_notebook` - Switch/Create Notebook
- `list_notebooks` - Notebook listing
- `restart_notebook` - Restart Kernel
- `unuse_notebook` - Disconnect

#### Improvement Content

[...]

PR #134, opened 2025-10-20
rename `execute_ipython` to `execute_code`

ChengJiale150:
Complete #132 and one of the part of #127


echarles:
Thx  @ChengJiale150 Could you also update the doc?

ChengJiale150:
I prefer to update the doc until all changes in #127 are applied. And of course, I can update the doc now, howerver it may cause confusion to user due to different tool name in current version.

Issue #146, opened 2025-10-24
UI edits stop syncing after Claude Code makes MCP edits


delphos-mike:
### Environment
- `jupyter-collaboration`: 4.1.1
- `jupyter-mcp-server`: 0.18.0
- `JupyterLab`: 4.4.1
- `jupyter-server-ydoc`: 2.1.1
- OS: macOS (Darwin 25.0.0)
- Claude Code using MCP to interact with notebooks

### Problem Description

When using JupyterLab with jupyter-collaboration and Claude Code editing notebooks via MCP, user edits made directly in the JupyterLab UI stop persisting after Claude makes edits. The workflow breaks down as follows:

1. **Initial state**: User can edit in UI, changes persist ✅
2. **Claude makes an edit via MCP**: Edit syncs correctly ✅
3. **User makes subsequent edits in UI**: Edits appear to work (cells execute, UI updates) but DON'T persist ❌
4. **After browser refresh**: User's changes are gone, only Claude's MCP edits remain

This creates a broken collaboration workflow where:
- ✅ Claude (via MCP) can successfully edit notebooks
- ❌ User edits in the UI disappear after Claude has made an edit
- One-way sync after MCP interaction

### Root Cause

After Claude Code makes an edit via MCP tools (e.g., \`overwrite_cell_source\`), the Y.js persistent SQLite store (\`.jupyter_ystore.db\`) appears to become out of sync. Server logs show:

\`\`\`
[I ServerApp] Content in room json:notebook:... loaded from the ystore SQLiteYStore
[I ServerApp] Content in file ... is out-of-sync with the ystore SQLiteYStore
[I ServerApp] Content in room json:notebook:... loaded from file ...
\`\`\`

After this point, UI changes update the browser's Y.js state but fail to propagate through the CRDT to the server's document store and eventually to disk. MCP edits continue to work correctly.

### Reproduction Steps

1. Start JupyterLab server with jupyter-collaboration enabled
2. Connect Claude Code to the notebook via MCP (\`use_notebook\`)
3. Have Claude make an edit (e.g., via \`overwrite_cell_source\`)
4. Verify Claude's edit appears in the UI ✅
5. Make a manual edit in the JupyterLab UI (e.g., change a variable name)
6. Execute the cell - edit appears successful, output reflects change
7. Check file on disk: \`grep "your_edit" notebook.ipynb\` - not found ❌
8. Refresh browser - only Claude's edit remains, user's edit is gone
[...]

PR #148, opened 2025-10-27
Fix: sync error

ChengJiale150:
fix the issue #146 

add thread lock and transaction to fix sync bug in `insert_cell` and `overwrite_cell` refer to the implement in `jupyter-nbmodel-client`.

It work well in my local test, and no sync problem occur.

### DIFF
--- a/jupyter_mcp_server/tools/delete_cell_tool.py
+++ b/jupyter_mcp_server/tools/delete_cell_tool.py
@@ -10,12 +10,12 @@ import nbformat
 from jupyter_server_client import JupyterServerClient
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 from jupyter_mcp_server.notebook_manager import NotebookManager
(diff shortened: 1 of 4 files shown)


ChengJiale150:
In Fact, this is as same as the code in `jupyter-nbmodel-client`

echarles:
> In Fact, this is as same as the code in jupyter-nbmodel-client

Instead of duplicating the code,could we just call jupyter-nbmodel-client or just not having a special case for JUPYTER_SERVER and do the same as for the MCP_SERVER case?

ChengJiale150:
It will be imporved in future depends on research of  the difference between serverapp and websocket.

echarles:
we should just make it simpler, and use the MCP_SERVER case when needed for JUPYTER_SERVER

ChengJiale150:
use same `NotebookModel` in `jupyter-nbmode-client` to simplify the api in JUPYTER_SERVER Mode.

echarles:
0.18.1 released with this fix https://github.com/datalayer/jupyter-mcp-server/releases/tag/v0.18.1

PR #177, opened 2025-11-27
add get_tool caching

eleonorecharles:


echarles:
Looks like a gain of speed also :+1:

PR #239, opened 2026-04-22
Fix test_jupyter_cite failing when both test modes run together

abbbe:
## Summary

- Switch `test_jupyter_cite` from `mcp_client_parametrized` to `mcp_client` fixture

## Why

`make test` (which sets both `TEST_MCP_SERVER=true` and `TEST_JUPYTER_SERVER=true`) always fails, while CI doesn't hit this because it runs `make test-mcp-server` and `make test-jupyter-server` separately.

`test_jupyter_cite` uses `mcp_client_parametrized`, which injects both `mcp_server` and `jupyter_extension` params. But prompts are only supported in MCP_SERVER mode — the `jupyter_extension` variant always fails with `Method not found: prompts/get`.

The module-level `pytestmark` skipif only guards against `TEST_MCP_SERVER=false`, it doesn't filter out the `jupyter_extension` parametrization. The fix is to use the `mcp_client` fixture, which only targets the standalone MCP server — matching the test's intent.

## Test plan

- [x] `make test` passes (142 passed, 0 failed)
- [x] Verified failure is pre-existing on unmodified code

### DIFF
--- a/tests/test_prompts.py
+++ b/tests/test_prompts.py
@@ -21,23 +21,23 @@ pytestmark = pytest.mark.skipif(
 
 @pytest.mark.asyncio
 @timeout_wrapper(60)
-async def test_jupyter_cite(mcp_client_parametrized: MCPClient):
+async def test_jupyter_cite(mcp_client: MCPClient):
     """Test jupyter cite prompt feature"""
-    async with mcp_client_parametrized:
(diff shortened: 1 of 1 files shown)


echarles (approved):
LGTM Thx @abbbe

PR #245, opened 2026-05-13
execute: Allow targeting a specific kernel

### LINKED ISSUE #244: Allow `execute_code` to target a kernel by id
### Problem

* Not all uses of Jupyter/kernels require having an active notebook (e.g. [here](https://github.com/emacs-jupyter/jupyter/blob/master/assets/workflow1.gif) )
* I would like to be able to introspect values in a kernel without an active notebook, and the tool API to do that seems missing.

### Proposed Solution

It seems that the codebase already has what's needed to do that, it's just about adding an argument to the exposed execute_code tool.

### Additional context

I'll open a PR.

nicolas-graves:
FIxes #244


echarles (approved):
LGTM Thx @nicolas-graves

Issue #183, opened 2025-12-18
Support XSRF-protected Jupyter deployments without Bearer tokens


EashanKaushik:
### Problem

The MCP server fails with `403 Forbidden` errors when connecting to Jupyter deployments that use XSRF protection but don't use Bearer token authentication.

**Error message:**
```
{"message": "'_xsrf' argument missing from POST", "reason": null}
```

This affects users in several common scenarios:

1. **Jupyter servers started without a token** (`--IdentityProvider.token=''`)
2. **Enterprise Jupyter deployments** with external authentication (SSO, OAuth, IAM)
3. **Managed Jupyter environments** (AWS SageMaker Studio, Google Colab Enterprise, Azure ML, etc.)
4. **JupyterHub deployments** where authentication is handled by the Hub, not the single-user server

In these environments, the collaboration API (`/api/collaboration/session/`) requires proper XSRF handling, but the current implementation only supports Bearer token authentication via the upstream `jupyter-nbmodel-client` library.

**Current behavior:**
- If `JUPYTER_TOKEN` is not provided, the server cannot authenticate to the collaboration API
- Users in token-less environments cannot use notebook collaboration features

### Proposed Solution

Add automatic XSRF cookie handling when no Bearer token is provided:

1. **Session-based authentication**: Make an initial GET request to obtain the `_xsrf` cookie from the Jupyter server
2. **XSRF header injection**: Include the `X-XSRFToken` header (matching the cookie value) in subsequent API requests
3. **Session caching**: Cache the authenticated session per server URL to avoid repeated cookie acquisition

**Implementation approach:**
- Create a custom `fetch_with_auth()` function that wraps HTTP requests with XSRF handling
- Modify `NotebookConnection` to use this function when no token is provided
- Fall back to standard Bearer token auth when `JUPYTER_TOKEN` is set (backward compatible)

**Example flow:**
```python
# 1. Get XSRF cookie
session = requests.Session()
session.get("http://jupyter-server:8888/")  # Sets _xsrf cookie

# 2. Make authenticated API call
xsrf = session.cookies.get('_xsrf')
session.put(
    "http://jupyter-server:8888/api/collaboration/session/notebook.ipynb",
    headers={"X-XSRFToken": xsrf},
    json={"format": "json", "type": "notebook"}
)
```

### Additional context

**Jupyter's XSRF protection:**
- Jupyter servers set an `_xsrf` cookie on first request
[...]

Issue #61, opened 2025-09-15
Connection to JupyterLab fails due to web socket authentication


Paul2708:
## Description

I'm running a JupyterHub to spawn JupyterLab instances. The spawned instance contains this MCP package.
I'm trying to connect to JupyterLab using the following configuration:

```json
{
  "mcpServers": {
    "jupyter": {
      "command": "docker",
      "args": [
        "run",
        "-i",
        "--rm",
        "-e",
        "DOCUMENT_URL",
        "-e",
        "DOCUMENT_TOKEN",
        "-e",
        "DOCUMENT_ID",
        "-e",
        "RUNTIME_URL",
        "-e",
        "RUNTIME_TOKEN",
        "datalayer/jupyter-mcp-server:latest"
      ],
      "env": {
        "DOCUMENT_URL": "https://<hub domain>/user/admin",
        "DOCUMENT_TOKEN": "6ed878dc2f2d8ba127ef4254f815d1b3",
        "DOCUMENT_ID": "notebook.ipynb",
        "RUNTIME_URL": "https://<hub domain>/user/admin",
        "RUNTIME_TOKEN": "6ed878dc2f2d8ba127ef4254f815d1b3"
      }
    }
  }
}
```

The token `6ed878dc2f2d8ba127ef4254f815d1b3` refers to the Jupyter server token. I verified that the token is correct by sending a request to the Jupyter server (e.g., `GET https://<hub domain>/user/admin/api/contents/notebook.ipynb`)

However, my MCP client (i.e., Claude Desktop) logs:

```
InvalidStatus: server rejected
WebSocket connection: HTTP 403
```

Similary, the JupyterLab environment logs:

```
[I 2025-09-15 14:20:04.766 ServerApp] Request for Y document 'notebook.ipynb' with room ID: 5a7c9985-1b5c-4255-a63d-b7b1d4ad281d
[I 2025-09-15 14:20:04.766 ServerApp] 200 PUT /user/ivv7386/api/collaboration/session/notebook.ipynb (ivv7386@::ffff:10.0.10.2) 20.93ms
[W 2025-09-15 14:20:04.867 ServerApp] Couldn't authenticate WebSocket connection
[...]

PR #250, opened 2026-05-18
feat: add password-based authentication support

EnyMan:
## Summary

Adds password-based authentication as an alternative to token auth, enabling the MCP server to work with Jupyter deployments that use password login and XSRF protection.

**Related issues:** Partially addresses #183 (XSRF-protected Jupyter deployments without Bearer tokens), related to #61 (WebSocket authentication failures)

### Changes

- New `JupyterPasswordAuth` class (`auth.py`) handles the `/login` flow: obtains XSRF cookie, POSTs credentials, verifies the session, and keeps a `requests.Session` alive so subsequent callers see live cookie values rather than a login-time snapshot.
- CLI options `--runtime-password`, `--document-password`, `--jupyter-password` (with corresponding env vars `RUNTIME_PASSWORD`, `DOCUMENT_PASSWORD`, `JUPYTER_PASSWORD`) follow the same priority/fallback pattern as existing token options. Password takes precedence over token when both are set for the same server.
- Auth cookies are injected into `JupyterServerClient`, `KernelClient`, and the collaboration WebSocket connection. The fresh `X-XSRFToken` header is read per-request from the live cookie jar, so cookie rotation is handled correctly.
- Two separate auth instances: `_runtime_password_auth` for kernel operations and `_document_password_auth` for collaboration. When `runtime_url == document_url` the runtime auth is reused; when URLs differ and `--document-password` is set, a second login is performed. A warning is logged when the document server is genuinely different and only `--runtime-password` is set.
- The collaboration WebSocket now uses `jupyter-nbmodel-client`'s native `additional_headers` (NbModelClient) and `headers` (`get_notebook_websocket_url`) parameters to inject the Cookie/XSRF auth — see "Removed monkey-patch" below.
- **Automatic re-login on session expiry**: Jupyter session cookies expire (30 days by default), so long-running sessions could go stale. `NotebookConnection` now catches `401/403` on the collaboration-session request, re-authenticates (`ServerContext.relogin_document()` → `JupyterPasswordAuth.relogin()`), re-injects the fresh cookies, and retries once.
- Refactored duplicated `_init_mcp_server_mode` logic in `server_context.py` into a single method with proper resource cleanup on partial failure. `ServerContext.reset()` now closes auth sessions to release connection pools.
- The CLI decorator was split into `_remote_connection_options` (the subset that `connect` can forward over `/api/connect`) and `_connection_options` (adds local-only password options used by `start`). This also fixed a pre-existing `TypeError` in `connect_command` on `main`.
- Comprehensive test suite (`test_auth.py`) covering the login flow (success, timeout, connection error at each stage, 5xx, 401/403, missing-xsrf, unexpected-status, response-body propagation), header generation, session injection, config normalization, dual runtime/document auth scenarios, session expiry / re-login, and e2e tests against a real password-protected Jupyter server.

### How it works

1. On startup, if a password is configured, `ServerContext` creates a `JupyterPasswordAuth` instance and calls `.login()` to authenticate.
2. The resulting session cookies are injected into `JupyterServerClient`'s HTTP session, and the auth's own `requests.Session` is kept alive so cookies remain readable for the document side.
3. For `KernelClient` and collaboration API requests, auth headers (`Cookie` + `X-XSRFToken`) are passed via the `headers` kwarg and computed per-request from the live cookie jar.
4. The `runtime_auth_headers` and `document_auth_headers` properties on `ServerContext` always read the latest cookies from the active session (not the stale login-time snapshot).
5. If a collaboration request returns `401/403` (expired cookie), the connection re-authenticates and retries once with the fresh cookies.

### Removed monkey-patch

Earlier revisions of this PR monkey-patched the module-level `connect` imported in `jupyter_nbmodel_client.client` to inject the Cookie header during the WebSocket handshake, because `NbModelClient` exposed no way to pass extra headers. That patch was not concurrency-safe.

This is now resolved upstream: [jupyter-nbmodel-client#61](https://github.com/datalayer/jupyter-nbmodel-client/pull/61) added `additional_headers=` to `NbModelClient` and `headers=` to `get_notebook_websocket_url` (released in `jupyter-nbmodel-client 0.14.8`). The monkey-patch and the re-implemented URL helper have been removed; the minimum dependency is bumped to `>=0.14.8`.

### Known limitation

The automatic re-login currently covers the **collaboration / document path** (`NotebookConnection`). Kernel-management HTTP requests that go through `jupyter_server_client` internals are not yet wrapped in the same retry; this can be a follow-up.

## Test plan

- [x] `uv run --extra test pytest tests/test_auth.py` passes (unit + e2e, including session-expiry/re-login tests against a real password-protected Jupyter server)
- [x] Test with a password-protected Jupyter server: `jupyter lab --IdentityProvider.token='' --ServerApp.password='argon2:...'`
- [x] Verify token-based auth still works unchanged (backward compatible)
- [x] Verify that `--jupyter-password` falls back correctly when individual passwords are not set
- [x] Verify that `jupyter-mcp-server connect --help` shows no `--*-password` options (passwords are local-only; not forwardable over `/api/connect`)

### DIFF
--- a/docs/docs/reference/configuration/index.mdx
+++ b/docs/docs/reference/configuration/index.mdx
@@ -52,11 +52,12 @@ For most users with standard JupyterLab setups:
 |----------|-------------|---------|---------|----------|
 | `JUPYTER_URL` | URL of your Jupyter server | `http://localhost:8888` | `http://localhost:8888` | No |
 | `JUPYTER_TOKEN` | Authentication token for Jupyter server | `my-secret-token` | `None` | No* |
+| `JUPYTER_PASSWORD` | Password for Jupyter server authentication (alternative to token) | `my-password` | `None` | No* |
 | `DOCUMENT_ID` | Default notebook path (relative to Jupyter root) | `notebook.ipynb` | `None` | No |
 | `ALLOWED_JUPYTER_MCP_TOOLS` | Comma-separated list of jupyter-mcp-tools to enable | `notebook_run-all-cells,notebook_get-selected-cell` | `notebook_run-all-cells,notebook_get-selected-cell` | No |
 | `ALLOW_IMG_OUTPUT` | Enable multimodal image support | `true` / `false` | `true` | No |
 
-*Required for authentication when connecting to secured Jupyter servers
+*At least one of `JUPYTER_TOKEN` or `JUPYTER_PASSWORD` is required when connecting to a secured Jupyter server. They are independent authentication mechanisms — a server may have both configured, and supplying either one is sufficient to connect. When both are provided to the MCP server, password authentication takes precedence. See [Security](../security) for details on when to use each.
 
 ### Advanced Configuration (Complex Deployments)
 
@@ -67,6 +68,7 @@ For deployments requiring granular control over document storage and runtime exe
 |----------|-------------|---------|---------|
 | `DOCUMENT_URL` | URL for notebook file operations | `http://notebook-storage:8888` | `http://localhost:8888` |
 | `DOCUMENT_TOKEN` | Authentication for document operations | `storage-access-token` | `None` |
+| `DOCUMENT_PASSWORD` | Password for document server authentication (alternative to token) | `storage-password` | `None` |
 | `DOCUMENT_ID` | Notebook path/ID | `shared/analysis.ipynb` | `None` |
 
 #### Runtime Execution Variables
@@ -74,6 +76,7 @@ For deployments requiring granular control over document storage and runtime exe
 |----------|-------------|---------|---------|
 | `RUNTIME_URL` | URL for kernel/execution operations | `http://compute-cluster:8888` | `http://localhost:8888` |
 | `RUNTIME_TOKEN` | Authentication for runtime operations | `compute-access-token` | `None` |
+| `RUNTIME_PASSWORD` | Password for runtime server authentication (alternative to token) | `compute-password` | `None` |
 | `RUNTIME_ID` | Specific kernel ID to use | `kernel-abc123` | `None` |
 
 #### MCP Client Authentication
@@ -173,6 +176,10 @@ Variables are resolved in this order:
 2. **Simplified variables** (`JUPYTER_*`) - fallback
 3. **Default values** - when no variables are set
 
+This applies to tokens, passwords, and URLs alike. For example, `RUNTIME_PASSWORD` takes precedence over `JUPYTER_PASSWORD` for runtime server authentication.
+
+When both a password and a token are configured for the same server, password authentication takes precedence and the token is ignored.
+
 ## Configuration Examples
 
 ### Standard JupyterLab Setup
@@ -202,6 +209,13 @@ RUNTIME_TOKEN=compute-access-token
 ALLOW_IMG_OUTPUT=false
 ```
 
+### Password-Protected Jupyter Server
+```bash
+JUPYTER_URL=http://localhost:8888
+JUPYTER_PASSWORD=my-jupyter-password
+DOCUMENT_ID=my-notebook.ipynb
+```
+
 ### Development Setup
 ```bash
 JUPYTER_URL=http://localhost:8888
@@ -224,6 +238,7 @@ Options:
   --jupyterlab BOOLEAN               Enable JupyterLab mode (default: true)
   --runtime-url TEXT                  Runtime URL for kernel operations (default: None)
   --runtime-token TEXT                Runtime authentication token (default: None)
+  --runtime-password TEXT             Password for runtime Jupyter server authentication (default: None)
   --mcp-token TEXT                    Token for authenticating MCP clients (required unless --insecure-mcp-noauth)
   --insecure-mcp-noauth               Allow streamable-http without MCP client auth (not recommended)
   --runtime-id TEXT                   Specific kernel ID to use (default: None)
@@ -231,8 +246,10 @@ Options:
   --document-url TEXT                 Document URL for notebook operations (default: None)
   --document-id TEXT                  Notebook path/ID (default: None)
   --document-token TEXT               Document authentication token (default: None)
+  --document-password TEXT            Password for document Jupyter server authentication (default: None)
   --jupyter-url TEXT                  Jupyter URL as default for both document and runtime URLs (default: None)
   --jupyter-token TEXT                Jupyter token as default for both document and runtime tokens (default: None)
+  --jupyter-password TEXT             Shared password for both runtime and document servers (default: None)
   --allowed-jupyter-mcp-tools TEXT    Comma-separated list of jupyter-mcp-tools to enable (default: notebook_run-all-cells,notebook_get-selected-cell)
   --reconnect-interval INTEGER         Seconds before retrying a dropped kernel WebSocket connection (default: 0)
   --execution-timeout INTEGER RANGE   Default timeout in seconds for code execution when a tool call passes no timeout (default: 120)
--- a/docs/docs/reference/security/index.mdx
+++ b/docs/docs/reference/security/index.mdx
@@ -161,28 +161,117 @@ Cross-Site Request Forgery (XSRF/CSRF) protection prevents unauthorized commands
 As described in [issue #183](https://github.com/datalayer/jupyter-mcp-server/issues/183), some Jupyter deployments don't use Bearer tokens but rely on XSRF cookies for authentication:
 
 **Affected Environments:**
-- Jupyter servers started without a token: `jupyter lab --IdentityProvider.token ''`
+- Jupyter servers that run without a Bearer token (for example, password-protected servers)
 - Enterprise deployments with SSO/OAuth/IAM
 - Managed environments (AWS SageMaker Studio, Google Colab Enterprise, Azure ML)
 - JupyterHub where authentication is handled by the Hub
 
-### Current Limitations
+## Password Authentication
 
-:::info Current Status
+For Jupyter servers configured with password-based login (instead of or in addition to tokens), the MCP server can authenticate by performing the standard Jupyter `/login` flow and using the resulting session cookies for all subsequent requests.
 
-**Bearer Token Required**: Currently, Jupyter MCP Server requires a Bearer token (`JUPYTER_TOKEN`) to authenticate with the Jupyter collaboration API.
+This is useful when:
+- Your Jupyter server uses a password set via `jupyter server password` or `--ServerApp.password`
+- You don't have a Bearer token available
+- Your deployment relies on XSRF cookie protection
 
-**Issue**: The server fails with `403 Forbidden` when connecting to Jupyter deployments that use XSRF protection without Bearer tokens.
+### How It Works
 
-**Tracking**: We're working on automatic XSRF cookie handling - see [issue #183](https://github.com/datalayer/jupyter-mcp-server/issues/183) for details.
+1. The MCP server POSTs to `/login` on the Jupyter server with the configured password
+2. Jupyter returns session cookies (including the `_xsrf` token)
+3. These cookies are injected into all HTTP requests (API calls, kernel operations) and WebSocket connections (notebook collaboration)
+4. The `X-XSRFToken` header is automatically included in requests that require XSRF protection
+
+### Configuration
+
+#### Simplified (Same Password for Both Servers)
+
+When document storage and runtime execution use the same Jupyter server:
+
+```json
+{
+  "env": {
+    "JUPYTER_URL": "http://localhost:8888",
+    "JUPYTER_PASSWORD": "my-jupyter-password"
+  }
+}
+```
+
+Or via CLI:
+
+```bash
+jupyter-mcp-server start \
+  --jupyter-url http://localhost:8888 \
+  --jupyter-password my-jupyter-password
+```
+
+#### Advanced (Separate Passwords)
+
+For deployments where document and runtime servers have different passwords:
+
+```json
+{
+  "env": {
+    "DOCUMENT_URL": "http://storage-server:8888",
+    "DOCUMENT_PASSWORD": "storage-password",
+    "RUNTIME_URL": "http://compute-server:8888",
+    "RUNTIME_PASSWORD": "compute-password"
+  }
+}
+```
+
+### Password vs Token
+
+| | Token Auth | Password Auth |
+|---|---|---|
+| **Mechanism** | Bearer token in `Authorization` header | Session cookies + XSRF token |
+| **Setup** | `--IdentityProvider.token MY_TOKEN` | `jupyter server password` |
+| **Best for** | API access, automation, JupyterHub | Local servers, password-protected deployments |
+| **XSRF handling** | Not needed (token bypasses XSRF) | Automatic (cookies include XSRF) |
+
+:::info Priority
+
+When both a password and a token are configured for the same server, **password authentication takes precedence**. The token is ignored and a warning is logged. This avoids ambiguity about which authentication method is active.
 
 :::
 
-### Workarounds for XSRF-only Environments
+### Setting a Jupyter Server Password
 
-Until automatic XSRF handling is implemented, you have these options:
+If your Jupyter server doesn't have a password configured yet:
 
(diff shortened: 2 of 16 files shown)


echarles:
> NbModelClient (from jupyter-nbmodel-client) does not expose an additional_headers parameter for its WebSocket connection. As a workaround, NotebookConnection.__aenter__ temporarily monkey-patches the connect function to inject the Cookie header during WebSocket establishment. This is thread-safe (restore happens in a finally block) but ideally should be replaced by an upstream additional_headers parameter in jupyter-nbmodel-client.


should this be fixed in upstream jupyter-nbmodel-client? not sure how this would help this PR?

EnyMan:
@echarles It would allow us to get rid of this not great piece of code:
 ```python
 def _connect_with_auth(uri, **kwargs):
    kwargs.setdefault("additional_headers", {})
    kwargs["additional_headers"]["Cookie"] = cookie_header
    return original_connect(uri, **kwargs)

nbmodel_client_module.connect = _connect_with_auth
```
in `jupyter_mcp_server/notebook_manager.py:113`

But if we are fine with it, we can keep it as is. But it's somewhat fragile if the `nbmodel_client_module` changes in how it works, this will break.

**Comment by echarles** [comment]
> Related issues: Partially addresses https://github.com/datalayer/jupyter-mcp-server/issues/183 (XSRF-protected Jupyter deployments without Bearer tokens), related to https://github.com/datalayer/jupyter-mcp-server/issues/61 (WebSocket authentication failures)

As we are talking about security, very clear explanations would be useful. Is it possible to create a security section / page in the docs, and explain the various ways and implications the jupyter mcp server can be configured, included those changes.

**Comment by abbbe** [comment]
I wonder if the tests run against the real instance of a jupyter server or the test code only test a standalone instance of the newly added class mean to implement password based auth?

**Comment by EnyMan** [comment]
@abbbe i tested this implementation with claude code agains real jupyter lab server multiple times

**Comment by abbbe** [comment]
I am not speaking on behalf of the project here, mere a party interested in the feature, but “works for me” is not quite the same as proper CI. The project test suite already has all the machinery to spawn Jupiter Lab server etc, it is not too complicated to integrate your tests.

**Comment by EnyMan** [comment]
@echarles Good point — I've added documentation for the password auth feature
  
  Security docs (docs/docs/reference/security/index.mdx):
  - New "Password Authentication" section covering the login flow, XSRF handling, simplified and advanced configuration (with JSON and CLI examples), a comparison table of password vs token auth, priority behavior, and how to set up a Jupyter server password
  - Updated the "Current Limitations" section to reflect that password-protected XSRF deployments are now supported, while noting that SSO/OAuth/IAM scenarios from #183 remain open 
  - Kept the token-based workarounds as "Alternatives for Token-less Environments"

  Configuration docs (docs/docs/reference/configuration/index.mdx):
  - Added JUPYTER_PASSWORD, DOCUMENT_PASSWORD, RUNTIME_PASSWORD to the environment variable tables
  - Added --jupyter-password, --runtime-password, --document-password to the CLI options reference
  - Added a "Password-Protected Jupyter Server" configuration example
  - Updated the priority section to document password-over-token precedence

  Let me know if you'd like any changes to the structure or content.

**Comment by echarles** [comment]
@EnyMan Awesome!

Maybe @abbbe can go through a first pass on these changes as he has been key to introduce more security recently, I can then review as follow-up?

**Comment by abbbe** [comment]
@EnyMan this is a very cool feature, actually wanted to implement it myself but never could wrap my head around it.

I am happy to try to review it, but would you be open to do the following to begin with:
* align your fork with the latest main (now it is 9 commits behind),
* Implement proper CI (run tests against real Jupyter instance as done by token auth, I can point you to specific old commits you can use for inspiration, it is not that complicated)
* confirm you have done a basic automated analysis of your PR (here is one done by Claude https://claude.ai/code/session_01EYGdzWYfnApH8AjX28nFLF). I do not imply AI has a final say here and aware it is subject to hallucinations, but I believe it is a valuable tool to catch low hanging fruits.

File: docs/docs/reference/configuration/index.mdx [hunk 1]
```diff
@@ -47,11 +47,12 @@ For most users with standard JupyterLab setups:
 |----------|-------------|---------|---------|----------|
 | `JUPYTER_URL` | URL of your Jupyter server | `http://localhost:8888` | `http://localhost:8888` | No |
 | `JUPYTER_TOKEN` | Authentication token for Jupyter server | `my-secret-token` | `None` | No* |
+| `JUPYTER_PASSWORD` | Password for Jupyter server authentication (alternative to token) | `my-password` | `None` | No* |
 | `DOCUMENT_ID` | Default notebook path (relative to Jupyter root) | `notebook.ipynb` | `None` | No |
 | `ALLOWED_JUPYTER_MCP_TOOLS` | Comma-separated list of jupyter-mcp-tools to enable | `notebook_run-all-cells,notebook_get-selected-cell` | `notebook_run-all-cells,notebook_get-selected-cell` | No |
 | `ALLOW_IMG_OUTPUT` | Enable multimodal image support | `true` / `false` | `true` | No |
 
-*Required for authentication when connecting to secured Jupyter servers
+*Either `JUPYTER_TOKEN` or `JUPYTER_PASSWORD` is required when connecting to secured Jupyter servers. See [Security](/reference/security) for details on when to use each.
```
abbbe: The link to 'Security' document is broken

abbbe: > Either `JUPYTER_TOKEN` or `JUPYTER_PASSWORD` is required when connecting to secured Jupyter servers

Technically, this is not a correct statement, these are two independent authentication mechanisms. Both can be active.

abbbe:
> @echarles It would allow us to get rid of this not great piece of code:
> 
> ```python
> def _connect_with_auth(uri, **kwargs):
>    kwargs.setdefault("additional_headers", {})
>    kwargs["additional_headers"]["Cookie"] = cookie_header
>    return original_connect(uri, **kwargs)
> 
> nbmodel_client_module.connect = _connect_with_auth
> ```
> 
> in `jupyter_mcp_server/notebook_manager.py:113`
> 
> But if we are fine with it, we can keep it as is. But it's somewhat fragile if the `nbmodel_client_module` changes in how it works, this will break.

I think this (extending nbmodel client) is the best approach, all this monkey patching business is rather sketchy.

EnyMan:
Better test are on the way

docs/docs/reference/configuration/index.mdx
EnyMan: Should be fixed

EnyMan:
@abbbe Thanks for looking at this I did several rounds against grumpy opus reviewer.

echarles:
@abbbe Any further comment before I review this?

abbbe:
I will have a look todaySent from my iPhoneOn 29 May 2026, at 09:38, Eric Charles ***@***.***> wrote:﻿echarles left a comment (datalayer/jupyter-mcp-server#250)
@abbbe Any further comment before I review this?

docs/docs/reference/security/index.mdx
```diff
@@ -155,23 +155,103 @@ As described in [issue #183](https://github.com/datalayer/jupyter-mcp-server/iss
...
+### Limitations
+
+:::info Partial Coverage of Issue #183
+
+Password authentication addresses the XSRF-protected scenario described in [issue #183](https://github.com/datalayer/jupyter-mcp-server/issues/183) for password-protected Jupyter servers. Other scenarios mentioned in that issue — SSO/OAuth, IAM-based auth in managed environments — are not yet supported.
 
 :::
 
-### Workarounds for XSRF-only Environments
+### Alternatives for Token-less Environments
```
abbbe: This section is about token-based deployment, maybe the title can be "Token-based authentication (alternative to XSFR)" or something.

jupyter_mcp_server/CLI.py
```diff
@@ -110,67 +158,64 @@ def _common_options(f):
...
     return f
 
 
-def _resolve_url_and_token_variables(
-    jupyter_url, jupyter_token,
-    document_url, document_token,
-    runtime_url, runtime_token,
-) -> tuple[str, str | None, str, str | None]:
-    """Resolve URL and token variables based on priority logic.
+class _ResolvedConnection(NamedTuple):
```
abbbe: Ideally, ResolvedConnection should cover the full connection input set. It currently holds only the fields _resolve_connection_variables merges (URLs/tokens/passwords), so callers pass resolved.runtime_url next to the un-resolved runtime_id into _do_start — 14 kwargs split across two sources, and the name overstates the scope. Suggest widening the NamedTuple to also carry document_id, runtime_id, provider, jupyterlab (threaded through the resolver unchanged) so _do_start takes a single _ResolvedConnection and the type matches its name.

jupyter_mcp_server/notebook_manager.py
```diff
@@ -39,16 +83,50 @@ async def __aenter__(self) -> NbModelClient:
...
+            import jupyter_nbmodel_client.client as nbmodel_client_module
+            from websockets.asyncio.client import connect as original_connect
+            cookie_header = auth_headers.get("Cookie", "")
+
+            def _connect_with_auth(uri, **kwargs):
+                kwargs.setdefault("additional_headers", {})
+                kwargs["additional_headers"]["Cookie"] = cookie_header
+                return original_connect(uri, **kwargs)
+
+            nbmodel_client_module.connect = _connect_with_auth
```
abbbe: I think the race condition highlighted in the very first Claude Code analysis is still here -- in case of two concurrent tool calls the monkey patch might be rolled back by 'finally' of one thread, while second thread still relies on it to be in effect. Do you need more explanations on this point?

abbbe: Alternatively, we have to serialise tool calls, but this is dirty...

tests/test_auth.py
```diff
@@ -0,0 +1,808 @@
...
+
+    @pytest.mark.asyncio
+    async def test_password_auth_execute_code(self, mcp_client_password):
+        """KernelClient works with password cookie/XSRF auth."""
+        async with mcp_client_password:
+            result = await mcp_client_password.execute_code("2 + 2")
+        assert result is not None
+        assert "result" in result
+        outputs = result["result"]
+        assert any("4" in str(output) for output in outputs)
```
abbbe: I would make a somewhat stricter test, character '4' might randomly appear in error message. Maybe multiply two large ints or something, to keep it simple.

tests/test_auth.py
```diff
@@ -0,0 +1,808 @@
...
+                notebook_name="password_test",
+                notebook_path="notebook.ipynb",
+                mode="connect",
+            )
+            assert isinstance(use_result, str)
+            assert "error" not in use_result.lower(), f"use_notebook returned an error: {use_result}"
+
+            # Read a cell to verify the notebook connection is working
+            cell = await mcp_client_password.read_cell(cell_index=0)
+            assert cell is not None
```
abbbe, line 843: I think more comprehensive testing is justified here too. Could rerun the same operation from the test about (multiply two large random ints or whatever) in a cell and check the content.

tests/conftest.py
```diff
@@ -418,3 +418,74 @@ def mcp_client_otel(mcp_server_url_otel):
...
+    port = _find_free_port()
+    yield from _start_server(
+        name="Jupyter MCP (password)",
+        host=host,
+        port=port,
+        command=[
+            "python", "-m", "jupyter_mcp_server",
+            "--transport", "streamable-http",
+            "--document-url", jupyter_server_password,
+            "--document-id", "notebook.ipynb",
```
abbbe: I realise that you are mimicking current test implementation, but as we are writing new tests I would use cleaner code. We should use random temporary notebooks instead of hardcoding stuff. Also, I would put them in some temp folder instead of ./dev/content (that folder should be for notebooks coming from the repo, not temporary scratch files). The problem with static hardcoded notebook names is test parallel tests can fail unpredicably.

abbbe:
There is something else. The current implementation logs in once, but does not retry if session goes stale (cookie expires) throughout the session. I understand by default jupyter cookies expire in 30 days and normal desktop users are unlikely to be affected, but long running sessions might fail. I did not investigate this in details, because things might change depending on how the race condition will be fixed. Something to look into...

abbbe:
Also, I think setting token to blank value is not useful but create risks. Not useful, because password-based authentication gives access to API even without IdentityProvider.token="".

Dangerous, because if things go wrong with the password based authentication (for whatever reason), user ends up with zero authentication.

jupyter_mcp_server/notebook_manager.py
EnyMan: How realistic would be to have the upstream changes done? Should I also look at it?

EnyMan:
Thanks for the review i should be able to look at it sometime this week.

docs/docs/reference/configuration/index.mdx
EnyMan: Ah. A different wording would perhaps be a better. maybe instead of `Either`, `At least one of` would make the sentence sound like both can be active at the same time? Because currently when both are configured the password takes precedence.

jupyter_mcp_server/notebook_manager.py
abbbe: I don’t see why not, this is a perfectly reasonable improvement I think.

EnyMan:
@abbbe Good catch on the Jupyter cookie expiration i haven't thought about it. As for the implementation of it i plan to look for 401/403 an re-auth. By any chance do you know if the collaboration WebSocket surfaces this cleanly? I am also going to wait for the `nbmodel_client_module` change so i can do that and the re-auth in the same pass.

abbbe:
> By any chance do you know if the collaboration WebSocket surfaces this cleanly?

My guess is web socket is not affected, in a sense that once it is open with a valid cookie, the web socket itself just stays open even if the cookie used to open it has expired.

EnyMan:
@abbbe Removed the monkey-patch and I added the re-login flow to the collaboration/document path. Do you want it added also to the Kernel management path `jupyter_server_client`?

EnyMan:
While testing this branch I hit a bug: `use_notebook(mode="create")` failed under password auth with `403: '_xsrf' argument missing from POST`. Fixed it and added an E2E test for create mode.

echarles:
@EnyMan Is this ready for review?

EnyMan:
> @EnyMan Is this ready for review?

Yes, it is. Let me rebase to master really quickly,  ok it might not be that quick since the CLI changes

EnyMan:
Fixing the test plus some new additions that were since added to main.

echarles:
thx, are the items listed in the first note still opened?

```
 Verify token-based auth still works unchanged (backward compatible)
 Verify that --jupyter-password falls back correctly when individual passwords are not set
 Verify that jupyter-mcp-server connect --help shows no --*-password options (passwords are local-only; not forwardable over /api/connect)
```

EnyMan:
> thx, are the items listed in the first note still opened?
> 
> ```
>  Verify token-based auth still works unchanged (backward compatible)
>  Verify that --jupyter-password falls back correctly when individual passwords are not set
>  Verify that jupyter-mcp-server connect --help shows no --*-password options (passwords are local-only; not forwardable over /api/connect)
> ```

The last one is done. For the first two, I can either test manually or try to write tests for them. The manual test might need to wait until tomorrow.

EnyMan:
OK verified the two last two points

echarles:

EnyMan:
Rebased and opened up datalayer/code-sandboxes#13; it's the same kind of issue as with the nb-client.
Also, the test fails due to an unbound MCP dependency; they released v2.

echarles:
> Rebased and opened up https://github.com/datalayer/code-sandboxes/pull/13; it's the same kind of issue as with the nb-client.

Great, merged and code sandboxes 0.17.0 release with that

> Also, the test fails due to an unbound MCP dependency; they released

https://github.com/datalayer/jupyter-mcp-server/pull/328 is merged and pins mcp<2 - new release on its way.

EnyMan:
Rebased and pushed. So unless something breaks again, ready to review @echarles

echarles (approved):
LGTM Thx @EnyMan

Issue #253, opened 2026-06-17
Kernels created by `jupyter-mcp-server` always have server root as working directory


rfilmyer:
## Description

When an AI tool uses `jupyter-mcp-server` to open a Jupyter notebook in Jupyter Lab, the opened notebook always has the server's root set as the working directory. Human users using Jupyter Lab will have the notebook's directory as the working directory.

In other words, if your directory structure is like so:
```
notebooks/
└── example-folder/
      └── example-notebook.ipynb
```

An agent opening `example-notebook.ipynb` via the MCP will have `notebooks/` as its working directory. A human user on Jupyter Lab who opens the notebook will instead have `example-folder/`

### Setup

  Create a notebook in a subdirectory of the Jupyter server root:
```
  <server_root>/
  └── subdir/
      └── test.ipynb
```

### Steps to reproduce

  1. Start Jupyter Lab from <server_root>/
  2. Open subdir/test.ipynb via the MCP using the use_notebook tool with `notebook_path="subdir/test.ipynb"`
  3. Execute the following code in a cell:
```python
  import os; print(os.getcwd())
```

### Actual result (for agents)
  /path/to/server_root

### Expected result (matches humans)
  /path/to/server_root/subdir

## Expected behavior

The CWDs should line up between humans and agents, so that notebooks that use relative paths will resolve to the same locations.

## Context

- Datalayer version: N/A
- Operating System and version: Mac OS Tahoe 26.5.1
- Browser and version: Firefox 151.0.4

Additional Details:
- `jupyter_mcp_server` 1.0.2
- Jupyter Lab 4.5.9
- Python 3.12
- Claude CLI version 2.1.179 with Sonnet 4.6 

Browser Output: N/A

## Diagnosis

`UseNotebookTool._start_kernel_local()` in tools/use_notebook_tool.py calls `kernel_manager.start_kernel()` with no arguments. `MappingKernelManager.start_kernel()` accepts a path parameter that it passes through `cwd_for_path()` to set the kernel's working directory to the notebook's folder — but it's never called with it. JupyterLab passes path when a human opens a notebook; the MCP extension does not.

## Possible Fix

(Yeah I used claude to figure it out)

Two-line change across two files:

  1. `tools/use_notebook_tool.py` — `_start_kernel_local(self, kernel_manager)` → `_start_kernel_local(self, kernel_manager, path=None)`, and `start_kernel()` →
[...]

PR #254, opened 2026-06-18
Working Directory Fix for #253

rfilmyer:
This is the 2-line change mentioned in #253, passing a path to the Jupyter notebook kernel opening function so that the MCP behavior lines up with Jupyter Lab human user behavior (opening a notebook will have the working directory set to the folder that it is in)


echarles (approved):
LGTM Thx @rfilmyer

PR #267, opened 2026-07-13
fix: self-heal restart_notebook when the kernel was culled (JUPYTER_SERVER mode)

### LINKED ISSUE #260: JUPYTER_SERVER mode: stale kernel binding after kernel culling/server restart — execute_cell times out and the agent cannot self-heal
## Environment
- jupyter-mcp-server 1.0.2 running as a server extension (JUPYTER_SERVER mode) inside a JupyterHub (TLJH) single-user server
- jupyter_server 2.x, jupyter_server_nbmodel 0.1.1a4, jupyter_server_fileid 0.9.3, python 3.10

## Summary
The `notebook_manager` keeps the kernel id recorded by `use_notebook`. When that kernel disappears — idle-culling (common on JupyterHub) or a single-user server restart — the binding goes stale, and the tool behavior makes it hard for an agent (or a human driving one) to recover:

1. `execute_cell` runs against the dead kernel and blocks until timeout (in some paths it errors with `HTTP 404: Not Found (Kernel does not exist: <id>)`).
2. `restart_notebook` fails with the same 404 instead of provisioning a fresh kernel:
   ```
   INFO   Restarting kernel      restart_notebook_tool.py:58
   ERROR  Failed to restart      restart_notebook_tool.py:62
          404: Not Found (Kernel does not exist: ...)
   ```
3. A plain retry of `use_notebook` with the same name/path is rejected with
   `Notebook '<name>' is already activated now. DO NOT REACTIVATE AGAIN.` — so the
   obvious self-heal path is blocked.
4. The only working recovery is `unuse_notebook` followed by `use_notebook`, which
   agents don't discover on their own; users see repeated execution timeouts.

## Steps to reproduce
1. `use_notebook` (create or connect) — a kernel is started and recorded.
2. Kill that kernel out-of-band (`DELETE /api/kernels/<id>`), or let the hub's idle culler do it, or restart the single-user server.
3. `execute_cell` → hangs until timeout or returns the 404 error.
4. `restart_notebook` → 404.
5. `use_notebook` (same name) → "already activated, DO NOT REACTIVATE".

## Expected
- `execute_cell` should validate kernel liveness first and (re)provision via the existing `ensure_kernel_alive` machinery instead of waiting on a dead kernel.
- `restart_notebook` should treat "kernel does not exist" as "start a new kernel and rebind", not as a failure.
- Alternatively/additionally, `use_notebook` on an already-registered notebook whose kernel is dead should rebind instead of refusing.

## Related
#21 reports the same class of problem (kernel connection presumed alive indefinitely, no revalidation before execute) in the MCP_SERVER/KernelClient architecture. This report covers the JUPYTER_SERVER extension mode in 1.0.2, where the failure surfaces through the notebook_manager binding and the restart/reactivate tools.

## Notes
This is especially visible on JupyterHub deployments where idle culling is standard and chat/agent sessions outlive kernels by design.

ebarkhordar:
## Root cause

In JUPYTER_SERVER mode, `restart_notebook` reads the kernel id recorded by `use_notebook` and calls `kernel_manager.restart_kernel(kernel_id)` directly. When that kernel has been idle-culled (standard on JupyterHub) or lost to a single-user server restart, `restart_kernel` raises `404: Not Found (Kernel does not exist: <id>)`. The tool catches it and returns "Failed to restart", so the notebook stays bound to a kernel that no longer exists and the agent cannot recover through the tool. The only working path today is `unuse_notebook` then `use_notebook`, which agents do not discover on their own. This is the `restart_notebook` part of #260.

## Invariant

A restart on a notebook whose kernel is gone must self-heal: provision a fresh kernel and rebind it, rather than fail on a 404. An agent should never be left with a dead binding it cannot recover through the restart tool.

## Fix

Before restarting, check kernel liveness with `kernel_id in kernel_manager`, the same idiom `use_notebook` already uses. If the kernel is gone, start a new kernel with the notebook's path (mirroring `local_backend`, which passes the path so the kernel's working directory matches the notebook) and rebind it via `notebook_manager.add_notebook`, then report success. An `except` branch covers the race where the kernel is culled between the liveness check and the `restart_kernel` call. The live-kernel path is unchanged, and after a rebind `notebook_manager.get_kernel_id` returns the new id, so the `restarted` lifecycle hook fires with the correct kernel.

Scope: this PR fixes only the `restart_notebook` self-heal (the second item under #260 "Expected"). The `execute_cell` liveness check and the `use_notebook` rebind suggestions are left out to keep one concern per PR.

## How I verified

Ran in a clean `python:3.10-slim` Docker container (editable install with `.[test]`):

- New offline unit test `tests/test_restart_notebook_tool.py` using a minimal in-memory kernel-manager stand-in (no running Jupyter server), covering three cases: a culled kernel reprovisions and rebinds to the new id and starts it with the notebook path; a kernel culled during restart (the race) reprovisions via the `except` branch; a live kernel restarts in place with the id unchanged and no new kernel started.
- Differential: the two self-heal tests FAIL on `origin/main`'s `restart_notebook_tool.py` (it returns "Failed to restart ... 404" and the binding stays stale) and PASS on this branch; the live-kernel test passes both ways, so the happy path does not regress. Provenance was checked by toggling the source file version.
- `tests/test_config.py` and `tests/test_hooks.py` still pass; `ruff check` on the changed file adds no rule categories beyond the file's existing style.

What I did not verify: I have no live JupyterHub with an idle-culler here, so I did not exercise a real culled kernel end to end. The 404 and the reprovision path are reproduced against an in-memory `MappingKernelManager` stand-in that mirrors its `__contains__`, `restart_kernel`, and `start_kernel` surface, not the real server. I also did not run the full parametrized integration suite (`tests/test_tools.py`), which needs a running Jupyter server.

Fixes #260

---

AI assistance: this change was prepared with the help of an AI agent (Claude). A human is accountable for it, and I am happy to adjust the approach.

### DIFF
--- a/jupyter_mcp_server/tools/restart_notebook_tool.py
+++ b/jupyter_mcp_server/tools/restart_notebook_tool.py
@@ -15,7 +15,41 @@ logger = logging.getLogger(__name__)
 
 class RestartNotebookTool(BaseTool):
     """Tool to restart the kernel for a specific notebook."""
-    
+
+    async def _reprovision_kernel(
+        self,
+        kernel_manager: Any,
+        notebook_manager: NotebookManager,
+        notebook_name: str,
+    ) -> str:
+        """Start a fresh kernel and rebind it to the notebook (JUPYTER_SERVER mode).
+
+        Used when the notebook's recorded kernel has been culled or lost, so the
+        agent can self-heal instead of being stuck with a dead kernel binding.
+        """
+        try:
+            notebook_path = notebook_manager.get_notebook_path(notebook_name)
+            new_kernel_id = await kernel_manager.start_kernel(path=notebook_path)
+            notebook_manager.add_notebook(
+                notebook_name,
+                {"id": new_kernel_id},
+                server_url="local",
+                token=None,
+                path=notebook_path,
+            )
+            logger.info(f"Provisioned fresh kernel {new_kernel_id} for notebook '{notebook_name}'")
+            return (
+                f"Notebook '{notebook_name}' kernel was no longer available and has been "
+                f"reprovisioned (new kernel '{new_kernel_id}'). Memory state and imported "
+                f"packages have been cleared."
+            )
+        except Exception as e:
+            logger.error(f"Failed to reprovision kernel for notebook '{notebook_name}': {e}")
+            return (
+                f"Failed to restart notebook '{notebook_name}': the kernel was no longer "
+                f"available and reprovisioning failed: {e}"
+            )
+
     async def execute(
         self,
         mode: ServerMode,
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks @echarles, appreciate the quick review and the merge. Glad this one is in.

PR #269, opened 2026-07-16
feat: add open_notebook_in_ui opt-in for use_notebook tab activation (fixes #218)

### LINKED ISSUE #218: use_notebook opens notebook in JupyterLab UI - unexpected side effect
### Problem

When calling use_notebook, the tool currently triggers docmanager_open to open the notebook in the JupyterLab UI. This feels like an unexpected side effect - use_notebook is conceptually a data/connection operation (register a notebook, start a kernel), but it also steals UI focus by switching the active tab in JupyterLab.

This can be surprising for users, especially when an MCP client is switching between multiple notebooks programmatically - each switch causes the JupyterLab UI to jump to a different tab.

### Proposed Solution

Would it make sense to make the docmanager_open behavior optional, and perhaps off by default? For example, a configuration flag like open_in_ui=False on use_notebook, or a server-level setting.

Users who want the tab to follow MCP operations could opt in, while the default behavior would keep use_notebook as a pure data operation without UI side effects.

### Additional context

The relevant code is in jupyter_mcp_server/tools/use_notebook_tool.py (lines ~300-343), where MCPToolsClient.execute_tool("docmanager_open", ...) is called at the end of execute() when JupyterLab mode is enabled.

ebarkhordar:
Follows up on the design discussion in #218.

## Root cause

`use_notebook` dispatches the jupyter-mcp-tools `docmanager_open` command whenever JupyterLab mode is on (`use_notebook_tool.py`, the block around lines 300-343). Opening the document activates its tab, so an agent that connects to or switches notebooks in the background pulls the UI focus away from the tab the user is working in. There is currently no way to opt out.

As @abbbe found in the issue, `JUPYTERLAB=false` is not that opt-out. The same flag also gates loading the jupyter-mcp-tools tool set (`get_tools()` in `server.py`, and `handlers.py`), so turning it off to stop the tab switch drops those tools as well. The two behaviours are coupled under one flag.

## Invariant

Switching notebooks over MCP must not activate a JupyterLab tab unless the user opted in.

## The fix

A dedicated `open_notebook_in_ui` option, defaulting to `False`, gating only the `use_notebook` UI-open block:

```python
if context.is_jupyterlab_mode() and get_config().open_notebook_in_ui:
    ... docmanager_open ...
```

It is wired like the existing `jupyterlab` flag: config field, CLI option (`--open-notebook-in-ui`), `OPEN_NOTEBOOK_IN_UI` env var, and extension trait. jupyter-mcp-tools loading is untouched, so anyone who wants the current behaviour sets the flag and gets it back.

This does change the default: background notebook switches no longer activate the tab. That matches your "I am open to set by default to false based on your feedbacks" in the issue, but it is your call, and flipping the default to `True` here is a one-line change if you would rather keep today's behaviour and let people opt out instead.

## How I verified

Docker `python:3.12-slim`, editable install at `a116ed6`, module resolved to `/src/jupyter_mcp_server/tools/use_notebook_tool.py`.

The regression tests drive `UseNotebookTool.execute` on the notebook-switch path with an in-memory stand-in for `MCPToolsClient` that records the tools it is asked to execute (the same stand-in approach as `tests/test_restart_notebook_tool.py`):

- Against unfixed source at `a116ed6`, `test_switching_notebook_does_not_open_ui_by_default` fails, recording `('docmanager_open', {'path': 'work/nb2.ipynb'})`. That is the focus-steal, reproduced offline.
- On this branch the same four tests pass: no dispatch by default, dispatch when `open_notebook_in_ui` is set, still no dispatch when JupyterLab mode is off, and the default is off.

```
TEST_MCP_SERVER=true TEST_JUPYTER_SERVER=false pytest
```

gives `93 passed, 1 failed, 7 skipped, 54 errors` on this branch and `89 passed, 1 failed, 7 skipped, 54 errors` on a clean checkout of `a116ed6` in the same container: an identical failure profile, plus the 4 tests added here. The 54 errors and the `test_otel_integration.py::test_span_log_summary` failure are the tests that need a live JupyterLab, which my container does not start; they are not related to this change. I also ran the CI smoke check (every module imports, `start --help` lists the new flag).

## What I did not verify

I have not exercised this against a live JupyterLab UI, so the tests cover the server-side decision to dispatch `docmanager_open`, not the browser behaviour that dispatch causes. @abbbe also reported that `JUPYTERLAB=false` did not suppress the activation, which you attributed to the jupyter-mcp-tools extension; that entrypoint-wiring question is out of scope here and this change does not address it.

I did not run `.pre-commit-config.yaml` over the changed files: `ruff` and `ruff-format` there fail on the repo's current files as well, so running them would have reformatted five files and buried this change in unrelated churn. I checked instead that the change adds no new lint errors of any class beyond the long `Field`/`help` lines that match their neighbours, and the new test file reports the same ruff findings as the existing `tests/test_restart_notebook_tool.py`.

Happy to rename the flag or adjust the default if you prefer something else.

Fixes #218

Written with AI assistance; I reproduced and verified the behaviour above myself.

### DIFF
--- a/docs/docs/reference/configuration/index.mdx
+++ b/docs/docs/reference/configuration/index.mdx
@@ -16,8 +16,13 @@ The server automatically detects the appropriate mode based on deployment contex
 ### JupyterLab Mode
 **Environment Variable:** `JUPYTERLAB=true` (default)
 
-- **Enabled**: Automatic notebook opening, enhanced UI tools, `notebook_run-all-cells` tool
+- **Enabled**: Enhanced UI tools, `notebook_run-all-cells` tool
 - **Disabled**: Basic operations only, headless usage
+
+### Open Notebook in UI
+**Environment Variable:** `OPEN_NOTEBOOK_IN_UI=false` (default)
+
+When enabled together with JupyterLab mode, `use_notebook` also opens the notebook in the JupyterLab UI, which activates its tab. It is off by default so that notebooks used over MCP do not pull the focus away from the tab you are working in.
 
 ## Transport & Providers
 
@@ -86,6 +91,7 @@ For deployments requiring granular control over document storage and runtime exe
 | `PORT` | Port for streamable HTTP transport | `4040` | `4040` |
 | `RECONNECT_INTERVAL` | Seconds before retrying a dropped kernel WebSocket connection. `0` disables auto-reconnect. | `5` | `0` |
 | `JUPYTERLAB` | Enable JupyterLab mode | `true` / `false` | `true` |
+| `OPEN_NOTEBOOK_IN_UI` | Open the notebook in the JupyterLab UI on `use_notebook`, activating its tab | `true` / `false` | `false` |
 | `JUPYTER_MCP_OTEL_FILE` | Path for OpenTelemetry span export (JSONL) | `/tmp/spans.jsonl` | `None` |
 
 ### Configuration Priority
--- a/jupyter_mcp_server/CLI.py
+++ b/jupyter_mcp_server/CLI.py
@@ -41,6 +41,13 @@ def _common_options(f):
             help="Enable JupyterLab mode. Defaults to True.",
         ),
         click.option(
+            "--open-notebook-in-ui",
+            envvar="OPEN_NOTEBOOK_IN_UI",
+            type=click.BOOL,
+            default=False,
+            help="Open the notebook in the JupyterLab UI when using it, which activates its tab. Defaults to False.",
+        ),
+        click.option(
             "--runtime-url",
             envvar="RUNTIME_URL",
             type=click.STRING,
@@ -198,6 +205,7 @@ def _do_start(
     port: int,
     provider: str,
     jupyterlab: bool,
+    open_notebook_in_ui: bool,
     allowed_jupyter_mcp_tools: str,
     otel_file: str = "",
     mcp_token: str = None,
@@ -236,6 +244,7 @@ def _do_start(
         document_token=document_token,
         port=port,
         jupyterlab=jupyterlab,
+        open_notebook_in_ui=open_notebook_in_ui,
         allowed_jupyter_mcp_tools=allowed_jupyter_mcp_tools,
         reconnect_interval=reconnect_interval,
     )
@@ -360,6 +369,7 @@ def server(
     port: int,
     provider: str,
     jupyterlab: bool,
+    open_notebook_in_ui: bool,
     allowed_jupyter_mcp_tools: str,
     otel_file: str,
     reconnect_interval: int,
@@ -398,6 +408,7 @@ def server(
         port=port,
         provider=provider,
         jupyterlab=jupyterlab,
+        open_notebook_in_ui=open_notebook_in_ui,
         allowed_jupyter_mcp_tools=allowed_jupyter_mcp_tools,
         otel_file=otel_file,
         mcp_token=mcp_token,
@@ -427,6 +438,7 @@ def connect_command(
     document_token: str,
     provider: str,
     jupyterlab: bool,
+    open_notebook_in_ui: bool,
     jupyter_url: str,
     jupyter_token: str,
     allowed_jupyter_mcp_tools: str,
@@ -452,7 +464,8 @@ def connect_command(
         document_url=resolved_document_url,
         document_id=document_id,
         document_token=resolved_document_token,
-        jupyterlab=jupyterlab
+        jupyterlab=jupyterlab,
+        open_notebook_in_ui=open_notebook_in_ui
     )
     
     # Also update the jupyter_extension ServerContext with the jupyterlab flag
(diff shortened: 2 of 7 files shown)


ebarkhordar:
The red `macos-latest, 3.10` leg is not from this branch. The only failure is `tests/test_tools.py::test_management_routes_accept_mcp_token` with `httpx.ReadTimeout: timed out`, which this PR does not touch. It fails the same way on main: run 29224805709 (`macos-latest, 3.10`, commit 41c04069) ends with the identical `1 failed, 140 passed` on that same test.

The other five matrix legs are green here, and the four new tests in `tests/test_use_notebook_ui_open.py` pass on all six. A rerun of that one leg should clear it whenever it suits you.

echarles:
I have relaunched and test is still failing. This is odd, is there a way to fix that failure?

ebarkhordar:
My rerun suggestion was wrong, and it cost you a run. Sorry about that. This is not a random flake: the test fails deterministically whenever the kernel shutdown takes longer than 5 seconds, and `macos-latest, 3.10` is evidently the leg that crosses that line.

The two timeouts are inverted:

- `tests/test_tools.py:239` calls `httpx.delete(f"{mcp_server_with_mcp_token}/api/stop", ...)` with no `timeout`, so it uses the httpx default, which is 5.0s.
- `/api/stop` (`server.py:266`) calls `notebook_manager.remove_notebook()`, which at `notebook_manager.py:143` calls `KernelClient.stop()`. That method is sync, and it runs `self._manager.shutdown_kernel(now=shutdown_now, timeout=timeout)` with `timeout=REQUEST_TIMEOUT`, which is `10` in jupyter-kernel-client.

So the route is allowed to spend up to 10s finishing the shutdown before it can write its response, while the client gives up at 5s. Any shutdown slower than 5s fails, which is why the rerun did not clear it.

What I checked, on `main` and without this branch in the picture:

- The same test fails the same way on main: run 29224805709 (commit 41c04069) and run 29141046322 (commit c2e98fca), both `macos-latest, 3.10`, both `1 failed, 140 passed` on this test with `httpx.ReadTimeout`. Other main runs on that leg pass, which fits a threshold being crossed intermittently rather than a code change.
- `inspect.iscoroutinefunction(KernelClient.stop)` is `False`, and `REQUEST_TIMEOUT` is `10` (python:3.10 container).
- Mechanism check in Docker: an async Starlette route that blocks for 6s gives `httpx.ReadTimeout` to a default-timeout client, and `200` to the same request with `timeout=30`.

What I did not work out is why the shutdown crosses 5s on the macOS runner and not on ubuntu.

The smallest fix that matches the cause is to let the client wait longer than the server's own shutdown budget:

```python
r = httpx.delete(
    f"{mcp_server_with_mcp_token}/api/stop",
    headers={"Authorization": f"Bearer {MCP_TOKEN}"},
    timeout=30.0,
)
```

Separate and optional: `remove_notebook` runs a blocking call inside an async route, so a slow kernel shutdown currently stalls every other request, including `/api/healthz`. Moving it off the loop with `asyncio.to_thread` looks worthwhile on its own, but it would not fix this test, since the response still cannot go out until the shutdown returns.

Happy to send the test fix as its own PR if that helps, so this one stays on its topic.

echarles:
> Happy to send the test fix as its own PR if that helps, so this one stays on its topic.

Better to bake the fix in this PR to keep CI green.

ebarkhordar:
Done, baked in as 8cd2029.

```python
r = httpx.delete(
    f"{mcp_server_with_mcp_token}/api/stop",
    headers={"Authorization": f"Bearer {MCP_TOKEN}"},
    timeout=30.0,
)
```

Verified in a `python:3.10-slim` container this session: `REQUEST_TIMEOUT` is `10` and `inspect.iscoroutinefunction(KernelClient.stop)` is `False`, so the route can spend up to 10s on the shutdown before it responds, against a client that gave up at 5s (`httpx` default, confirmed `Timeout(timeout=5.0)`). A stand-in async route blocking 6s reproduces it exactly: `ReadTimeout` with the default client, `200` with `timeout=30.0`.

To be straight about the limit of that: I cannot make the shutdown cross 5s on demand, so the macOS leg going green here is the real check, not anything I ran. If it still fails, the cause is not the one I described and I would want to look again rather than raise the number.

`ruff check` reports the same findings on the file before and after, and the change survives `ruff-format` unchanged, so the lint leg should be unaffected.

The blocking-call-in-an-async-route point stands separately (a slow shutdown stalls `/api/healthz` too). Happy to open that as its own issue if it is worth tracking, since it needs `asyncio.to_thread` rather than a test change.

ebarkhordar:
Green on 8cd2029, including `test (macos-latest, 3.10)`. That leg was the real check on whether the inverted timeout was actually the cause, and it went green, so the diagnosis held.

Ready for review whenever it suits you.

echarles (approved):
LGTM

ebarkhordar:
Thanks for the review and the merge. Good call on baking the timeout fix in here: the `macos-latest, 3.10` leg going green is what confirmed the inverted timeout was the real cause.

PR #273, opened 2026-07-17
fix: preserve outputs, execution_count and metadata when moving a cell

### LINKED ISSUE #272: move_cell drops outputs, execution_count and metadata on the YDoc and WebSocket backends
## What happens

`move_cell` is a reorder, but on two of its three backends it does not carry the cell's payload along with it. Moving an executed code cell leaves it at the new index with `outputs` empty, `execution_count` back to `None`, `metadata` (including tags) emptied, and a freshly minted cell `id`.

`_move_cell_ydoc` and `_move_cell_websocket` delete the cell and re-create it from `(source, cell_type)` alone, so every other field is dropped. `_move_cell_file` reorders the cell object itself and keeps everything, so the three backends disagree about what a move means.

## Reproduction

Verified on HEAD 2113db05, in a clean `python:3.12-slim` container (non-editable `pip install .`; the imported module resolves to the installed copy).

Against a real Jupyter server and a real ipykernel, driving the MCP tools:

```
insert_cell(1, "code", "print('hello')")
execute_cell(1)          # cell 1 now has outputs and execution_count
move_cell(1, 2)
read_cell(2, include_outputs=True)
```

Before the move, `read_cell` reports `=====Cell 1 | type: code | execution count: 1=====` with a `hello` output. After the move, the same cell reads `=====Cell 2 | type: code | execution count: N/A=====` and the outputs region is empty.

The same loss reproduces as a unit differential on `_move_cell_websocket`: a cell with `outputs=[stream 'hello\n']`, `execution_count=7`, `metadata={'tags': ['keep-me']}` comes back with `outputs=[]`, `execution_count=None`, `metadata={}` and a different `id`.

The file backend is not affected. Run headless with no collaborative room open, the same MCP sequence in JUPYTER_SERVER mode falls back to `_move_cell_file` and the outputs and execution count both survive. I did not exercise `_move_cell_ydoc` itself: that needs a live collaborative room, which a headless run does not create. It carries the same five-line delete-and-recreate block as `_move_cell_websocket`, over the same `NotebookModel`, so I am reporting it from reading the code rather than from running it.

## Expected

A move changes a cell's position and nothing else: source, cell type, outputs, execution count, metadata and attachments all survive, identically on all three backends.

## Notes

`tests/test_move_cell.py::test_move_cell_preserves_outputs` already asserts this intent ("Moving a cell that has been executed should preserve its outputs"), which is why the defect is easy to miss: `read_cell` returns `[header, source, *outputs]` and the test asserts `"hello" in` the joined result, but the cell's source is literally `print('hello')`. The assertion matches the source, so it passes whether or not the outputs survived.

Happy to send a PR.

Reported by an AI agent (Claude), run and verified by @ebarkhordar.

ebarkhordar:
Fixes #272

## Root cause

`move_cell` has three backends implementing one contract, and two of them implement it as a re-creation rather than a reorder.

`_move_cell_ydoc` and `_move_cell_websocket` both delete the cell and rebuild it from `(source, cell_type)`:

```python
deleted = nb.delete_cell(source_index)
cell_type = deleted.get("cell_type", "code")
cell_source = deleted.get("source", "")
nb.insert_cell(target_index, cell_source, cell_type)
```

`NotebookModel.insert_cell(index, source, cell_type)` routes to `nbformat.v4.new_code_cell(source)`, so the new cell has only the two fields that were passed in. Everything else the `deleted` node still holds is dropped on the floor: `outputs`, `execution_count`, `metadata` and tags, attachments, and the cell `id`. `_move_cell_file` reorders the cell object itself (`_apply_move`), which is why the file backend was never affected.

## The invariant

A move changes a cell's position and nothing else: source, cell type, outputs, execution count, metadata and attachments survive, identically on all three backends.

The fails-on-main test below is an existence proof that the payload now survives the path it runs; it cannot establish "nothing else rebuilds a cell". That part is settled by enumeration instead. Parsing the package for calls to the lossy primitive `NotebookModel.insert_cell(...)` returns exactly four call sites:

| call site | verdict |
|---|---|
| `insert_cell_tool.py:85` (`_insert_cell_ydoc`) | correct: creating a genuinely new cell from source, nothing to preserve |
| `insert_cell_tool.py:177` (`_insert_cell_websocket`) | correct: same |
| `move_cell_tool.py:91` (`_move_cell_ydoc`) | the bug, fixed here |
| `move_cell_tool.py:152` (`_move_cell_websocket`) | the bug, fixed here |

`delete_cell(...)` has exactly two call sites in the package, both in `move_cell_tool.py`, both fixed here. So the complete set of places that could destroy a cell payload on a move is the two lines this PR changes.

## Fix

Reinsert the deleted cell through `NotebookModel.insert(index, value: dict)`, which hands the whole cell to `YNotebook.create_ycell(value)`. That path rebuilds source, metadata and outputs as CRDT types and keeps every other key, including an incoming `id`. It is 4 insertions and 4 deletions across the two call sites, with no public API change and no new dependency.

## Verification

Docker `python:3.12-slim`, non-editable `pip install .`, module provenance printed on both sides (`/usr/local/lib/python3.12/site-packages/jupyter_mcp_server/tools/move_cell_tool.py`), source file sha256 `4e50c846...` on main against `20e73f9d...` on this branch.

Live end-to-end against a real Jupyter server and a real ipykernel, on main, MCP_SERVER mode:

```
BEFORE read_cell(1): ['=====Cell 1 | type: code | execution count: 1=====', "print('hello')", 'hello\n']
AFTER  read_cell(2): ['=====Cell 2 | type: code | execution count: N/A=====', "print('hello')"]
```

On this branch the same sequence returns `execution count: 1` and the `hello` output at the new index.

The repo's own suite, both modes CI runs:

- `make test-mcp-server` equivalent: 39 passed
- `make test-jupyter-server` equivalent: 39 passed

Reverting only the source file and keeping the tests: `test_move_cell_preserves_outputs` and `TestMoveCellPreservesPayload` both fail in MCP_SERVER mode (`assert 'hello' in ''`, and `assert [] == [{...'text': 'hello\n'...}]`).

### The existing test, and why it passed

`test_move_cell_preserves_outputs` already encoded exactly this invariant, with the docstring "Moving a cell that has been executed should preserve its outputs." It could not fail. `read_cell` returns `[header, source, *outputs]`, the test joined the whole result and asserted `"hello" in cell_text`, and the cell's source is literally `print('hello')`. The assertion was matching the source, which a move never loses.

It now unpacks the outputs region and compares against a read taken before the move. The execution count is compared before against after rather than pinned to a literal: the kernel's counter depends on what ran earlier in the session, so a hardcoded `1` passes in MCP_SERVER mode and fails in JUPYTER_SERVER mode for reasons that have nothing to do with this bug.

`TestMoveCellPreservesPayload` is added alongside the file's existing Section A unit tests, covering metadata and tags and the cell `id`, which the end-to-end path does not surface. It mocks only the connection manager: `delete_cell` and `insert` are the real `NotebookModel` methods, so no writer of the cell payload is mocked out.

## What I did not verify

- **`_move_cell_ydoc` was not executed.** A headless run creates no collaborative room, so `get_notebook_model()` returns `None` and the tool correctly falls back to the file path. I exercised its identical twin `_move_cell_websocket` over the same `NotebookModel` type and read `_move_cell_ydoc` by eye. Both call sites carry the same five-line block, but I am reporting one from execution and one from reading.
- **Attachments** on markdown cells: read from the code (`new_markdown_cell(source)` takes no attachments), not measured.
- **The cell `id`.** Today a move re-mints it, and this fix preserves it, because `create_ycell` keeps an incoming `id`. I have treated that as correct on the grounds that a move should not change cell identity, and the added unit test pins it. If you would rather a moved cell take a fresh `id`, say so and I will drop that assertion and the behavior with it.
- Not a regression: `move_cell` has a single commit (c45d8ec, "Add move_cell tool (#226)") and has not been touched since, so this is original behavior rather than something that broke later.

I left `ruff format` alone deliberately: it wants to reformat 52 files on a pristine main, so running it here would have buried a four-line fix in an unrelated reformat. The changed lines follow the surrounding style.

---

This PR was prepared by an AI agent (Claude) operating on @ebarkhordar's behalf. Every command above was executed and its output read; the wording of what was and was not verified is deliberate.

### DIFF
--- a/jupyter_mcp_server/tools/move_cell_tool.py
+++ b/jupyter_mcp_server/tools/move_cell_tool.py
@@ -83,12 +83,12 @@ class MoveCellTool(BaseTool):
                 cell_type = nb_dict["cells"][source_index].get("cell_type", "code")
                 return Notebook(**nb_dict), {"cell_type": cell_type, "source": cell_source}
 
-            deleted = nb.delete_cell(source_index)
+            deleted = dict(nb.delete_cell(source_index))
             cell_type = deleted.get("cell_type", "code")
             cell_source = deleted.get("source", "")
             if isinstance(cell_source, list):
                 cell_source = "".join(cell_source)
-            nb.insert_cell(target_index, cell_source, cell_type)
+            nb.insert(target_index, deleted)
             return Notebook(**nb.as_dict()), {"cell_type": cell_type, "source": cell_source}
         else:
             return await self._move_cell_file(notebook_path, source_index, target_index)
@@ -144,12 +144,12 @@ class MoveCellTool(BaseTool):
                 cell_type = nb_dict["cells"][source_index].get("cell_type", "code")
                 return Notebook(**nb_dict), {"cell_type": cell_type, "source": cell_source}
 
-            deleted = notebook.delete_cell(source_index)
+            deleted = dict(notebook.delete_cell(source_index))
             cell_type = deleted.get("cell_type", "code")
             cell_source = deleted.get("source", "")
             if isinstance(cell_source, list):
                 cell_source = "".join(cell_source)
-            notebook.insert_cell(target_index, cell_source, cell_type)
+            notebook.insert(target_index, deleted)
             return Notebook(**notebook.as_dict()), {"cell_type": cell_type, "source": cell_source}
 
     async def execute(
--- a/tests/test_move_cell.py
+++ b/tests/test_move_cell.py
@@ -24,6 +24,8 @@ $ pytest tests/test_move_cell.py -v
 ```
 """
 
+import contextlib
+
 import pytest
 
 from .test_common import MCPClient, timeout_wrapper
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

PR #276, opened 2026-07-17
fix: return the timeout log and keep image outputs on the execute_cell streaming path

### LINKED ISSUE #275: execute_cell with stream=True discards the timeout log and replaces image outputs with an error string
## What happens

`execute_cell` with `stream=True` mishandles two cases at the end of a run. Both are specific to the streaming branch: the non-streaming branch of the same function gets each one right.

**1. A timeout throws away the timeout log it just built.** The monitoring loop cancels the execution task, appends `[TIMEOUT at ...s: Cancelling execution]` and `[Sent interrupt signal to kernel]`, then breaks. The final block is guarded by `if not execution_task.cancelled():` (execute_cell_tool.py:320). `Task.cancelled()` is still `False` immediately after `cancel()`, because the task has not processed the cancellation yet, so the guard passes and the `await execution_task` inside it raises `CancelledError`. `CancelledError` derives from `BaseException`, so the `except Exception` at :334 does not catch it. It propagates out of `execute()`: the caller gets an exception instead of the partial log, and the `AFTER_EXECUTE` hook never fires. The non-streaming branch returns partial outputs plus a `[TIMEOUT ERROR: ...]` marker (:371-381).

**2. An image output on the terminal drain path is replaced by an error string.** Outputs that land between the last one-second poll and completion are drained at :327-332 with `if extracted.strip():`. `extract_output` is declared `-> Union[str, ImageContent]` (utils.py:49) and returns `ImageContent` for `image/png` when `ALLOW_IMG_OUTPUT` is on, which is the default. `ImageContent` has no `.strip()`, so this raises `AttributeError`, which the `except Exception` at :334 swallows into `[ERROR: 'ImageContent' object has no attribute 'strip']`. The image is not just dropped, it is replaced by a string the caller cannot distinguish from a genuine execution error. The monitoring loop at :302-307 handles the same union correctly with `isinstance(extracted, str)`.

This is reached by a fast `plt.show()`, where the output lands after the final poll.

## Reproduction

Verified on HEAD 4d07938d, in a clean `python:3.12-slim` container (non-editable `pip install .`; the imported module resolves to `/usr/local/lib/python3.12/site-packages/jupyter_mcp_server/tools/execute_cell_tool.py`).

Driving the real `ExecuteCellTool.execute(mode=ServerMode.MCP_SERVER, stream=True, ...)` with in-memory stand-ins for the kernel and the notebook connection, so the control flow and the return value under test are the real code:

Case 2, a cell whose only output is an `image/png` that appears at completion:

```
RESULT: ['[PROGRESS: 0.0s elapsed, 0 outputs so far]',
         '[COMPLETED in 1.0s]',
         "[ERROR: 'ImageContent' object has no attribute 'strip']"]
image preserved? False
```

Case 1, a cell that runs past `timeout_seconds`:

```
*** RAISED CancelledError ***
timeout log returned? False
```

The primitives behind case 1, printed from the same container:

```
cancel() -> True
cancelled() immediately after cancel() -> False
await task RAISED: CancelledError
isinstance(CancelledError(), Exception) -> False
```

## Expected

A streaming timeout returns the partial log it collected, the way the non-streaming branch returns its `[TIMEOUT ERROR: ...]` marker, and fires `AFTER_EXECUTE`. An image output survives the terminal drain the same way it survives the monitoring loop.

## Notes

`grep -rn 'stream=True' tests/` returns nothing, which is why both have survived: the flag is user-facing (`server.py:618`) but has no test coverage. Both predate the three refactors that moved this code (#28 introduced the `cancelled()` guard, #95 and #111 moved it).

I did not drive a live kernel or a live RTC server. These are claims about `execute()`'s own control flow and return value, and the reproduction above does not need one.

Happy to send a PR.

Reported by an AI agent (Claude), run and verified by @ebarkhordar.

ebarkhordar:
Fixes #275.

## Root cause

Two defects in the terminal handling of `ExecuteCellTool.execute`'s `stream=True` branch. Neither is present in the non-streaming branch of the same function, which is what makes them look like oversights rather than intent.

**Timeout discards its own log.** The loop cancels the execution task, appends the timeout entries, and breaks. The final block is then guarded by `if not execution_task.cancelled():`. `Task.cancelled()` is `False` immediately after `cancel()`, because the task has not processed the cancellation yet, so the guard always passes and the `await execution_task` inside raises `CancelledError`. That derives from `BaseException`, so the `except Exception` below does not catch it: it escapes `execute()`, the timeout log is dropped and `AFTER_EXECUTE` never fires. The non-streaming branch instead returns partial outputs plus a `[TIMEOUT ERROR: ...]` marker. The `.cancelled()` call is dead code as written.

**The terminal drain assumes a str.** `if extracted.strip():` runs on the result of `extract_output`, declared `-> Union[str, ImageContent]`, which returns `ImageContent` for `image/png` when `ALLOW_IMG_OUTPUT` is on (the default). The `AttributeError` is swallowed by the `except Exception` below, so an image is replaced by `[ERROR: 'ImageContent' object has no attribute 'strip']`, which the caller cannot tell apart from a real execution error. The monitoring loop 28 lines above already type-checks the same union.

## Fix

Track the timeout in a local `timed_out` flag and gate the completion block on it, so the timeout path returns `outputs_log` like the non-streaming branch. Mirror the monitoring loop's `isinstance` check on the drain path so `ImageContent` passes through, keeping the existing empty-string filter for strings.

Nine lines, one function, no signature or behavior change on the paths that already worked.

## Verification

Both defects were reproduced on HEAD `4d07938d` and re-run against the branch, in a clean `python:3.12-slim` container, with the module path printed on both sides to prove the installed copy is the one under test.

Two regression tests in `tests/test_execute_cell_stream.py`, one per defect. They drive the real `execute()` with in-memory stand-ins for the kernel and the notebook connection, following the pattern in `tests/test_restart_notebook_tool.py`, so no live server or kernel is needed. Both **fail on main and pass on the branch**:

```
main:   2 failed
branch: 2 passed
```

Full suite, same command and container on both sides:

```
main:   1 failed, 102 passed, 1 skipped, 100 errors
branch: 1 failed, 104 passed, 1 skipped, 100 errors
```

The +2 are the new tests. The pre-existing failure and the 100 errors are identical on main and need a live JupyterLab that my container does not provide, so I am relying on CI for the integration legs. `license-eye header check` and `ruff` on the changed files are clean (the new test file adds only the `S101` asserts that every test file here has).

I did not drive a live kernel or a live RTC server. Both claims are about `execute()`'s own control flow and return value, and the tests do not need one.

Written with AI assistance (Claude), run and verified by @ebarkhordar.

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -277,6 +277,7 @@ class ExecuteCellTool(BaseTool):
 
                     start_time = time.time()
                     last_output_count = 0
+                    timed_out = False
 
                     # Monitor progress
                     while not execution_task.done():
@@ -285,6 +286,7 @@ class ExecuteCellTool(BaseTool):
                         # Check timeout
                         if elapsed > timeout_seconds:
                             execution_task.cancel()
+                            timed_out = True
                             outputs_log.append(f"[TIMEOUT at {elapsed:.1f}s: Cancelling execution]")
                             try:
                                 kernel.interrupt()
@@ -316,8 +318,10 @@ class ExecuteCellTool(BaseTool):
 
                         await asyncio.sleep(1)
 
-                    # Get final result
-                    if not execution_task.cancelled():
+                    # Get final result. On timeout the task was cancelled above, so
+                    # awaiting it would raise CancelledError, which is a BaseException
+                    # and is not caught below, discarding the timeout log just built.
+                    if not timed_out:
                         try:
                             await execution_task
                             final_outputs = notebook[cell_index].get("outputs", [])
@@ -328,7 +332,9 @@ class ExecuteCellTool(BaseTool):
                                 remaining = final_outputs[last_output_count:]
                                 for output in remaining:
                                     extracted = extract_output(output)
-                                    if extracted.strip():
+                                    if not isinstance(extracted, str):
+                                        outputs_log.append(extracted)
+                                    elif extracted.strip():
                                         outputs_log.append(extracted)
 
                         except Exception as e:
--- a/tests/test_execute_cell_stream.py
+++ b/tests/test_execute_cell_stream.py
@@ -1,0 +1,128 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Unit tests for ExecuteCellTool's streaming (stream=True) branch.
+
+These tests exercise the tool's own control flow with in-memory stand-ins for
+the kernel and the notebook connection, so no running Jupyter server or kernel
+is required.
+"""
+
+import contextlib
+import time
+
+import pytest
+
+from jupyter_mcp_server.tools._base import ServerMode
+from jupyter_mcp_server.tools.execute_cell_tool import ExecuteCellTool
+
+# 1x1 transparent PNG.
+PNG_B64 = (
+    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk"
+    "YPhfDwAChwGA60e6kgAAAABJRU5ErkJggg=="
+)
+
+
+class FakeKernel:
+    """Minimal stand-in covering the surface the streaming branch uses."""
+
+    def __init__(self):
+        self.interrupted = False
+
+    def interrupt(self):
+        self.interrupted = True
+
+
+class FakeNotebook:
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick turnaround @echarles. Good to have the streaming path returning its timeout log now, that was one more gap left over from the configurable-timeout work in #270.

echarles:
Awesome work @ebarkhordar Thank you so much!

PR #278, opened 2026-07-17
fix: keep each output's nbformat type when writing results back to a notebook file

### LINKED ISSUE #277: execute_cell in JUPYTER_SERVER file mode persists every kernel output as execute_result
## What happens

In JUPYTER_SERVER mode, when the notebook is not open in a collaborative session (no YDoc), `execute_cell` writes the results back to the `.ipynb` itself (`execute_cell_tool.py:245` calling `_write_outputs_to_cell`). The kernel reports every output with its nbformat type, but that type is dropped on the way in and then guessed back:

1. `execute_via_execution_stack` gets `result["outputs"]` from `ExecutionStack`, already nbformat-shaped, and returns only `safe_extract_outputs(...)`, which is a list of display strings (`utils.py:549`).
2. `_write_outputs_to_cell` re-derives the type from the string form: anything starting with `[ERROR:`, `[TIMEOUT ERROR:` or `[PROGRESS:` becomes `output_type: stream`, and everything else becomes `output_type: execute_result` (`execute_cell_tool.py:72-84`).

So a cell that prints has its stdout persisted as `execute_result`, and a cell that raises has its traceback persisted as `execute_result` as well, because `safe_extract_outputs` returns the bare traceback text and that carries no `[ERROR:` prefix. Nothing on this path ever writes `output_type: error`. Separately, the synthesized `execute_result` gets `execution_count: None` while the cell it belongs to gets `max + 1`, so the output and the cell disagree.

The notebook stays schema valid, so `nbformat.validate` never complains and nothing surfaces the problem. What breaks is downstream tooling that keys on `output_type`: nbconvert renders stdout as a result value, and anything scraping a notebook for failures (papermill-style error detection, CI notebook checks) sees zero errors in a notebook whose cells actually raised. It is silent wrong data rather than a crash.

The RTC path is not affected, since it syncs outputs through the YDoc and never reaches this function. This only bites headless use, which is the normal case for an agent executing a notebook that no human has open in a browser tab.

## Reproduction

On HEAD `8586b559`, in a clean `python:3.11-slim` container (`pip install -e .`; the imported module resolves to `/build/jupyter_mcp_server/__init__.py`, the copy under test).

A real `python3` kernel driven through the real `ExecutionStack`, then the real `safe_extract_outputs` and the real `_write_outputs_to_cell`, ending at the file on disk. Nothing is mocked:

```
cell source        kernel emitted   persisted in .ipynb  verdict
--------------------------------------------------------------------------
print('hello')     stream           execute_result       WRONG
1/0                error            execute_result       WRONG
2                  execute_result   execute_result       ok
--------------------------------------------------------------------------
outputs persisted with the wrong type: 2/3   (nbformat.validate passed for all)
```

What the kernel actually reports for those cells, printed from the same container, to show the types exist before they are discarded:

```
print('hello')                    -> output_type = stream          | name = stdout
1/0                               -> output_type = error           | ename = ZeroDivisionError
2                                 -> output_type = execute_result
sys.stderr.write('warn\n')        -> output_type = stream          | name = stderr
```

Note the `1/0` case arrives as an `error` output inside `result["outputs"]`, not through the `if "error" in result` branch at `utils.py:508`, so it reaches `safe_extract_outputs` and loses its type there.

## Expected

The `output_type` persisted to the notebook matches the one the kernel reported: `stream` stays `stream`, `error` stays `error` with its `ename`, `evalue` and `traceback`, and an `execute_result` reports the same execution count as its cell.

Reported with AI assistance (Claude), run and verified by @ebarkhordar.

ebarkhordar:
Fixes #277.

## Root cause

In JUPYTER_SERVER mode with the notebook not open in a collaborative session, `execute_cell` writes results straight to the `.ipynb`. The kernel reports each output with its nbformat type, and that type is thrown away and then guessed back:

- `execute_via_execution_stack` receives `result["outputs"]` from `ExecutionStack`, already nbformat-shaped, and returns only `safe_extract_outputs(...)`, a list of display strings.
- `_write_outputs_to_cell` re-derives the type from the string: `[ERROR:` / `[TIMEOUT ERROR:` / `[PROGRESS:` prefixes become `stream`, and everything else becomes `execute_result`.

A `print()` is therefore persisted as `execute_result`, and a traceback is persisted as `execute_result` too, since `safe_extract_outputs` hands back the bare traceback text with no `[ERROR:` prefix. Nothing on this path writes `output_type: error`. The synthesized `execute_result` also gets `execution_count: None` while its cell gets `max + 1`.

The notebook stays schema valid, so nothing complains. The damage is downstream, in tooling that keys on `output_type`: a notebook whose cells raised contains no `error` output at all. The RTC path is unaffected, it syncs through the YDoc and never calls this function.

## Fix

The structured outputs are already in hand one frame up, they are just discarded. `execute_via_execution_stack` takes an optional `raw_outputs` list and, when given one, appends the kernel's nbformat-shaped outputs to it. Its return value, and therefore the MCP tool's return value, is unchanged. `_write_outputs_to_cell` persists those outputs as they arrived instead of re-inflating strings, and gives each `execute_result` the execution count of its cell.

When no raw outputs are available (the timeout path reaches the write-back with strings only) the previous string form is kept, so that path behaves exactly as before.

`clean_notebook_outputs` moves to after the outputs are assigned, so it also strips `transient` from the kernel's own outputs. On the early-return paths it had no effect before, since the file is not written there.

## The invariant, and who else writes these outputs

Invariant: on the JUPYTER_SERVER file-mode `execute_cell` path, every output persisted to the `.ipynb` carries the `output_type` the kernel reported for it.

That quantifies over all writers of a cell's persisted `outputs`, which a passing test cannot check, so I enumerated them by parsing the package with `ast` (`X.outputs = ...`, subscript assignment and `append`/`extend`/`update`) and adjudicated each to its own type and path. Twenty candidates:

| writer | verdict |
| --- | --- |
| `execute_cell_tool._write_outputs_to_cell` (:63-84) | the fault site, and the only writer on this path. Fixed here. |
| `execute_cell_tool.py:384` `partial_outputs.append` | a local list of strings on the MCP_SERVER streaming path, not a cell field. |
| `utils.execute_cell_local` (:857, :860, :914, :918, :925) | same defect shape, but unreachable: zero callers in the package. Left alone. |
| `utils.execute_code_local` (:693-717) | builds correctly typed nbformat outputs from kernel messages. This is the shape the fix keeps. |
| `local_backend.append_cell` / `insert_cell` (:198, :249) | initialize `outputs` to `[]` on cell creation, not execution write-back. |
| `local_backend.execute_cell` (:414, :427) | a separate execution path that does not run under `ExecuteCellTool`'s JUPYTER_SERVER branch. Out of scope. |

The two self-labelled "simplified" comments on this behavior are in `execute_cell_local`, which is the dead one, so they do not describe the live path this changes. None of these writers is mocked in the tests below: the intersection of "writers" and "things the tests replace" is empty, because the tests mock nothing.

## Verification

Clean `python:3.11-slim` container, HEAD `8586b559`, module path printed on both sides so the copy under test is the installed one.

A real `python3` kernel driven through the real `ExecutionStack`, the real `safe_extract_outputs` and the real `_write_outputs_to_cell`, ending at the file on disk, with nothing mocked:

```
                   kernel emitted   persisted   main     branch
print('hello')     stream           ->          WRONG    ok
1/0                error            ->          WRONG    ok
2                  execute_result   ->          ok       ok

wrong type:  main 2/3    branch 0/3     (nbformat.validate passed on both sides)
```

Ten regression tests in `tests/test_execute_cell_output_fidelity.py`, one per behavior claimed above: the four output types round-tripping, `error` keeping `ename`/`evalue`/`traceback`, `stream` keeping `stderr` apart from `stdout`, `execute_result` agreeing with its cell's execution count, ordering across two outputs, `transient` being stripped, and the string fallback still working. They need no live server:

```
main:   9 failed, 1 passed
branch: 10 passed
```

The one that passes on both is the fallback test, which pins existing behavior. The other nine fail on main partly because `raw_outputs` does not exist there, so they pin the fix rather than prove the bug: the end-to-end run above is what demonstrates the bug on unmodified HEAD.

Pre-existing suites, same command and container on both sides:

```
main:   6 passed, 35 errors
branch: 6 passed, 35 errors
```

The 35 errors are identical on main and need a live JupyterLab my container does not provide, so I am relying on CI for those legs. `ruff` and `mypy` report the same counts on the two changed files before and after (0 and 22), the license header matches `.licenserc.yaml`, and the bare-install smoke test imports all 44 modules.

What I did not verify: I did not run nbconvert or papermill to watch them misread the notebook, so the specific rendering consequences are read from what `output_type` means, not measured. I did not stand up a full Jupyter Server with the extension loaded; the end-to-end run drives `ExecutionStack` directly with `ydoc_extension=None`, which is the file-mode configuration, but it is not a live server. I did not exercise the RTC path, which this does not touch.

Written with AI assistance (Claude), run and verified by @ebarkhordar.

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -9,7 +9,7 @@ import logging
 import time
 import nbformat
 from pathlib import Path
-from typing import Union, List
+from typing import Optional, Union, List
 from mcp.types import ImageContent
 
 from jupyter_mcp_server.hooks import HookEvent, HookRegistry
@@ -36,14 +36,19 @@ class ExecuteCellTool(BaseTool):
         self,
         notebook_path: str,
         cell_index: int,
-        outputs: List[Union[str, ImageContent]]
+        outputs: List[Union[str, ImageContent]],
+        raw_outputs: Optional[List[dict]] = None
     ):
-        """Write execution outputs back to a notebook cell."""
+        """Write execution outputs back to a notebook cell.
+
+        When raw_outputs is given, the kernel's own nbformat-shaped outputs are
+        persisted, which keeps each output's real output_type. Without them only
+        the formatted strings are available (the timeout path, for example), and
+        the output type can only be guessed from the string form.
+        """
 
         with open(notebook_path, 'r', encoding='utf-8') as f:
             notebook = nbformat.read(f, as_version=4)
-
-        clean_notebook_outputs(notebook)
 
         # Handle negative indices (e.g., -1 for last cell)
         num_cells = len(notebook.cells)
@@ -59,29 +64,38 @@ class ExecuteCellTool(BaseTool):
             logger.warning(f"Cell {cell_index} is not a code cell, cannot write outputs")
             return
 
-        # Convert formatted outputs to nbformat structure
-        cell.outputs = []
-        for output in outputs:
-            if isinstance(output, ImageContent):
-                cell.outputs.append(nbformat.v4.new_output(
-                    output_type='display_data',
-                    data={output.mimeType: output.data},
-                    metadata={}
-                ))
-            elif isinstance(output, str):
-                if output.startswith('[ERROR:') or output.startswith('[TIMEOUT ERROR:') or output.startswith('[PROGRESS:'):
+        if raw_outputs:
+            # The kernel already reported each output's type: keep it.
+            cell.outputs = [nbformat.from_dict(output) for output in raw_outputs]
+        else:
+            # Only the formatted strings are available, so the type has to be
+            # guessed from the string form.
+            cell.outputs = []
+            for output in outputs:
+                if isinstance(output, ImageContent):
                     cell.outputs.append(nbformat.v4.new_output(
-                        output_type='stream',
-                        name='stdout',
-                        text=output
+                        output_type='display_data',
+                        data={output.mimeType: output.data},
+                        metadata={}
                     ))
-                else:
-                    cell.outputs.append(nbformat.v4.new_output(
-                        output_type='execute_result',
-                        data={'text/plain': output},
-                        metadata={},
-                        execution_count=None
-                    ))
+                elif isinstance(output, str):
+                    if output.startswith('[ERROR:') or output.startswith('[TIMEOUT ERROR:') or output.startswith('[PROGRESS:'):
+                        cell.outputs.append(nbformat.v4.new_output(
+                            output_type='stream',
+                            name='stdout',
+                            text=output
+                        ))
+                    else:
+                        cell.outputs.append(nbformat.v4.new_output(
+                            output_type='execute_result',
+                            data={'text/plain': output},
+                            metadata={},
+                            execution_count=None
+                        ))
+
+        # Strip kernel-protocol fields that are not part of the nbformat schema
+        # (the raw outputs above come straight from the kernel).
+        clean_notebook_outputs(notebook)
 
         # Update execution count
         max_count = 0
@@ -89,6 +103,11 @@ class ExecuteCellTool(BaseTool):
             if c.cell_type == 'code' and c.execution_count:
                 max_count = max(max_count, c.execution_count)
         cell.execution_count = max_count + 1
+
+        # An execute_result carries the execution count of the cell that produced it.
+        for output in cell.outputs:
+            if output.get('output_type') == 'execute_result':
+                output['execution_count'] = cell.execution_count
 
         with open(notebook_path, 'w', encoding='utf-8') as f:
             nbformat.write(notebook, f)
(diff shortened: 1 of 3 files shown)


echarles:
can you bump jupyter-nbmodel-client in the deps, I think we need the latest release.

echarles (approved):
LGTM

echarles:
> can you bump jupyter-nbmodel-client in the deps, I think we need the latest release.

not needed for this PR, reviewed and merging now.

ebarkhordar:
Thanks for the quick review and the merge. If you do want the jupyter-nbmodel-client floor raised, happy to send that as its own PR.

echarles:
yes please, I will then cut a new release.

PR #280, opened 2026-07-17
fix: honour execute_code kernel_id in MCP_SERVER mode

### LINKED ISSUE #279: execute_code(kernel_id=...) is ignored in MCP_SERVER mode and runs on the current notebook's kernel
## What happens

`execute_code` accepts a `kernel_id` parameter, documented as "Target an existing kernel by ID (e.g. a raw kernel with no notebook). If omitted, uses the current notebook's kernel." In MCP_SERVER mode, which is the default for the standalone `uvx jupyter-mcp-server` deployment, that argument has no effect at all.

`server.py:799` resolves `kernel_id` only when `server_context.mode == ServerMode.JUPYTER_SERVER`, but a user-supplied `kernel_id` is still passed down to the tool in either mode. `ExecuteCodeTool.execute` then routes MCP_SERVER to `_execute_via_notebook_manager` (`execute_code_tool.py:214`), whose signature has no `kernel_id` parameter (`execute_code_tool.py:54-62`). The argument is dropped, the code runs on the current notebook's kernel, and that kernel's output is returned as a successful result.

The failure is silent rather than loud. There is no error, and the caller cannot tell from the response that it reached a different kernel than the one it asked for. For a read-only cell it returns another kernel's state; for a state-mutating cell it mutates the wrong kernel.

The JUPYTER_SERVER branch of the same function honours `kernel_id` fully, 40 lines up, so the two modes disagree about what the parameter means.

## Reproduction

On HEAD `982b7ae`, in a clean `python:3.11-slim` container. Real `jupyter lab`, two real kernels, the real `ExecuteCodeTool`, real `NotebookManager`, real `wait_for_kernel_idle` and `safe_extract_outputs`. Nothing is mocked, and the imported module resolves to the copy under test, printed as provenance.

Each kernel is given its own `MARKER` variable, the first is bound to the current notebook, and the second is a raw kernel with no notebook attached. `execute_code` is then asked to run `print(MARKER)` in the raw one:

```
provenance: /src/jupyter_mcp_server/tools/execute_code_tool.py
current notebook's kernel : d516abb3-6bcb-4db6-b447-81137cdfd8db  (MARKER='current-notebook-kernel')
raw kernel, no notebook   : cb6c7852-eca2-4b13-9de8-08af36422309  (MARKER='raw-kernel')

execute_code(code="print(MARKER)", kernel_id="cb6c7852-eca2-4b13-9de8-08af36422309")

  returned : 'current-notebook-kernel'
  expected : 'raw-kernel'
  verdict  : WRONG KERNEL
```

The raw kernel executed nothing. The current notebook's kernel executed the code, and its output came back as success.

An unknown `kernel_id` behaves the same way: it is discarded and the code runs on the current notebook's kernel, so a typo in the id also reports success from the wrong place.

## Expected

`kernel_id` selects the kernel the code runs in, in both modes, as the docs and the parameter description state. An id that does not name a kernel on the server is reported rather than silently redirected.

The mechanism already exists in this mode: `use_notebook_tool.py:215-226` validates a kernel id against `server_client.kernels.list_kernels()` and connects with `KernelClient(server_url=..., token=..., kernel_id=...)` in MCP_SERVER mode today.

I have a fix and tests ready, and will open a PR against this issue.

ebarkhordar:
Fixes #279

## Root cause

`server.py:799` resolves `kernel_id` only in JUPYTER_SERVER mode, but a user-supplied `kernel_id` is passed to the tool in either mode. `ExecuteCodeTool.execute` routes MCP_SERVER to `_execute_via_notebook_manager`, which had no `kernel_id` parameter, so the argument was dropped and the code always ran on the current notebook's kernel. The wrong kernel's output was then returned as success, with no error for the caller to notice. MCP_SERVER is the default mode, and the docs publish `kernel_id` without qualifying it by mode.

## Fix

`kernel_id` is threaded through to the notebook-manager path. When it names a kernel other than the current notebook's, the tool validates it against `server_client.kernels.list_kernels()` and connects with the same `KernelClient(server_url=..., token=..., kernel_id=...)` pattern `use_notebook_tool.py:215-226` already uses in this mode. An unknown id now returns the same "not found, check `list_kernels`" message `use_notebook` gives, instead of silently redirecting to whichever kernel is current. Omitting `kernel_id` is unchanged.

Such a kernel is borrowed, not owned, so it is released in a `finally` with an explicit `stop(shutdown_kernel=False)` and never shut down. The execution body moved into `_execute_on_kernel` unchanged so that both paths share it, which is why the diff is larger than the behaviour change.

## Invariant, and how it is discharged

The claim is that a borrowed kernel outlives the call. A passing test only shows that this path does not kill it, so I enumerated every writer of kernel liveness in the package by parsing rather than grepping, and adjudicated each:

| writer | reaches a borrowed kernel? |
| --- | --- |
| `local_backend.py:450` `shutdown_kernel()` | no, JUPYTER_SERVER mode only |
| `notebook_manager.py:143` `kernel.stop()` (shuts down if `_own_kernel`) | no, only kernels registered via `add_notebook`; this path never registers one |
| `execute_code_tool.py` (this diff) | yes, explicit `shutdown_kernel=False` |
| `utils.py:279` `kernel.stop()` | no, `create_kernel` failure cleanup for a kernel it just created |

`KernelClient._own_kernel` is `self._manager.kernel is None`, so connecting to an existing kernel already leaves it false. The explicit `shutdown_kernel=False` does not rely on that.

## Verification

Clean `python:3.11-slim` container, against the real `jupyter lab` fixture with two real kernels. Nothing is mocked: real `ExecuteCodeTool`, real `NotebookManager`, real `KernelClient`, real `wait_for_kernel_idle` and `safe_extract_outputs`.

`tests/test_execute_code_kernel_id.py` adds four tests, one per behaviour claimed above:

| test | on main | on this branch |
| --- | --- | --- |
| runs in the requested kernel | fail | pass |
| targeted kernel left running, state intact | fail | pass |
| unknown kernel_id reported, not redirected | fail | pass |
| no kernel_id still uses the current notebook | pass | pass |

The last one passes on both by design; it pins the default against regression.

```
python -m pytest tests/test_execute_code_kernel_id.py -q
```

Existing suites stay green in the same container: `test_tools.py`, `test_execute_cell_stream.py`, `test_execute_cell_output_fidelity.py` (51 passed), plus `test_hooks.py`, `test_otel_hooks.py`, `test_config.py`, `test_common.py`, `test_use_notebook_create_local.py`, `test_restart_notebook_tool.py`, `test_move_cell.py`, `test_edit_cell_source.py`, `test_prompts.py`, `test_use_notebook_ui_open.py` (145 passed, 1 skipped). `license-eye header check` passes on the new file, and the sdist and wheel build.

## What I did not verify

The tests target a raw kernel on the same local server the fixture starts. I did not run this against a remote `runtime_url` deployment, which is the setup the feature is aimed at, so the connection there is covered by the shared `KernelClient` path and not by an executed test.

`BEFORE_EXECUTE` and `AFTER_EXECUTE` hooks now fire with the targeted kernel's id rather than the current notebook's, which is what the hook payload claims to carry. `test_hooks.py` and `test_otel_hooks.py` pass, but neither exercises hooks with a targeted `kernel_id`, so that combination is untested.

### DIFF
--- a/jupyter_mcp_server/tools/execute_code_tool.py
+++ b/jupyter_mcp_server/tools/execute_code_tool.py
@@ -51,6 +51,33 @@ class ExecuteCodeTool(BaseTool):
             logger=logger
         )
     
+    def _connect_to_kernel(self, kernel_id: str, server_client):
+        """Connect to an existing kernel by ID (MCP_SERVER mode).
+
+        Returns (kernel, None) on success, or (None, error_message) if the id
+        does not name a kernel on the server.
+        """
+        from jupyter_kernel_client import KernelClient
+
+        from jupyter_mcp_server.config import get_config
+
+        if server_client is not None:
+            kernels = server_client.kernels.list_kernels()
+            if not any(kernel.id == kernel_id for kernel in kernels):
+                return None, (
+                    f"[ERROR: Kernel '{kernel_id}' not found in jupyter server, please check "
+                    f"whether the kernel already exists using 'list_kernels' tool.]"
+                )
+
+        config = get_config()
+        kernel = KernelClient(
+            server_url=config.runtime_url,
+            token=config.runtime_token,
+            kernel_id=kernel_id,
+        )
+        kernel.start()
+        return kernel, None
+
     async def _execute_via_notebook_manager(
         self,
         notebook_manager: NotebookManager,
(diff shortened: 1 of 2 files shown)


ebarkhordar:
The red `test (ubuntu-latest, 3.13)` leg is my own new test, so let me say what I found and what I did not.

I could not reproduce it. On `ac414e7`, the commit that leg failed on, the test passes 5/5 standalone in a 2-core container, and the whole file passes 4/4 with the change below. So I do not have a verified cause for that leg and I am not going to call it flaky.

What I did find is that the fixture never checked its own premise. It creates two kernels and seeds `MARKER` into each immediately, but `start()` returns while the server can still report the kernel as `starting`, and nothing confirmed the seeding ran. On the failing leg the server was cold, since it starts lazily on the first test that needs it and that test is the first one to use it. `4bf1eed` waits for the server's own view of each kernel to reach `idle` and then asserts the seeding succeeded, so a kernel that is not up fails in setup instead of as a 30s timeout inside the assertion.

Two things I measured while looking. Both are pre-existing and neither comes from this PR.

`wait_for_kernel_idle` never waits. `is_kernel_busy` (`utils.py:369`) probes `kernel._client`, and `KernelClient` has no `_client` attribute, only `_manager`, so it returns `False` unconditionally. Measured against a kernel sitting in `time.sleep(8)`: `is_kernel_busy` returns `False`, and `wait_for_kernel_idle(kernel, max_wait_seconds=30)` returns in 0.001s. It is the only readiness gate before `execute`, which matters because `execute_interactive` computes a deadline but never raises on it, so a missed reply is bounded only by the caller's own `asyncio.wait_for`. I have left it alone: making it actually wait would add real waiting to every execute path in the repo, and that is your call rather than mine.

`KernelClient.stop()` on a borrowed kernel costs a flat 10s. `stop_channels()` joins the websocket thread with `REQUEST_TIMEOUT` and the thread does not exit, so it burns the full join every time and logs `Failed to stop websocket connection thread`. That is 10s on each `execute_code(kernel_id=...)` call. It looks like a `jupyter_kernel_client` issue rather than something to work around here, so I have not touched it, but I am happy to file it upstream.

echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick review and merge. Happy to follow up if anything related to the kernel_id path turns up.

PR #281, opened 2026-07-17
fix: support synchronous ContentsManager via ensure_async

mucan54:
jupyter-server supports both synchronous and asynchronous ContentsManager implementations, but the JUPYTER_SERVER-mode helpers await contents_manager calls directly. With a synchronous manager those calls return plain values, so awaiting them raises:

    TypeError: object dict can't be used in 'await' expression

This is not hypothetical: jupytext.TextFileContentsManager resolves to SyncJupytextContentsManager, and setting

    c.ServerApp.contents_manager_class = 'jupytext.TextFileContentsManager'

is a common deployment. It breaks use_notebook, list_files, read_cells and read_notebook. list_files fails silently -- the except branch swallows the TypeError and returns an empty list, so the tool reports 'No files found' while the server's own /api/contents lists the directory fine.

Wrap the seven contents_manager calls in jupyter_core.utils.ensure_async, which awaits awaitables and passes plain values through. This is what jupyter-server's own handlers do and is a no-op for async managers.

Adds regression tests parametrized over a sync and an async manager.

### DIFF
--- a/jupyter_mcp_server/tools/jupyter_cite_prompt.py
+++ b/jupyter_mcp_server/tools/jupyter_cite_prompt.py
@@ -10,6 +10,7 @@ from jupyter_server_client import JupyterServerClient
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 from jupyter_mcp_server.notebook_manager import NotebookManager
 from jupyter_mcp_server.models import Notebook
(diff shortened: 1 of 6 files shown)


echarles (approved):
LGTM Thx @mucan54

PR #282, opened 2026-07-18
fix: apply the execution timeout config to execute_code

ebarkhordar:
## Root cause

Three tools run code with a `timeout`: `execute_cell`, `insert_execute_code_cell`, and `execute_code`. The first two resolve the timeout against the config before running:

```python
config = get_config()
effective_timeout = config.execution_timeout if timeout == 0 else min(timeout, config.max_execution_timeout)
```

`execute_code` did neither. It forwarded its raw `timeout` straight to the tool (`server.py:810`), with a hardcoded `Field(le=3600)` as its only bound. Two consequences:

- **`max_execution_timeout` does not bind.** An operator who lowers the ceiling (`--max-execution-timeout 60`) still has it enforced on `execute_cell` and `insert_execute_code_cell`, but `execute_code` accepts and passes through any value up to 3600, so the ceiling is silently unenforced on one of the three.
- **`timeout=0` is not the config default here.** The other two tools document `0` as "use config default"; `execute_code` forwarded `0` verbatim to `asyncio.wait_for(timeout=0)` (`execute_code_tool.py:165`), which expires immediately.

The `le=3600` cap is a stale duplicate: it was raised to 3600 in #257, thirteen days before the config landed (#266), and was never reconciled with `max_execution_timeout` (whose own default is 3600).

## Fix

Resolve `effective_timeout` in `execute_code` with the same expression the other two tools use, and drop `le=3600` in favour of the config ceiling. Three lines, no new behaviour beyond aligning the contract:

- `timeout=0` now resolves to `config.execution_timeout` instead of expiring at once.
- a positive `timeout` is clamped to `config.max_execution_timeout`.
- a positive `timeout` under the ceiling is unchanged.

## The one thing I did not change: the default

`execute_code` still defaults to `timeout=30`, while its two siblings default to `timeout=0` (which now means `execution_timeout`, 120 by default). Aligning it would change the out-of-the-box `execute_code` timeout from 30s to 120s. That is a public-behaviour change and your call, so I left it. Happy to switch the default to `0` in this PR if you would rather the three tools match exactly.

Context: this follows the "Left out on purpose" note in #270 and the discussion on #264, where the scope of the default was flagged as yours to decide.

## Verification

Run in a clean `python:3.12-slim` container against this branch (`server.__file__` confirmed to be the installed source).

New `tests/test_execute_code_timeout.py` captures the timeout that reaches `ExecuteCodeTool`, so the resolution is observable without a kernel. With `execution_timeout=45`, `max_execution_timeout=60`:

- `timeout=3600` resolves to `60` (clamped). Fails on `main` (passes 3600 through).
- `timeout=0` resolves to `45`. Fails on `main` (passes 0 through).
- `timeout=30` resolves to `30`, and omitting `timeout` keeps `30`. Unchanged on `main`.

All four pass on the branch; the first two fail on `main`, so the test pins the fix.

What I did not do: this is verified at the resolution boundary, not driven end-to-end through a live kernel. The change only alters the timeout value handed to `ExecuteCodeTool`; the tool and its kernel routing are untouched and already covered by `test_execute_code_kernel_id.py`. That `asyncio.wait_for(timeout=0)` raises immediately is confirmed on stock asyncio this session, not observed through a kernel.

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -776,7 +776,7 @@ async def move_cell(
 @with_hooks("execute_code")
 async def execute_code(
     code: Annotated[str, Field(description="Code to execute (supports magic commands with %, shell commands with !)")],
-    timeout: Annotated[int, Field(description="Execution timeout in seconds",le=3600)] = 30,
+    timeout: Annotated[int, Field(description="Maximum seconds to wait for execution (0 = use config default)")] = 30,
     kernel_id: Annotated[str | None, Field(description="Target an existing kernel by ID (e.g. a raw kernel with no notebook). If omitted, uses the current notebook's kernel.")] = None,
 ) -> Annotated[list[str | ImageContent], Field(description="List of outputs from the executed code")]:
     """Execute code directly in a kernel (not saved to notebook).
@@ -796,6 +796,10 @@ async def execute_code(
     1. Import new modules or perform variable assignments that affect subsequent Notebook execution
     2. Execute dangerous code that may harm the Jupyter server or the user's data without permission
     """
+    config = get_config()
+    # Use config default if timeout is 0, otherwise clamp to max
+    effective_timeout = config.execution_timeout if timeout == 0 else min(timeout, config.max_execution_timeout)
+
     if kernel_id is None and server_context.mode == ServerMode.JUPYTER_SERVER:
         current_notebook = notebook_manager.get_current_notebook() or "default"
         kernel_id = notebook_manager.get_kernel_id(current_notebook)
@@ -807,7 +811,7 @@ async def execute_code(
             kernel_manager=server_context.kernel_manager,
             notebook_manager=notebook_manager,
             code=code,
-            timeout=timeout,
+            timeout=effective_timeout,
             kernel_id=kernel_id,
             ensure_kernel_alive_fn=__ensure_kernel_alive,
             wait_for_kernel_idle_fn=wait_for_kernel_idle,
--- a/tests/test_execute_code_timeout.py
+++ b/tests/test_execute_code_timeout.py
@@ -1,0 +1,75 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Unit tests for execute_code's timeout resolution in server.py.
+
+execute_cell and insert_execute_code_cell resolve their ``timeout`` against the
+config before running: ``0`` means "use config default" (``execution_timeout``)
+and any positive value is clamped to ``max_execution_timeout``. execute_code is
+the third execution tool and did neither. These tests pin it to the same
+contract by capturing the timeout that reaches ExecuteCodeTool, so no kernel is
+required.
+"""
+
+import pytest
+
+from jupyter_mcp_server import server
+from jupyter_mcp_server.config import reset_config, set_config
+from jupyter_mcp_server.tools.execute_code_tool import ExecuteCodeTool
+
+
+@pytest.fixture
+def captured_timeout(monkeypatch):
+    """Record the timeout ExecuteCodeTool receives, so the value server.py
+    resolves is observable without a running kernel. A low ceiling and a
+    distinct default keep every assertion unambiguous."""
+    seen = {}
+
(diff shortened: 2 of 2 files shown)


echarles:
> ERROR tests/test_move_cell.py::test_move_cell_basic_forward[mcp_server] - Failed: Jupyter MCP failed to start after max retries. Port 42047 may be in use or server crashed.

Not sure if related to these changes?

ebarkhordar:
Thanks for checking. It is not from this change. The only non-pass is tests/test_move_cell.py::test_move_cell_basic_forward, reported as an ERROR at server startup ("Port 42047 may be in use or server crashed"), not a test assertion, and that file is unrelated to this PR. The diff only touches the execute_code timeout path in server.py plus a new test file, and the same commit is green on the other five matrix legs (ubuntu 3.10, macos 3.10/3.13, windows 3.10/3.13). tests/test_tools.py::test_execute_code, which exercises the code I changed, passed on that same 3.13 run. Happy to help dig in if that startup error recurs on a re-run of the ubuntu 3.13 leg.

echarles:
Relaunched the job, and now failing with another error. We may have introduced a regression in the latest PR (maybe [here](https://github.com/datalayer/jupyter-mcp-server/pull/280#issuecomment-5003871272) ? ). I remember we had to relaunch a test and the CI was succeeding after, but finally we may have to face flaky ci now. Before moving to anything else, we should solve this flakyness.

```
FAILED tests/test_execute_code_kernel_id.py::test_execute_code_leaves_the_targeted_kernel_running - AssertionError: assert 'raw-kernel' in '[TIMEOUT ERROR: IPython execution exceeded 30 seconds and was interrupted]'
 +  where '[TIMEOUT ERROR: IPython execution exceeded 30 seconds and was interrupted]' = <built-in method join of str object at 0x7fd86dd10a40>(<generator object test_execute_code_leaves_the_targeted_kernel_running.<locals>.<genexpr> at 0x7fd867514a00>)
 +    where <built-in method join of str object at 0x7fd86dd10a40> = ''.join
```

echarles:
We also had an failing test here https://github.com/datalayer/jupyter-mcp-server/pull/269#issuecomment-4988992630

ebarkhordar:
On #282: this red leg is not from this PR. The test calls execute_code with timeout=30 (tests/test_execute_code_kernel_id.py:89), and this PR only changes the resolved timeout when the caller passes 0 (use the config default) or a value above max_execution_timeout=3600 (clamp down). min(30, 3600) is 30, the exact value that reached the tool before this change, so #282 is a no-op on this test's timeout.

Both the failing test and the borrowed-kernel path it runs came in with #280, not here, so this is separate from #282's timeout resolution.

Whether it is a real #280 regression or CI load, I can't tell from the logs alone and I don't want to guess. What I see: the two red runs failed in unrelated files. The first was test_move_cell failing at server startup with "Port 42047 may be in use", which is port contention, not a logic fault. The second was this 30s timeout on a trivial print(MARKER), meaning the kernel never returned a result in time. Those are consistent with a resource-starved runner, but also with the borrowed-kernel path (a fresh KernelClient started per call) losing the reply under load, and I have not separated the two.

I'll look into the #280 connect-then-execute path for a real race and open a separate PR if I find one. #282 is independent of that, so it can merge on its own whenever you are happy with it.

echarles:
Launched again and now CI is screen - flaky ci was maybe already there before and surfacing from time to time? Thoughts?

echarles (approved):
LGTM Thx @ebarkhordar

echarles:
> I'll look into the https://github.com/datalayer/jupyter-mcp-server/pull/280 connect-then-execute path for a real race and open a separate PR if I find one. https://github.com/datalayer/jupyter-mcp-server/pull/282 is independent of that, so it can merge on its own whenever you are happy with it.

I have merged, will be great to identify and fix the cause of those failures in a separate PR

ebarkhordar:
Thanks, and glad it landed. I'll look into the #280 connect-then-execute path (a fresh KernelClient per call is my first suspect for dropping the reply under load) and open a separate PR if I find a real race. Will report back either way.

PR #284, opened 2026-07-18
fix: await LocalBackend contents_manager calls with ensure_async (#283)

### LINKED ISSUE #283: LocalBackend list_notebooks silently returns [] on a default (async) ContentsManager: to_thread returns an un-awaited coroutine (mirror of #281)
## Summary

#281 fixed the `ContentsManager` sync/async mismatch (`ensure_async`) on the tool-side files, but the same defect is still latent in `jupyter_extension/backends/local_backend.py`, which that PR did not touch.

`LocalBackend` wraps every `contents_manager` call as `await asyncio.to_thread(self.contents_manager.get/new/save, ...)`. `asyncio.to_thread` runs the callable in a worker thread and returns its result. For jupyter_server's default `AsyncLargeFileManager`, `get`/`new`/`save` are coroutine functions, so the thread returns an un-awaited coroutine object instead of a model dict, and the next line subscripts it.

## Verified mechanism

Clean container (`python:3.12-slim`), `jupyter_server 2.20.0`, on current `main` (2d50432):

```python
import asyncio, os, tempfile, nbformat
from jupyter_server.serverapp import ServerApp
from jupyter_server.services.contents.largefilemanager import AsyncLargeFileManager
from jupyter_core.utils import ensure_async

# ServerApp's default is the ASYNC manager
print(ServerApp.contents_manager_class.default_value)  # AsyncLargeFileManager

d = tempfile.mkdtemp()
nbformat.write(nbformat.v4.new_notebook(), open(os.path.join(d, "test.ipynb"), "w"))
cm = AsyncLargeFileManager(root_dir=d)

async def main():
    # local_backend.py:59-65 pattern
    model = await asyncio.to_thread(cm.get, "test.ipynb", type="notebook", content=True)
    print(type(model).__name__)   # coroutine
    model["content"]              # TypeError: 'coroutine' object is not subscriptable

asyncio.run(main())
```

Output:

```
<class '...largefilemanager.AsyncLargeFileManager'>
coroutine
TypeError: 'coroutine' object is not subscriptable
```

`await ensure_async(cm.get("test.ipynb", type="notebook", content=True))` returns the model dict as expected (the pattern #281 established).

## Reachable impact today (from reading the code, not run end to end)

The one tool wired through `LocalBackend` right now is `list_notebooks` (`handlers.py._execute_tool` is otherwise a `# TODO: Implement actual tool routing` stub). `list_notebooks` calls `_list_notebooks_recursive`, whose `to_thread(cm.get)` sits inside `except Exception: pass` (local_backend.py:98-99), so on a default (async) install it silently returns `[]` instead of the notebook list, with no error surfaced. The `get_notebook_content` / `create_notebook` / save paths (lines 59, 132, 205, 256, 282, 327, 428) would raise the `TypeError` above on the same manager once they are wired up.

## Question before I send a PR

Is `jupyter_extension/backends/LocalBackend` meant to be a reachable/production surface, or is it still experimental while the tool-routing stub is filled in? If it is meant to be reachable, I am happy to send a PR mirroring #281: import `ensure_async` and wrap the 9 `contents_manager` calls (leaving the genuinely sync `client.get_iopub_msg` at :404 alone), with a test parametrized over a sync and an async `ContentsManager` (same shape as `tests/test_contents_manager_sync.py`). If it is experimental, feel free to close.

ebarkhordar:
## Root cause

`LocalBackend` reaches the notebook file system through `serverapp.contents_manager`, wrapping every call in `asyncio.to_thread(self.contents_manager.<method>, ...)`. `asyncio.to_thread` runs the callable in a worker thread and returns its result. jupyter-server's default contents manager, `AsyncLargeFileManager`, defines `get`/`new`/`save` as coroutine functions, so the worker thread returns an un-awaited coroutine object rather than a model dict.

On a default (async) install the caller then mishandles that coroutine:

- `_list_notebooks_recursive` subscripts it (`model['type']`), the bare `except Exception` swallows the resulting `TypeError`, and `list_notebooks` silently returns `[]`.
- `get_notebook_content`, `create_notebook`, and the save paths raise `TypeError: 'coroutine' object is not subscriptable` or leave the coroutine un-awaited.
- `notebook_exists` reports every path as present, because the un-awaited coroutine is truthy and the lookup body never runs.

This is the `LocalBackend` mirror of the tool-side bug fixed in #281. #281 corrected `await contents_manager.get(...)` (right for async, broken for sync); this backend has the opposite shape, `to_thread(contents_manager.get, ...)` (right for sync, broken for the default async).

## Fix

Wrap the nine `contents_manager` calls (`get`/`new`/`save`, across the read, list, exists, create, append, insert, delete, overwrite, and execute-save paths) in `jupyter_core.utils.ensure_async`, which awaits awaitables and passes plain values through, so both flavours work. This is the same pattern #281 applied to the tool files. The one genuinely-synchronous `to_thread` call, `client.get_iopub_msg`, is left as is.

Invariant: every `contents_manager` call in `LocalBackend` is awaited correctly for both a synchronous and an asynchronous ContentsManager. I enumerated the call sites by parsing the module (nine `self.contents_manager.{get,new,save}` calls, no other `contents_manager` access); all nine are wrapped.

## Verification

Run in a clean `python:3.12-slim` container against current HEAD, `jupyter_server==2.20.0`, `jupyter_core==5.9.1`, with the default `contents_manager` resolving to `AsyncLargeFileManager`:

- Added `tests/test_local_backend_contents_manager.py`, parametrized over a synchronous and an asynchronous ContentsManager (same shape as `tests/test_contents_manager_sync.py`), covering `get_notebook_content`, `list_notebooks`, `notebook_exists` (present and absent), `create_notebook`, and the get+save round trip through `append_cell`.
- On `main` the five async cases fail: `get_notebook_content`/`create_notebook`/`append_cell` raise `TypeError`, `list_notebooks` returns `[]`, and `notebook_exists` reports a missing path as present. With the fix all ten pass.
- `tests/test_contents_manager_sync.py` (#281) still passes, so the tool-side path is unaffected.

Not verified: I did not run the manual MCP-client end-to-end flow from CONTRIBUTING. The save paths other than `append_cell` (`insert_cell`/`delete_cell`/`overwrite_cell`/`execute_cell`) use the identical wrapped call and are covered by inspection, not each exercised individually.

Fixes #283.

### DIFF
--- a/jupyter_mcp_server/jupyter_extension/backends/local_backend.py
+++ b/jupyter_mcp_server/jupyter_extension/backends/local_backend.py
@@ -14,6 +14,7 @@ import asyncio
 from mcp.types import ImageContent
 from jupyter_mcp_server.jupyter_extension.backends.base import Backend
 from jupyter_mcp_server.utils import safe_extract_outputs
+from jupyter_core.utils import ensure_async
 
 if TYPE_CHECKING:
     from jupyter_server.serverapp import ServerApp
@@ -56,11 +57,8 @@ class LocalBackend(Backend):
         Returns:
             Notebook content dictionary
         """
-        model = await asyncio.to_thread(
-            self.contents_manager.get,
-            path,
-            type='notebook',
-            content=True
+        model = await ensure_async(
+            self.contents_manager.get(path, type='notebook', content=True)
         )
         return model['content']
     
@@ -81,12 +79,10 @@ class LocalBackend(Backend):
     async def _list_notebooks_recursive(self, path: str, notebooks: list[str]) -> None:
         """Helper to recursively list notebooks."""
         try:
-            model = await asyncio.to_thread(
-                self.contents_manager.get,
-                path,
-                content=True
+            model = await ensure_async(
+                self.contents_manager.get(path, content=True)
             )
-            
+
             if model['type'] == 'directory':
                 for item in model['content']:
                     item_path = f"{path}/{item['name']}" if path else item['name']
@@ -110,10 +106,8 @@ class LocalBackend(Backend):
             True if exists
         """
         try:
-            await asyncio.to_thread(
-                self.contents_manager.get,
-                path,
-                content=False
+            await ensure_async(
+                self.contents_manager.get(path, content=False)
             )
             return True
         except Exception:
@@ -129,9 +123,8 @@ class LocalBackend(Backend):
         Returns:
             Created notebook content
         """
-        model = await asyncio.to_thread(
-            self.contents_manager.new,
-            path=path
+        model = await ensure_async(
+            self.contents_manager.new(path=path)
         )
         return model['content']
     
@@ -202,13 +195,14 @@ class LocalBackend(Backend):
         content['cells'] = cells
         
         # Save updated notebook
-        await asyncio.to_thread(
-            self.contents_manager.save,
-            {
-                'type': 'notebook',
-                'content': content
-            },
-            path
+        await ensure_async(
+            self.contents_manager.save(
+                {
+                    'type': 'notebook',
+                    'content': content
+                },
+                path
+            )
         )
         
         return len(cells) - 1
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

PR #286, opened 2026-07-19
fix: don't persist the [No output generated] sentinel as a cell output

### LINKED ISSUE #285: execute_cell (JUPYTER_SERVER file mode) persists the [No output generated] sentinel as a phantom execute_result for output-less cells
### Summary

In `JUPYTER_SERVER` mode, when a notebook is not open in RTC (file mode), running a cell that produces no output (for example `x = 1`, an `import`, or a `def`) writes a fabricated `execute_result` carrying the string `[No output generated]` into the persisted `.ipynb`. A cell that produced nothing ends up with a phantom result output that is visible when the notebook is reopened.

### Mechanism

For an output-less cell, `execute_via_execution_stack` (`jupyter_mcp_server/utils.py:562`) returns the display sentinel `["[No output generated]"]` while leaving `raw_outputs` empty. `ExecuteCellTool._write_outputs_to_cell` (`execute_cell_tool.py:70-94`) then takes the `raw_outputs`-empty fallback branch. Because `[No output generated]` does not start with `[ERROR:`, `[TIMEOUT ERROR:`, or `[PROGRESS:`, it is written as an `execute_result` with `data={'text/plain': '[No output generated]'}` and stamped with the cell's execution count.

The sentinel is meant only for the tool response (so the calling agent sees a message), not for `cell.outputs` persistence.

### Reproduction

Verified on HEAD `a043d85` in a clean `python:3.12-slim` container (`pip install -e .`), driving `execute_via_execution_stack` with an ExecutionStack that reports no outputs, then persisting via `_write_outputs_to_cell`:

```
[step 1] execute_via_execution_stack(code="x = 1") -> ['[No output generated]'], raw_outputs == []
[step 2] _write_outputs_to_cell(['[No output generated]'], raw_outputs=[]) persists:
  [{"output_type": "execute_result",
    "data": {"text/plain": "[No output generated]"},
    "execution_count": 1, "metadata": {}}]
```

Expected: `cell.outputs == []` (the cell produced nothing), with the execution count still set.

### Notes

This is a residual of #278, which fixed the real-output fidelity path in the same function and intentionally kept the string fallback for the error/timeout/progress cases. Those cases are correct and should stay; only the empty-output sentinel should not be persisted.

ebarkhordar:
Fixes #285.

### Root cause

In `JUPYTER_SERVER` file mode, an output-less cell (`x = 1`, an `import`, a `def`) makes `execute_via_execution_stack` return the display sentinel `["[No output generated]"]` while leaving `raw_outputs` empty (`utils.py:562`). `_write_outputs_to_cell` then takes the `raw_outputs`-empty fallback, and since the sentinel does not start with `[ERROR:`, `[TIMEOUT ERROR:`, or `[PROGRESS:`, it was written back as an `execute_result` with `data={'text/plain': '[No output generated]'}`. A cell that produced nothing was persisted with a phantom result output.

This is a residual of #278, which fixed the real-output fidelity path in this function and intentionally kept the string fallback for the error, timeout, and progress cases. Those cases are correct; only the empty-output sentinel should not reach `cell.outputs`.

### Fix

In the fallback branch, skip the `[No output generated]` sentinel so an output-less cell persists no output. The sentinel is still returned to the caller (the tool response is unchanged), and the cell still gets its execution count. The error, timeout, and progress strings are untouched.

### Verification

Reproduced on HEAD `a043d85` in a clean `python:3.12-slim` container (`pip install -e .`), driving `execute_via_execution_stack` with an ExecutionStack reporting no outputs, then persisting via `_write_outputs_to_cell`: the `.ipynb` gained an `execute_result` carrying `[No output generated]`. With the fix, `cell.outputs == []`.

Tests added to `tests/test_execute_cell_output_fidelity.py`, one per scenario named above:

- `test_no_output_sentinel_is_not_persisted`: the sentinel yields `cell.outputs == []`, execution count still set. Fails on `main`, passes here.
- `test_error_string_still_persisted_without_raw_outputs`: an `[ERROR: ...]` string is still written as a stream output.
- the existing `test_falls_back_to_formatted_strings_without_raw_outputs` covers the timeout string still being written.

Full `tests/test_execute_cell_output_fidelity.py` module passes (12 tests). Not driven end-to-end against a live kernel: the empty-cell to sentinel step (`utils.py:562`) is exercised with a stub ExecutionStack, and the persisted-output defect is exercised directly through `_write_outputs_to_cell`, which is where the sentinel was written.

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -79,6 +79,12 @@ class ExecuteCellTool(BaseTool):
                         metadata={}
                     ))
                 elif isinstance(output, str):
+                    if output == '[No output generated]':
+                        # A display-only sentinel for the tool response (see
+                        # execute_via_execution_stack in utils.py): a cell that
+                        # produced nothing must persist no output, not a
+                        # fabricated execute_result.
+                        continue
                     if output.startswith('[ERROR:') or output.startswith('[TIMEOUT ERROR:') or output.startswith('[PROGRESS:'):
                         cell.outputs.append(nbformat.v4.new_output(
                             output_type='stream',
--- a/tests/test_execute_cell_output_fidelity.py
+++ b/tests/test_execute_cell_output_fidelity.py
@@ -168,3 +168,42 @@ async def test_falls_back_to_formatted_strings_without_raw_outputs(tmp_path):
     assert len(outputs) == 1
     assert outputs[0]["output_type"] == "stream"
     nbformat.validate(notebook)
+
+
+@pytest.mark.asyncio
+async def test_error_string_still_persisted_without_raw_outputs(tmp_path):
+    """An [ERROR: ...] fallback string keeps being written as a stream output."""
+    path = _write_notebook(tmp_path)
+    tool = ExecuteCellTool()
+
+    await tool._write_outputs_to_cell(path, 0, ["[ERROR: RuntimeError: boom]"])
+
+    notebook = _read_notebook(path)
+    outputs = notebook.cells[0].outputs
+    assert len(outputs) == 1
+    assert outputs[0]["output_type"] == "stream"
+    assert outputs[0]["text"] == "[ERROR: RuntimeError: boom]"
+    nbformat.validate(notebook)
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

PR #288, opened 2026-07-19
fix: keep a notebook's binding when ensure_kernel_alive regenerates a dead kernel

### LINKED ISSUE #287: ensure_kernel_alive rebinds a notebook to the default document after a dead-kernel restart (MCP_SERVER mode)
### Summary

In MCP_SERVER mode, if a notebook's kernel dies and is regenerated, `NotebookManager.ensure_kernel_alive` silently rebinds that notebook to the **default** document. Subsequent read/edit/execute operations then run against the wrong notebook without any error.

### Root cause

`use_notebook` records a notebook's binding via `add_notebook(name, kernel, server_url=runtime_url, token=runtime_token, path=notebook_path)`, which stores `notebook_info = {server_url, token, path}`.

When the kernel is later found dead, `ensure_kernel_alive` regenerates it with:

```python
new_kernel = kernel_factory()
self.add_notebook(name, new_kernel)   # notebook_manager.py:282
```

`add_notebook` unconditionally rebuilds the whole entry, and with no `server_url`/`token`/`path` passed it falls back to the config defaults (`config.document_url`, `config.document_token`, `config.document_id`). The original binding is overwritten, so the next `get_current_connection()` points at the default document.

The sibling reprovision path in `restart_notebook_tool.py` deliberately re-passes `path=notebook_path` to avoid exactly this; `ensure_kernel_alive` does not.

### Reproduction

Pure in-memory, no server needed (HEAD e56e7d1):

```python

class FakeKernel:
    def __init__(self, kid, alive=True): self.id, self._alive = kid, alive
    def is_alive(self): return self._alive

nm = NotebookManager()
nm.add_notebook("nb", FakeKernel("dead", alive=False),
                server_url="https://runtime.example/", token="tok", path="work/analysis.ipynb")

nm.ensure_kernel_alive("nb", lambda: FakeKernel("fresh"))

print(nm.get_notebook_path("nb"))   # expected 'work/analysis.ipynb'
# actual: the configured default document id (e.g. 'notebook.ipynb')
```

`get_notebook_path("nb")` returns the default document id instead of `work/analysis.ipynb` after the restart.

### Impact

An agent working a non-default notebook whose kernel is culled will, after the transparent kernel restart, have its cell reads/edits/executions silently redirected to the default notebook. There is no exception, so the corruption is invisible until the wrong file is modified.

### Environment

- HEAD e56e7d1, `notebook_manager.py:267-283`.

ebarkhordar:
## Root cause

In MCP_SERVER mode, `use_notebook` records a notebook's binding through `add_notebook(name, kernel, server_url=runtime_url, token=runtime_token, path=notebook_path)`, which stores `notebook_info = {server_url, token, path}`.

When that kernel later dies, `ensure_kernel_alive` regenerated it with a bare `self.add_notebook(name, new_kernel)` (notebook_manager.py:282). `add_notebook` rebuilds the whole entry, and with no `server_url`/`token`/`path` it falls back to the config defaults (`config.document_url`, `config.document_token`, `config.document_id`). The original binding is overwritten, so the next `get_current_connection()` builds a `NotebookConnection` against the default document. Reads, edits and executes then run on the wrong notebook with no error.

The sibling reprovision path in `restart_notebook_tool.py:31-38` already re-passes `path=notebook_path` for exactly this reason; `ensure_kernel_alive` did not.

## Invariant

Regenerating a dead kernel replaces only the kernel. It must never change which server/document the notebook entry points at.

`notebook_info` has a single writer in the package: `add_notebook` (notebook_manager.py:109). Every caller either establishes a fresh binding (`use_notebook_tool.py:261/269`), re-passes `path` on reprovision (`restart_notebook_tool.py:33`), or operates on the `default`/local notebook (`enroll.py`, `extension.py`, `execute_*_tool.py`, `utils.py`). The only site that rebound an already-bound, non-default notebook to the defaults was `ensure_kernel_alive`. This is the fix.

## Fix

When the notebook is already tracked, swap only the kernel field in place; otherwise fall back to `add_notebook`:

```python
if name in self._notebooks:
    self._notebooks[name]["kernel"] = new_kernel
else:
    self.add_notebook(name, new_kernel)
```

## Verification

Ran in a clean `python:3.12-slim` Docker container at HEAD e56e7d1 (`pip install -e .`, `__file__` confirmed as the repo tree):

- New `tests/test_notebook_manager_ensure_kernel_alive.py` (pure in-memory `NotebookManager`, no server): the binding-preservation test **fails on unmodified HEAD** (`server_url` comes back as the `http://localhost:8888` default) and **passes with the fix**. Three tests total (binding preserved on restart; live kernel returned untouched; absent-notebook fallback still uses `add_notebook`).
- `tests/test_restart_notebook_tool.py` still passes (the sibling path).
- `ruff check` and `mypy` on the changed lines are clean.

Not verified: I did not reproduce the downstream wrong-document read/edit end-to-end against a live remote RTC server; the trace from `get_current_connection` to a misdirected operation is read from code. The state corruption itself (`notebook_info` reset to defaults) is what the test asserts directly.

Fixes #287

### DIFF
--- a/jupyter_mcp_server/notebook_manager.py
+++ b/jupyter_mcp_server/notebook_manager.py
@@ -279,7 +279,15 @@ class NotebookManager:
         if kernel is None or not hasattr(kernel, 'is_alive') or not kernel.is_alive():
             # Create new kernel
             new_kernel = kernel_factory()
-            self.add_notebook(name, new_kernel)
+            if name in self._notebooks:
+                # Swap only the kernel: a restart replaces the kernel, never the
+                # notebook the entry points at. Re-adding via add_notebook without
+                # the original server_url/token/path would reset notebook_info to
+                # the config defaults (see add_notebook) and silently rebind this
+                # notebook to the default document.
+                self._notebooks[name]["kernel"] = new_kernel
+            else:
+                self.add_notebook(name, new_kernel)
             return new_kernel
         return kernel
     
--- a/tests/test_notebook_manager_ensure_kernel_alive.py
+++ b/tests/test_notebook_manager_ensure_kernel_alive.py
@@ -1,0 +1,90 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Unit tests for NotebookManager.ensure_kernel_alive kernel regeneration.
+
+These tests use an in-memory NotebookManager with minimal KernelClient
+stand-ins, so no running Jupyter server is required.
+"""
+
+from jupyter_mcp_server.notebook_manager import NotebookManager
+
+# A non-default binding, as use_notebook records in MCP_SERVER mode. Held in
+# constants (not call-site literals) so the "token" argument reads as data.
+RUNTIME_URL = "https://runtime.example/"
+RUNTIME_AUTH = "runtime-token"
+NB_PATH = "work/analysis.ipynb"
+
+
+class FakeKernel:
+    """Minimal KernelClient stand-in exposing only the liveness surface that
+    ensure_kernel_alive checks (`is_alive`), plus an id for identity asserts."""
+
+    def __init__(self, kernel_id, alive=True):
+        self.id = kernel_id
+        self._alive = alive
+
+    def is_alive(self):
+        return self._alive
+
+
+def _manager_with_bound_notebook():
+    """A notebook bound to an explicit (non-default) server_url/token/path,
+    as use_notebook does in MCP_SERVER mode, with a dead kernel."""
+    nm = NotebookManager()
+    nm.add_notebook(
+        "nb",
+        FakeKernel("dead-kernel", alive=False),
+        server_url=RUNTIME_URL,
+        token=RUNTIME_AUTH,
+        path=NB_PATH,
+    )
+    return nm
+
+
+def test_ensure_kernel_alive_preserves_binding_on_restart():
+    """Regenerating a dead kernel must keep the notebook's server_url/token/path
+    binding, and swap in the fresh kernel."""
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

echarles:
relaunching the ci to make it green...

echarles:
Current failure is

```
SKIPPED [3] tests/test_jupyter_extension.py:145: TEST_JUPYTER_SERVER is disabled
SKIPPED [2] tests/test_jupyter_extension.py:160: TEST_JUPYTER_SERVER is disabled
FAILED tests/test_execute_code_kernel_id.py::test_execute_code_runs_in_the_requested_kernel - AssertionError: assert 'raw-kernel' in '[TIMEOUT ERROR: IPython execution exceeded 30 seconds and was interrupted]'
 +  where '[TIMEOUT ERROR: IPython execution exceeded 30 seconds and was interrupted]' = <built-in method join of str object at 0x7fcf2e910a40>(<generator object test_execute_code_runs_in_the_requested_kernel.<locals>.<genexpr> at 0x7fcf28174790>)
 +    where <built-in method join of str object at 0x7fcf2e910a40> = ''.join
======= 1 failed, 195 passed, 7 skipped, 3 warnings in 317.72s (0:05:17) =======
make: *** [Makefile:33: test-mcp-server] Error 1
Error: Process completed with exit code 2.
````

echarles:
ci green now, merging

ebarkhordar:
Thanks for rerunning CI and getting this merged. Glad it could extend the binding-preservation pattern from the restart path in #260.

Issue #274, opened 2026-07-17
Proposal: Exposing Kernel Specs to Support Multi-Language Runtimes


yezhenqing:
### Problem

Currently, the MCP server's kernel handling mechanism primarily focuses on active/running sessions, leaving a gap in detecting and selecting installed **Kernel Specs** (e.g., Python, R, Bash, PySpark). 

In [`jupyter_mcp_server/tools/list_kernels_tool.py`](https://github.com/datalayer/jupyter-mcp-server/blob/main/jupyter_mcp_server/tools/list_kernels_tool.py), kernel specs appear to be used solely for augmenting existing session entries rather than being exposed as discoverable, language-agnostic environments. When no kernels are running, `ListKernelsTool` simply returns `"No kernels found on the Jupyter server"`, leaving the LLM completely unaware of available runtimes on the system.

---

### Proposed Solution

To extend this language-agnostic execution model to Jupyter's MCP capabilities, I have refactored the current repository with two key enhancements:

1. **Enhanced `list_kernels_tool.py` (`ListKernelsTool`)**:
   Restructured the tool output into two distinct sections:
   * `=== Available Kernel Specs ===` (Displaying `name`, `display_name`, `language`, and `environment`)
   * `=== Active Running Kernels ===` (Displaying `id`, `name`, `state`, `connections`, and `last_activity`)

2. **Upgraded `use_notebook` (`UseNotebookTool`)**:
   * Introduced an explicit `kernel_name` parameter (default: `"python3"`).
   * Added **Smart Metadata Resolution**: When attaching to an existing notebook without specifying a `kernel_id`, the tool automatically extracts `metadata.kernelspec.name` directly from the `.ipynb` file.
   * **Resolution Hierarchy**:
     1. Active `kernel_id` *(Highest priority)*
     2. Notebook Metadata Kernel Spec *(if `mode="connect"` and file exists)*
     3. Caller-provided `kernel_name`
     4. Default Fallback (`"python3"`)

Supporting changes have also been applied to `execute_cell_tool.py` and `execute_code_tool.py` to handle these execution flows seamlessly. In this way, `jupyter-mcp-server` will be able to seamlessly handle kernels for other programming languages.

---

### Additional context

* **Fork Implementation**: I have implemented and tested these changes in my fork: 👉 [yezhenqing/jupyter-mcp-server](https://github.com/yezhenqing/jupyter-mcp-server)
* **Demos & References**:
[...]

PR #292, opened 2026-07-19
feat: enable support for multiple kernel specs

yezhenqing:
This PR addresses issue #274.  For your convenience, I have attached a change summary generated by Gemini 3 on the patch file. I'm generally satisfied with the summary and have pasted its full content in the section below.

# 🛠️ Code Change Summary: Support for Notebook Kernel Specs Recognition

This Git Diff summary for the `jupyter-mcp-server` enhancement focuses on introducing support for non-default kernels (e.g., other languages or virtual environment kernels beyond `python3`). When connecting to an existing Notebook or dynamically starting a kernel, the system will now prioritize reading and matching the kernel configuration specified in the Notebook's metadata.

---

## 📋 Overview

1. **Added a new kernel metadata extraction utility** (`utils.py`).
2. **Dynamic kernel binding when connecting to or creating a Notebook** (`use_notebook_tool.py`, `server.py`).
3. **Intelligent kernel matching during on-demand kernel startup** (`execute_cell_tool.py`, `execute_code_tool.py`).
4. **Enhanced the kernel list tool to differentiate between available specs and active instances** (`list_kernels_tool.py`).

---

### 1. Utilities and Infrastructure Updates
* **`utils.py`**: 
  * Added the `extract_kernelspec_from_notebook` async function. It supports reading the metadata of a specified Notebook in both `JUPYTER_SERVER` (local) and `MCP_SERVER` (HTTP API) modes to extract the `kernelspec` information (`name` and `display_name`).
* **`server.py`**:
  * Added an optional `kernel_spec_name: str` parameter (defaulting to `"python3"`) to the `use_notebook` interface, allowing the frontend or caller to specify the desired kernel spec when creating or connecting to a notebook.

### 2. Core Notebook Tool Enhancements
* **`use_notebook_tool.py`**:
  * **When connecting to an existing Notebook:** Prioritizes calling the new utility function to extract the kernel name from the notebook's metadata. If the extracted kernel is installed in the current Jupyter environment, it forces the startup using that kernel. If not installed, it logs a warning and falls back to the default `python3`.
  * **When creating a new Notebook:** Writes the corresponding `kernelspec` configuration (`name` and `display_name`) into the initial file metadata, ensuring the newly generated `.ipynb` file is tagged with the correct kernel.
  * Modified the local kernel startup logic to explicitly pass the `target_kernel_name` to `kernel_manager.start_kernel()`.

### 3. Execution Tools Robustness Optimization (On-demand Kernel Startup)
* **`execute_cell_tool.py` & `execute_code_tool.py`**:
  * Optimized the fallback logic for "automatically triggering a new kernel startup when no kernel is available." Instead of blindly starting the default `python3`, it now extracts the file path from the current Notebook context and identifies its kernel spec. If a matching kernel spec exists in the system, it starts the new kernel using that specific spec.

### 4. Comprehensive Upgrade to Kernel List Display
* **`list_kernels_tool.py`**:
  * Refactored the return values of `_list_kernels_http` and `_list_kernels_local` from simply returning a list of "running kernels" to returning a tuple of `(available_specs, active_kernels)`.
  * **Output Formatting Optimization:** Split the final TSV output text into two distinct sections, making it much easier for both users and LLMs to parse:
    1. `=== Available Kernel Specs ===`: Displays all **available** kernel specs on the server that can be used to create/start new kernels (includes Name, Display_Name, Language, and Environment).
    2. `=== Active Running Kernels ===`: Displays the currently **running** kernel instances (includes ID, Spec_Name, State, and Connections).

---

## ⚠️ Remaining TODOs
In `execute_cell_tool.py` and `execute_code_tool.py`, there are `TODO` comments left regarding the dynamic adaptation logic for kernel specs under `ServerMode.MCP_SERVER` (HTTP mode). Equivalent support for passing kernel specs in this mode may need to be implemented in the future.


echarles:
> Equivalent support for passing kernel specs in this mode may need to be implemented in the future.

We try keeping features parity for all modes,would you be on to Kernelspecs support for Server Mode ?

PS: CI is failing with FAILED tests/test_tools.py::test_list_kernels[mcp_server] - AssertionError: assert 'ID\tName\tDisplay_Name\tLanguage\tState\tConnections\tLast_Activity\tEnvironment' in '=== Available Kernel Specs (Can be used to create/start new kernels) ===\nName (kernel_name)\tDisplay_Name\tLanguage\tEnvironment\npython3\tPython 3

yezhenqing:
Yes, I can fix those, including support for MCP Server mode. I've been thinking about a better way to minimize the impact on the core codebase. I'll get on it.

ebarkhordar (commented):
Thanks for adding this, kernel-spec selection is a real gap. One thing blocks it from working end to end: `extract_kernelspec_from_notebook` (utils.py:1088-1092) returns a dict, but the three call sites unpack it as a tuple, so the values are silently lost.

- use_notebook_tool.py:215 `extracted_name, extracted_display_name = extracted`
- execute_cell_tool.py:189 `extracted_name, _ = extracted`
- execute_code_tool.py (same pattern)

Unpacking a dict iterates its keys, so `extracted_name` becomes the literal string `"name"`, not the spec name. `"name" in kernel_spec_manager.find_kernel_specs()` is then always False, so every notebook falls back to `python3`, including the R and bash notebooks this feature is meant to serve.

Reproduced against the head commit (f8880a2) with the verbatim helper plus the use_notebook unpack, on an `ir` notebook whose `ir` spec is installed on the server:

```
return value : {'name': 'ir', 'display_name': 'R'}   type: dict
unpacked name: 'name'  display: 'display_name'
spec_exists  : False
=> kernel chosen: python3 (fallback)
```

Either fix works:
- return `(name, display_name)` from the helper to match the current unpack, or
- keep the dict and read it by key at the call sites, e.g. `extracted_name = extracted["name"]`.

Minor, on top: the annotation is `Optional[Dict[str, str]]` while the callers expect a 2-tuple, so aligning those would keep the type honest.

yezhenqing:
@echarles @ebarkhordar Thanks for your comments. 

I have revised create_kernel in utils.py to add support for the kernel-spec option in MCP_SERVER mode.

The main idea is that the kernel spec can be determined from config.document_id when it's available. To achieve this, I added a helper function, _get_kernel_spec_name(config, logger), in utils.py. However, I'm wondering if it would be cleaner to move this logic into config.py (e.g., as a property) to dynamically derive the kernel spec from document_id. 
Any feedback or suggestions would be greatly appreciated!

Additionally, I made minor updates to the tests (test_use_notebook_create_local.py and test_tools.py) to ensure the CI passes.


Summary of Changes
1. Dynamic Kernel Spec Selection from Notebooks
Notebook Metadata Extraction: Added logic (_get_kernel_spec_name) to extract the kernelspec from a notebook’s metadata (config.document_id).
Fallback & Validation: Validates whether the extracted kernel spec exists on the server. If valid, it starts the kernel using that spec; otherwise, it logs a warning and falls back to "python3".


2. Updated Kernel Lifecycle & Mode Handling
Kernel Startup: Updated kernel.start() to accept the dynamically resolved kernel spec name (kernel.start(name=kernel_spec_name)).


3. Tests and Minor Cleanups
Test Updates:
Updated RecordingKernelManager.start_kernel() in test mocks to accept the optional kernel_name parameter.
Updated test_list_kernels assertions to match updated output sections (=== Available Kernel Specs and === Active Running Kernels).


Note: I run the `make test` locally, the below one always failed.

FAILED tests/test_prompts.py::test_jupyter_cite - AssertionError: Cell 1 should contain Import Required Libraries
===================================================================== 1 failed, 242 passed, 3 warnings in 314.39s (0:05:14) ======================================================================
make: *** [Makefile:27: test] Error 1

I suspect this failure is unrelated to my code modifications. If possible, please help me to look into it. 
Thanks!

yezhenqing:
Also, it would be great to test a small demo in pure MCP_SERVER mode using Claude Code or similar clients. I don't have that setup on my machine right now, so I haven't verified it yet, though JUPYTER_SERVER mode is working well via JupyDeep.

echarles:
> Additionally, I made minor updates to the tests (test_use_notebook_create_local.py and test_tools.py) to ensure the CI passes.

Could you foresee a dedicated test_kernelspecs.py where the new tool is tested?

Is the doc updated with that new tool and behavior?

yezhenqing:
I agree that a dedicated test suite is needed. Ideally, it should include a new notebook with a different kernel spec to cover create_kernel and kernel description discovery. Down the road, we can also expand it to cover more tools like KernelSwitch. I don't have a clear layout for it just yet, but I'll keep them floating in my mind.

Haven't updated the docs yet either. Once I double check that pure MCP_SERVER mode works as expected, I'll get those updated.

AmirF194 (commented):
I ran `extract_kernelspec_from_notebook` against this branch (head `af91e913`) in a container and found a bug that defeats the PR's core feature.

The function returns a dict, e.g. `{"name": "ir", "display_name": "R"}`, but every call site unpacks it as a 2-tuple:

```python
extracted_name, _ = extracted
```

Unpacking a dict this way iterates its keys, not its values, so `extracted_name` is always the literal string `"name"`, never the kernel name pulled from the notebook's metadata. `spec_exists` then checks whether `"name"` is a registered kernel spec, which it never is, so detection silently falls back to `python3` every time. This hits all four call sites: `execute_cell_tool.py`, `execute_code_tool.py`, `use_notebook_tool.py`, and `_get_kernel_spec_name` in `utils.py`.

Repro (`extract_kernelspec_from_notebook` called directly, real repo code, Docker, python:3.11-slim):

```
extract_kernelspec_from_notebook(...) returned: {'name': 'ir', 'display_name': 'R'}
extracted_name used by the PR's spec_exists check: 'name'
AssertionError: BUG: expected the kernel spec name 'ir' from notebook metadata, got 'name' instead
```

Fix is one line at each site, e.g. `extracted_name = extracted["name"]` (and `extracted["display_name"]` where that value is used), or have the function return a 2-tuple instead of a dict.

yezhenqing:
To test the pure `MCP_SERVER` mode, I didn't use clients like Claude Code; instead, I used the equivalent `pydantic-deep` code snippet as below:

```
import asyncio

from pydantic_deep import create_deep_agent, DeepAgentDeps, StateBackend, build_mcp_server, MCPServerConfig


async def main():
    jupyter_server = build_mcp_server(
        MCPServerConfig(
            name="jupyter",
            transport="stdio",
            command="uvx",
            #args=["jupyter-mcp-server@latest"],
            args=["--from", "/home/yezhe/Developer/jupyterdev/github/jupyter-mcp-server", "jupyter-mcp-server"],
            env={
                "JUPYTER_URL": "http://127.0.0.1:8888",
                "JUPYTER_TOKEN": "MY_TOKEN",
                "ALLOW_IMG_OUTPUT": "true"
            }
        )
    )

    agent = create_deep_agent(
        model="openai:gpt-5.2",
        mcp_servers=[jupyter_server],
        instructions="You are a helpful coding assistant.",
    )

    deps = DeepAgentDeps(backend=StateBackend())

    result = await agent.run(
        #"Please list out all jupyter kernels",
        "Create a new R Notebook plotting sine with tidyverse (do not overwrite existing notebooks).",
        deps=deps,
    )
    print(result.output)


asyncio.run(main())
```

**Prompt:** *Please list out all jupyter kernels.*  
**Output:**

### Available kernel specs (you can start these)
- `python3` — **Python 3 (ipykernel)** (language: python)
- `ir` — **R** (language: R)

### Currently running kernels
- `818698af-3205-4193-b6df-f192808f4069` — spec: `python3`, state: `idle`, connections: `1`, last activity: `2026-07-23 16:43:17`

---

**Prompt:** *Create a new R Notebook plotting sine with tidyverse (do not overwrite existing notebooks).*

As shown in the attached screenshot, the new R notebook was successfully created and worked as expected.

mcp_test


In summary, this verifies that the `kernel_spec` functionality now works properly in both `MCP_SERVER` and `JUPYTER_SERVER` modes.

For the documentation, I have only updated the relevant sections:
1. `list_kernels` (major update on the `Returns` part)
2. `use_notebook` (minor update adding the description for the `kernel_spec_name` argument)
3. No need to update the docs for execute_cell_tool and execute_code_tool, as the kernel_spec is implicitly determined.


No code changes compared to the previous commit; only documentation was updated.
I've resolved all rebase conflicts. Please review and revise as needed. Thanks!

yezhenqing:
@ebarkhordar @AmirF194 Sorry about that—I was focused on the MCP_SERVER mode and missed that you both pointed out the dict unpacking issue.

I’d prefer to keep extract_kernelspec_from_notebook returning a dictionary since we might need to return additional metadata in the future—for example, a description field, which would help agents identify and distinguish between kernel specs.

Fixed! Thanks!

jupyter_mcp_server/tools/list_kernels_tool.py
```diff
@@ -12,22 +12,42 @@
...
+    def _list_kernels_http(
+            self, server_client: JupyterServerClient
+    ) -> tuple[List[Dict[str, str]], List[Dict[str, str]]]:
+        """List active kernels and available kernel specs using HTTP API (MCP_SERVER mode)."""
         try:
-            # Get all kernels from the Jupyter server
-            kernels = server_client.kernels.list_kernels()
+            # 1. Fetch Available Kernel Specs
+            available_specs = []
+            kernels_specs = server_client.kernelspecs.list_kernelspecs()
```
echarles, line 24: why asking the kernelspecs everytime we list the kernels? I am afraid this bring additional latency.

Kernelspec are quite static, would it make sense to list them at the start of the mcp server, keep them in cache so they can be reused, and only refresh the cache with a refresh_kernelspecs tool?

echarles:

jupyter_mcp_server/tools/list_kernels_tool.py
yezhenqing: Good point! We should cache the kernel specs for better performance. I'll revise the code and follow up with more details later. Thanks for the suggestion!

yezhenqing:
All good, I'll take care of it

echarles:
@yezhenqing kind ping on this :)

yezhenqing:
@echarles  Thanks for the heads-up! It’s great to see jupyter-mcp-server evolving so fast recently, though this PR has definitely fallen behind a bit now. :)

I've actually updated the code locally to cache the kernel spec in the ServerContext at startup, which avoids querying the kernels on every single tool call.

However, before moving on this PR here, I might need some guidance on the code_sandboxes side. Currently, kernel creation is encapsulated within that package with limited external exposure. I'm wondering if it makes sense to submit a small PR to code-sandboxes first, or if there's a better approach.

I’ll take a closer look this weekend and report back! Might need your help!

echarles:
> However, before moving on this PR here, I might need some guidance on the code_sandboxes side. Currently, kernel creation is encapsulated within that package with limited external exposure. I'm wondering if it makes sense to submit a small PR to code-sandboxes first, or if there's a better approach.

Yes, there should be some abstractions on the code-sandbox repo - kernelspec may need a better generic name (like `environment` for example)

jupyter_mcp_server/server.py
```diff
@@ -393,6 +395,7 @@ async def use_notebook(
             notebook_path=notebook_path,
             use_mode=mode,
             kernel_id=kernel_id,
+            kernel_spec_name=kernel_spec_name,
             ensure_kernel_alive_fn=__ensure_kernel_alive,
             contents_manager=server_context.contents_manager,
             kernel_manager=server_context.kernel_manager,
```
AmirF194, line 401: I ran UseNotebookTool.execute against this branch's head (5e7314d) in a container, using the exact kwargs server.py passes.

kernel_spec_manager is never forwarded into any of the four call sites that need it: use_notebook (line 401 below), both execute_cell call sites, and execute_code (which also drops contents_manager). All three tool classes gate the "is this spec actually installed" check on kernel_spec_manager being truthy, so with it always None, every connect to an existing notebook falls back to python3, and it does that while logging a false "not installed on this server" warning for a spec that is in fact installed. list_kernels (line 362) already threads kernel_spec_manager through; these four call sites don't.

Repro (real UseNotebookTool.execute, mocked kernel_manager/contents_manager matching what server.py passes, python:3.11-slim):

```
notebook metadata kernelspec.name = 'ir' (an R kernel)
a real kernel_spec_manager on this server would report 'ir' installed
kernel actually started with kernel_name='python3'
```

Fix is the same one-line addition at each site: `kernel_spec_manager=server_context.kernel_spec_manager,` (execute_code also needs `contents_manager=server_context.contents_manager,`, which it currently omits too).

echarles:
@yezhenqing This is opened since some time. Happy to leave it as is if you think you will have time to work in the future, or close, just tell me.

yezhenqing:
@Eric Thank you for the follow-up.  I would like to close this PR for now,
as it has diverged significantly from the current main branch and would
require extensive test coverage (pytest) that I don't want to rush into the
codebase incomplete.

Unfortunately, I'm currently tied up with other commitments and won't have
time to dive deep into this anytime soon. If anyone else wants to take this
over, it’s probably best to start fresh on a new branch. Hopefully, the
notes and discussions here serve as a good reference point. Sorry for this.

Best,
Zhenqing

Eric Charles ***@***.***> 于2026年9月7日周一 23:03写道：

echarles:
No issue at all and thank you for the work and discussions. I will close it for now, that can be reopened as needed.

PR #294, opened 2026-07-20
fix(delete_cell): reject negative and out-of-range cell indices

### LINKED ISSUE #293: delete_cell accepts negative indices: raw IndexError on out-of-range, silent wrong-cell deletion on in-range
### Summary

`delete_cell` does not validate negative cell indices. Depending on the value, a negative index either raises a raw `IndexError` that leaks out as an MCP tool error, or silently deletes the wrong cell, even though the tool is documented as 0-based and marked `destructiveHint=True`.

### Root cause

`delete_cell` is the only cell-index tool whose schema does not constrain indices to `ge=0`. In `server.py`, `read_cell`, `overwrite_cell_source`, `edit_cell_source`, `execute_cell`, and `move_cell` all declare `Field(..., ge=0)`, so pydantic rejects negatives at the MCP boundary. `delete_cell` declares `cell_indices: list[int], Field(..., min_items=1)` with no per-item `ge=0`, so negatives reach the tool.

Inside the tool, all three mode methods (`_delete_cell_ydoc`, `_delete_cell_file`, `_delete_cell_websocket`) bounds-check with `if max(cell_indices) >= len(...)`, which validates the upper bound only. Negative indices are never caught.

### Reproduction (no server needed)

```python
import asyncio, tempfile, os, nbformat
from jupyter_mcp_server.tools.delete_cell_tool import DeleteCellTool

nb = nbformat.v4.new_notebook()
for i in range(5):
    nb.cells.append(nbformat.v4.new_code_cell(source=f"cell {i}"))
fd, path = tempfile.mkstemp(suffix=".ipynb"); os.close(fd)
nbformat.write(nb, open(path, "w"))

tool = DeleteCellTool()
asyncio.run(tool._delete_cell_file(path, [-100]))  # IndexError: list index out of range
asyncio.run(tool._delete_cell_file(path, [-1]))    # no error: silently deletes the LAST cell
asyncio.run(tool._delete_cell_file(path, [0, -1])) # deletes cell 0 AND the last cell (max()==0 passes the check)
```

The schema gap is easy to confirm too: validating `[-100]` against `delete_cell`'s exact `Field(..., min_items=1)` annotation passes, whereas the `ge=0` on `read_cell`'s `cell_index` rejects `-100` at the boundary.

### Expected vs actual

- Expected: an out-of-range index (negative or too large) returns the clean `Cell index N is out of range. Notebook has M cells.` message, consistent with every other cell tool and with the positive out-of-range case.
- Actual: out-of-range negatives raise a raw `IndexError`; in-range negatives silently delete the wrong cell; `[0, -1]` deletes both cell 0 and the last cell because `max()==0` passes the check.

### Environment

`jupyter-mcp-server` main at `b215107` (v1.0.6), Python 3.13.0, mcp 1.28.1, pydantic 2.13.4, nbformat 5.10.4. The behavior is version and platform independent (pure server-side logic).

### Fix

I have a fix and tests ready, and will open a PR that adds per-index validation (rejecting any index `< 0` or `>= total_cells` with the existing message) across all three modes, keeping `delete_cell` consistent with the other cell-index tools.

sachhg:
Fixes #293.

### Problem

`delete_cell` validated indices with `max(cell_indices) >= len(...)`, which checks the upper bound only. Because it is also the one cell-index tool whose schema lacks `ge=0`, negative indices reach the tool and either raise a raw `IndexError` (out of range) or silently delete the wrong cell (in range, for example `-1` deletes the last cell). `[0, -1]` slips past entirely since `max()==0`.

### Fix

Added `DeleteCellTool._validate_indices()`, mirroring `MoveCellTool._validate_move()`, which rejects any index `< 0` or `>= total_cells` with the existing `Cell index N is out of range. Notebook has M cells.` message, and replaced the `max(...) >= len` check in all three modes (YDoc, file, WebSocket). This keeps `delete_cell` consistent with the rest of the cell-index tools (all of which already reject negatives) and introduces no new behavior.

### Tests

New `tests/test_delete_cell.py`, following the `test_move_cell.py` structure:

- Section A: unit tests for `_validate_indices` (rejects negatives, out-of-range, and mixed lists; accepts valid 0-based indices).
- Section B: file-based tests confirming an out-of-range negative now raises the clean `ValueError` instead of `IndexError`, that `-1` no longer silently deletes the last cell, that `[0, -1]` is rejected as a whole, and that valid deletes still work.

These tests fail on `main` and pass with this change. The full suite is green in both modes:

```
TEST_MCP_SERVER=true TEST_JUPYTER_SERVER=true pytest
261 passed
```

### Scope

One source file (`delete_cell_tool.py`) plus one new test file. No dependency, schema, or unrelated changes.

### DIFF
--- a/jupyter_mcp_server/tools/delete_cell_tool.py
+++ b/jupyter_mcp_server/tools/delete_cell_tool.py
@@ -24,6 +24,26 @@ class DeleteCellTool(BaseTool):
         else:
             return str(cell_source)
 
+    def _validate_indices(self, cell_indices: list[int], total_cells: int) -> None:
+        """Validate that every index is a valid 0-based cell position.
+
+        Args:
+            cell_indices: Indices of cells to delete (0-based)
+            total_cells: Total number of cells in the notebook
+
+        Raises:
+            ValueError: When any index is negative or >= total_cells. Checking
+                each index (rather than only ``max(cell_indices)``) rejects
+                out-of-range negatives, which would otherwise raise a raw
+                IndexError or silently delete the wrong cell.
+        """
+        for cell_index in cell_indices:
+            if cell_index < 0 or cell_index >= total_cells:
+                raise ValueError(
+                    f"Cell index {cell_index} is out of range. "
+                    f"Notebook has {total_cells} cells."
+                )
+
     async def _delete_cell_ydoc(
         self,
         serverapp: Any,
@@ -42,11 +62,8 @@ class DeleteCellTool(BaseTool):
         """
         nb = await get_notebook_model(serverapp, notebook_path)
         if nb:
-            if max(cell_indices) >= len(nb):
-                raise ValueError(
-                    f"Cell index {max(cell_indices)} is out of range. Notebook has {len(nb)} cells."
-                )
-            
+            self._validate_indices(cell_indices, len(nb))
+
             cells = nb.delete_many_cells(cell_indices)
             return cells
         else:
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @sachhg

PR #139, opened 2025-10-22
Feat: better cell-tools

ChengJiale150:
See in #127 .

### add `Cell` and `Notebook` Model

same as #133 

### union `list_cells` and `read_cells` to `read_notebook`

- add param `response_format` to control the output (`brief` for `list_cells`, `detailed` for `read_cells`)
- add `notebook_name` to read all connected notebook without activate it.
- add `start_index` and `limit` to control the token return by this tool when reading large notebook (>100 cells)
- change the `tests` to adapt above feature.

### `insert_cell`
- use same format API to format output in `insert_cell` to reduce inconsistence.

### `overwrite_cell_source`
- better description

### `execute_cell`
- better code origization and description
- set default timeout to 90s

### `insert_execute_code_cell`
- remove `tools/insert_execute_code_cell.py` and combine it with `insert_cell` and `execute_cell`

### `read_cell`
- union the output format as `list` to adapt standard MCP MutiModel ouput format.
- add `include_outputs` param to control whether to return output.

### `delete_cell`
- return output add the cell source of deleted cell to avoid delete cell incorrectly and help LLMs to recover.
- add description to tell LLMs to delete many cells in descending order of their index to avoid index shifting.

PR #299, opened 2026-07-23
fix(read-tools): check the live YDoc before falling back to file

### LINKED ISSUE #297: read_cell/read_notebook return stale content right after a write lands in the live YDoc
### Environment

- `jupyter-mcp-server`: 1.0.6, HEAD `076c669ed14db13c8c643a2701f423670ce1d9e9`
- `jupyter-nbmodel-client`: 0.14.9, `jupyter-collaboration`: installed from the repo's own declared dependency (no pin)
- Mode: `JUPYTER_SERVER` (extension, `--ServerApp.jpserver_extensions '{"jupyter_mcp_server": True}'`)
- Docker: `python:3.12-slim`, non-root user, repo bind-mounted (not COPY'd, so `dev/content` is present)

### Problem

In `JUPYTER_SERVER` mode, `read_cell` and `read_notebook` always read the notebook straight
off disk:

- `jupyter_mcp_server/tools/read_cell_tool.py:47-59`: `contents_manager.get(notebook_path, content=True, type='notebook')`
- `jupyter_mcp_server/tools/read_notebook_tool.py:48-55`: same call

Every cell-mutation tool does the opposite: it calls `get_notebook_model()` first
(`jupyter_mcp_server/utils.py:1191`) and only falls back to file I/O when no
collaborative YDoc room is open for that notebook (e.g.
`overwrite_cell_source_tool.py:57` `_overwrite_cell_ydoc`, `if nb: ... else: ... _overwrite_cell_file`).

When the notebook is open in the JupyterLab UI (or any other client holding a live
YDoc room), a write made through an MCP tool lands in the YDoc immediately, but the
on-disk file only catches up once `jupyter_server_ydoc`'s debounced autosave fires
(`document_save_delay`, ~1s default). Any `read_cell`/`read_notebook` call made in
that window returns the **pre-write** content, even though the write already
succeeded and is visible to every other YDoc-aware consumer.

This looks like a regression: `git log -G '_get_jupyter_ydoc' --oneline -- \
jupyter_mcp_server/tools/read_notebook_tool.py jupyter_mcp_server/tools/read_cell_tool.py`
on an unshallowed clone shows exactly two hits: `661dca2` (#138) added the
YDoc-first check to both read tools, and `1ea92f2`, the very next day, removed it
again as a clean ~110-line deletion. Nothing since restores it. `git log -G
'get_notebook_model'` on the same two files returns zero hits, ever.

### Repro (Docker-verified this session)

Seeds cell 1 with `sentinel_before_edit` (no YDoc room open yet, so this lands on
disk directly and is consistent by construction). Then opens a second, independent
`NbModelClient` connection to the same notebook (exactly what happens when a human
has it open in the UI), which forces a live YDoc room. `overwrite_cell_source` then
writes `sentinel_after_edit` through that room. Immediately after:

- The YDoc itself (read directly via the second connection) already has `sentinel_after_edit`.
- `read_cell(1)` still returns `sentinel_before_edit`.
- `read_notebook("default")` still contains `sentinel_before_edit` and not `sentinel_after_edit`.
- After waiting ~2.5s (past the autosave debounce) both `read_cell` and `read_notebook`
  return `sentinel_after_edit`.

```python
# test_stale_read_repro.py -- run from the repo root with
# pip install -e '.[test]' jupyter-collaboration && python3 test_stale_read_repro.py
import asyncio, logging, sys
sys.path.insert(0, "tests")
from conftest import JUPYTER_TOKEN, _start_server, _jupyter_extension_command, _find_free_port
from test_common import MCPClient
from jupyter_nbmodel_client import NbModelClient, get_notebook_websocket_url

BEFORE, AFTER = "sentinel_before_edit", "sentinel_after_edit"

def cell_source(result):
    return result["result"][1]

async def main():
    host, port = "localhost", _find_free_port()
    gen = _start_server(name="JupyterLab+MCP", host=host, port=port,
                         command=_jupyter_extension_command(host, port),
                         readiness_endpoint="/api", max_retries=15)
    url = next(gen)
    async with MCPClient(url, token=JUPYTER_TOKEN) as client:
        await client.insert_cell(1, "code", BEFORE)
        assert cell_source(await client.read_cell(1)) == BEFORE

        ws_url = get_notebook_websocket_url(server_url=url, token=JUPYTER_TOKEN,
                                             path="notebook.ipynb", provider="jupyter")
        ui_room = NbModelClient(ws_url)
        await ui_room.start()
        await asyncio.sleep(0.5)

        await client.overwrite_cell_source(1, AFTER)
        print("ydoc now:", ui_room.get_cell_source(1))          # -> sentinel_after_edit
        print("read_cell immediately:", cell_source(await client.read_cell(1)))  # -> sentinel_before_edit (STALE)

        await asyncio.sleep(2.5)
        print("read_cell after debounce:", cell_source(await client.read_cell(1)))  # -> sentinel_after_edit
        await ui_room.stop()

asyncio.run(main())
```

Console output from the actual run:

```
INFO:repro:seeded read_cell -> {'result': [..., 'sentinel_before_edit']}
INFO:repro:ydoc content immediately after write -> 'sentinel_after_edit'
INFO:repro:read_cell immediately after write -> 'sentinel_before_edit'
INFO:repro:read_notebook immediately after write contains BEFORE? True / AFTER? False
INFO:repro:read_cell after autosave window -> 'sentinel_after_edit'
INFO:repro:read_notebook after autosave window contains AFTER? True
```

### Expected

`read_cell`/`read_notebook` should check the live YDoc first, the same way the
write tools already do, so a read immediately following a write (from any client)
never observes the pre-write file.

### Possibly related

#146 ("UI edits stop syncing after Claude Code makes MCP edits") describes a
similar-shaped symptom from the reporter's side; I have not confirmed it is the
same root cause, but the timing window here (agent write -> immediate read/verify
in the same tool-call sequence) matches the pattern.

Happy to send a PR for this (the fix looks like calling `get_notebook_model()`
from both read tools, mirroring `overwrite_cell_source_tool.py`, with a fallback
to `contents_manager.get()` when no YDoc room is open) if that's useful.

AmirF194:
`read_cell` and `read_notebook` unconditionally read the notebook from disk in JUPYTER_SERVER mode (`contents_manager.get(notebook_path, content=True, type='notebook')`), unlike every cell-mutation tool: `overwrite_cell_source`, `insert_cell`, `delete_cell`, `move_cell`, and `clear_cell_output` all call `get_notebook_model()` first and only fall back to file I/O when no collaborative session is open. A write that already landed in the live YDoc is invisible to a read until the autosave debounce flushes it to disk, so an agent reading its own just-made edit sees stale content. `#138` added the YDoc-first check to these two read tools; `#139`, merged the next day, removed it as part of an unrelated refactor and nothing has restored it since.

Both tools now call `get_notebook_model()` first, using the same `serverapp.root_dir` path resolution the write tools use, and fall back to the existing file read only when no YDoc room exists.

Verified in a clean Docker container this session (`python:3.13-slim`, non-root, bind-mounted clone at HEAD `076c669`):
- Two new regression tests per tool: one asserts a read returns YDoc content over a stale file (and that `contents_manager.get()` is never called when a YDoc room exists), the other confirms the fallback to file when no room is open is unchanged. All 4 fail on `076c669` (`AttributeError: module has no attribute 'get_notebook_model'`, since the import doesn't exist pre-fix) and pass on this branch.
- Full suite (`TEST_MCP_SERVER=true TEST_JUPYTER_SERVER=true pytest`, both transport modes): 274 passed, 1 skipped (`test_move_cell_preserves_outputs`, a pre-existing platform-timeout skip unrelated to this change), 0 failures.
- `header-license-fix` (skywalking-eyes) clean on the new file.

Not verified: the macOS/Windows CI legs (this session ran Linux only) and `test_lint`'s mypy/ruff output beyond a local run, since neither gates a merge in this repo's workflow (`lint.sh` has no `set -e`, so only its last command, `validate-pyproject`, determines the job's exit code).

Fixes #297

### DIFF
--- a/jupyter_mcp_server/tools/read_cell_tool.py
+++ b/jupyter_mcp_server/tools/read_cell_tool.py
@@ -4,12 +4,13 @@
 
 """Read cell tool implementation."""
 
+from pathlib import Path
 from typing import Any, Optional
 from jupyter_server_client import JupyterServerClient
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 from jupyter_mcp_server.notebook_manager import NotebookManager
 from jupyter_mcp_server.models import Notebook
-from jupyter_mcp_server.utils import get_current_notebook_context
+from jupyter_mcp_server.utils import get_current_notebook_context, get_notebook_model
 from mcp.types import ImageContent
 from jupyter_core.utils import ensure_async
 
@@ -45,7 +46,9 @@ class ReadCellTool(BaseTool):
             Cell information dictionary
         """
         if mode == ServerMode.JUPYTER_SERVER and contents_manager is not None:
-            # Local mode: read notebook directly from file system.
+            # Local mode: try the live YDoc first (collaborative session), same
+            # as every cell-mutation tool, so a read right after our own write
+            # sees it instead of the on-disk copy the autosave hasn't flushed yet.
             # Guard against no active notebook — without this, a None path causes
             # 'quote_from_bytes() expected bytes' deep in the contents manager.
             notebook_path, _ = get_current_notebook_context(notebook_manager)
@@ -53,10 +56,22 @@ class ReadCellTool(BaseTool):
             if not notebook_path:
                 return ["No active notebook. Use the use_notebook tool to activate a notebook first."]
 
-            model = await ensure_async(contents_manager.get(notebook_path, content=True, type='notebook'))
-            if 'content' not in model:
-                raise ValueError(f"Could not read notebook content from {notebook_path}")
-            notebook = Notebook(**model['content'])
+            from jupyter_mcp_server.jupyter_extension.context import get_server_context
+
+            context = get_server_context()
+            serverapp = context.serverapp
+            ydoc_path = notebook_path
+            if serverapp and not Path(ydoc_path).is_absolute():
+                ydoc_path = str(Path(serverapp.root_dir) / ydoc_path)
+
+            nb_model = await get_notebook_model(serverapp, ydoc_path) if serverapp else None
+            if nb_model:
+                notebook = Notebook(**nb_model.as_dict())
+            else:
+                model = await ensure_async(contents_manager.get(notebook_path, content=True, type='notebook'))
+                if 'content' not in model:
+                    raise ValueError(f"Could not read notebook content from {notebook_path}")
+                notebook = Notebook(**model['content'])
         elif mode == ServerMode.MCP_SERVER and notebook_manager is not None:
             # Remote mode: use WebSocket connection to Y.js document.
             # get_current_connection() falls back to the default pre-configured
--- a/jupyter_mcp_server/tools/read_notebook_tool.py
+++ b/jupyter_mcp_server/tools/read_notebook_tool.py
@@ -4,11 +4,13 @@
 
 """List cells tool implementation."""
 
+from pathlib import Path
 from typing import Any, Optional, Literal
 from jupyter_server_client import JupyterServerClient
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 from jupyter_mcp_server.notebook_manager import NotebookManager
 from jupyter_mcp_server.models import Notebook
+from jupyter_mcp_server.utils import get_notebook_model
 from jupyter_core.utils import ensure_async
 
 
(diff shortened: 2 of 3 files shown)


echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for the quick review and merge, echarles.

PR #302, opened 2026-07-23
fix(auth): use constant-time comparison for the MCP bearer token

### LINKED ISSUE #301: MCP bearer token comparison is not constant-time (CWE-208)
`RuntimeTokenVerifier.verify_token` (jupyter_mcp_server/server.py:77-80) is the sole check protecting the streamable-http MCP endpoint (execute_code/execute_cell and friends) when `--mcp-token`/`MCP_TOKEN` is set, and it is reused by `ManagementRouteSecurityMiddleware` for `/api/connect` and `/api/stop`:

```python
async def verify_token(self, token: str) -> AccessToken | None:
    if token != self._token:
        return None
    return AccessToken(token=token, client_id="mcp-client", scopes=[])
```

`!=` on strings short-circuits at the first mismatched byte, so the comparison time leaks how many leading bytes of a guessed token are correct (CWE-208). The standard fix is `hmac.compare_digest`, which runs in time independent of where the strings first differ.

To be clear about what this issue does and does not claim: I ran a local timing check (200k reps, byte-0 vs byte-63 mismatch on a 64-char token) and did not measure a distinguishable difference over this code path, interpreter and asyncio overhead dominate at this string length. So this is defense-in-depth hardening, not a demonstrated exploit against this server. It is still worth closing off, since a bearer-token check is exactly the kind of comparison `compare_digest` exists for.

Happy to open a PR swapping the comparison for `hmac.compare_digest` plus a regression test, if that is welcome.

AmirF194:
`RuntimeTokenVerifier.verify_token` (`server.py:77-80`) is the sole check protecting the streamable-http MCP endpoint (execute_code/execute_cell and friends) when `--mcp-token`/`MCP_TOKEN` is set, and it is reused by `ManagementRouteSecurityMiddleware` for `/api/connect` and `/api/stop`. It compared the token with a plain `!=`, which short-circuits at the first mismatched byte and leaks timing information about how many leading bytes of a guessed token are correct (CWE-208).

Fix: use `hmac.compare_digest`, the standard constant-time comparison for secrets.

Verified: added `tests/test_runtime_token_verifier.py`, which patches `hmac.compare_digest` to confirm it is actually invoked on every check (fails on main, since main never calls it; passes here), plus accept/reject/wrong-length cases for the unchanged accept/reject behavior. Ran in a clean `python:3.12-slim` Docker container against both `main` and this branch. What I did not verify: an actual measurable timing side-channel against this server, a local 200k-rep timing check (byte-0 vs byte-63 mismatch on a 64-char token) showed no distinguishable difference at this string length, so this is defense-in-depth hardening (CWE-208), not a demonstrated exploit.

Fixes #301

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -6,6 +6,7 @@
 Jupyter MCP Server Layer
 """
 
+import hmac
 from typing import Annotated, Literal
 from urllib.parse import urlsplit
 
@@ -76,7 +77,7 @@ class RuntimeTokenVerifier:
         self._token = token
 
     async def verify_token(self, token: str) -> AccessToken | None:
-        if token != self._token:
+        if not hmac.compare_digest(token, self._token):
             return None
         return AccessToken(token=token, client_id="mcp-client", scopes=[])
 
--- a/tests/test_runtime_token_verifier.py
+++ b/tests/test_runtime_token_verifier.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""
+Unit tests for RuntimeTokenVerifier (jupyter_mcp_server.server): no server needed,
+imports directly from the server module.
+
+Launch the tests:
+```
+$ pytest tests/test_runtime_token_verifier.py -v
+```
+"""
+
+import hmac
+from unittest.mock import patch
+
+import pytest
+
+from jupyter_mcp_server.server import RuntimeTokenVerifier
+
+
+class TestRuntimeTokenVerifier:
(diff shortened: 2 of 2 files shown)


echarles:
@abbbe this proposed change sounds legit to me, I would love your security expert eyes on this...

echarles:

AmirF194:

AmirF194:
Rebased onto the code-sandboxes refactor (#306/#319), the conflict was just an unused typing import (`Any`) that the refactor dropped. Tests still pass (4/4, `tests/test_runtime_token_verifier.py`), no other changes needed.

echarles (approved):
LGTM Thx @AmirF194

PR #305, opened 2026-07-23
fix: surface rich-text cell outputs instead of the object repr

### LINKED ISSUE #304: extract_output returns the object repr for Markdown/LaTeX/JSON cell outputs, dropping the content
### Problem

In the `display_data` / `execute_result` branch of `extract_output` (`jupyter_mcp_server/utils.py`), `text/plain` is checked before any richer key. For an `IPython.display.*` object the kernel emits a MIME bundle whose `text/plain` is only the bare object repr, while the real content lives in a richer text key (`text/markdown`, `text/latex`, `application/json`). The tool returns the repr and discards the content.

This value reaches the model through `safe_extract_outputs`, which `execute_cell`, `execute_code` and `read_cell` all call, so a cell that renders Markdown, LaTeX or JSON is handed to the agent as an unusable placeholder.

### Reproduction

Clean venv on current HEAD, pure function, no live server:

```python
from jupyter_mcp_server.utils import extract_output

md = {"output_type": "execute_result",
      "data": {"text/plain": "<IPython.core.display.Markdown object>",
               "text/markdown": "# hi"}}
print(extract_output(md))
# actual:   '<IPython.core.display.Markdown object>'
# expected: '# hi'
```

`text/latex` and `application/json` bundles behave the same way. A plain result (`text/plain` only) and a DataFrame (`text/plain` + `text/html`, where the ASCII table is the readable form) are already correct and should keep their current output.

ebarkhordar:
Fixes #304

### Root cause

In the `display_data` / `execute_result` branch of `extract_output` (`jupyter_mcp_server/utils.py`), `text/plain` is checked before any richer key. For an `IPython.display.*` object the kernel emits a MIME bundle whose `text/plain` is only the bare object repr, while the real content lives in `text/markdown`, `text/latex` or `application/json`. The tool returned e.g. `"<IPython.core.display.Markdown object>"` and dropped the actual `# hi`. The value flows to the model through `safe_extract_outputs`, used by `execute_cell`, `execute_code` and `read_cell`.

### Invariant

For a rich-display output, `extract_output` surfaces the readable text representation, preferring `text/markdown`, `text/latex` and `application/json` over an object-repr `text/plain`. A genuine `text/plain` (an ordinary result) is unchanged, and it still takes precedence over `text/html`.

### Fix

Before the existing `text/plain` check, prefer `text/markdown` / `text/latex`, and pretty-print `application/json`. Everything else is untouched: a plain result keeps its `text/plain`, a DataFrame (`text/plain` + `text/html`) keeps its ASCII table, and the `text/html`-only placeholder is left as-is (out of scope here).

### Verification

Clean Docker container (`python:3.11-slim`), current HEAD, editable install, provenance confirmed (`utils.__file__` is the checkout).

- New unit test `tests/test_extract_output_fidelity.py`: 3 cases fail on `main` (Markdown, LaTeX, JSON return the repr) and pass on this branch; 3 no-regression cases (plain-only, `text/plain` + `text/html`, `text/html`-only) pass on both.
- `tests/test_execute_cell_output_fidelity.py` still green.
- `ruff check` and `mypy` on `jupyter_mcp_server/utils.py`: no new findings versus `main`.

Not verified: no live-kernel end-to-end run. The change is a pure function over the MIME bundle and is exercised only by the unit tests above.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -5,9 +5,9 @@
 import re
 import asyncio
 import time
-import json
 from typing import Any, Optional, Union
 from mcp.types import ImageContent
+from jupyter_kernel_client import get_mimebundle_text
 from jupyter_mcp_server.config import ALLOW_IMG_OUTPUT
 from jupyter_mcp_server.hooks import HookEvent, HookRegistry
 from jupyter_nbmodel_client import NotebookModel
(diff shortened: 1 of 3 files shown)


echarles:
Is that a specific concern for jupyter mcp sever or could this be solved/managed in https://github.com/datalayer/jupyter-kernel-client?

ebarkhordar:
This is specific to the MCP server rather than to jupyter-kernel-client. The kernel client faithfully transports the full MIME bundle the kernel emits (text/plain alongside text/markdown, text/latex, application/json, image/png), and that is correct: a notebook UI legitimately wants text/html or the image, so collapsing the bundle in the client would change output for every consumer.

extract_output is the MCP-specific step that reduces that bundle to the single string a tool call returns to the model. An LLM cannot use text/html or a repr like `<IPython.core.display.Markdown object>`, so choosing which representation the model should see is a consumer-side decision, and extract_output is the only place in this pipeline that makes it (it feeds execute_code, execute_cell, and the read paths). That is why the fix sits here and stays scoped to the text an agent reads.

If you would rather push some of this into jupyter-kernel-client instead, I am glad to adjust.

echarles:
> If you would rather push some of this into jupyter-kernel-client instead, I am glad to adjust.

yes, would love this approach instead, so any other consumers of jupyter-kernel-client benefit from that util feature.

ebarkhordar:
Sounds good. I'll open a PR on jupyter-kernel-client that adds a helper to pick the richest text representation from a MIME bundle (text/markdown, text/latex, application/json, text/html over an object-repr text/plain), then rework this PR to call it so the logic lives in one place and other consumers of the client benefit too. I'll link the upstream PR here once it's up.

ebarkhordar:
Opened the upstream helper here:

https://github.com/datalayer/jupyter-kernel-client/pull/44

It adds `get_mimebundle_text`, which picks the richest readable text from a MIME bundle (text/markdown, text/latex, application/json over an object-repr text/plain). Once it is released I will rework this PR to call it so the selection logic lives in the client and every consumer benefits.

echarles:
Awesome! jupyter-kernel-client 0.12.0 is released with that get_mimebundle_text

ebarkhordar:
Done. The PR now delegates the text selection to `get_mimebundle_text` from jupyter-kernel-client 0.12.0, with the dependency floor bumped to `>=0.12.0`. `extract_output` unwraps CRDT YText values to their source and hands the key preference (markdown/latex/json over an object-repr text/plain, plain as the fallback) to the helper, so the selection logic lives in the client and every consumer shares it.

The fidelity tests still pass in a clean container against HEAD: markdown, latex and JSON return their real content, while the plain-only result and the text/html placeholder are unchanged.

echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the fast turnaround on the 0.12.0 release and the reviews. Good to have the text selection living in one place now.

PR #308, opened 2026-07-23
fix(execute): track a timed-out execute_cell's background task so is_kernel_busy sees it

### LINKED ISSUE #307: execute_cell timeout abandons its background thread; is_kernel_busy() is a silent no-op that never waits for it
### Bug description

`execute_cell` (both the streaming and non-streaming paths) runs `notebook.execute_cell(cell_index, kernel)` in a background asyncio Task wrapping `asyncio.to_thread(...)`. On timeout it calls `execution_task.cancel()` and `kernel.interrupt()`, then immediately returns the timeout result to the caller without ever awaiting the task.

`asyncio.Task.cancel()` on a task wrapping `asyncio.to_thread()` cannot stop the underlying OS thread once it has started running: cancellation only takes effect at the next `await` inside the coroutine, and the wrapped call has none. The thread keeps executing `notebook.execute_cell()`, and keeps mutating the shared notebook/kernel state, after the tool call has already returned a `TimeoutError` to the caller.

The intended guard against a second call starting while the first is still running is `wait_for_kernel_idle()` / `is_kernel_busy()` (`jupyter_mcp_server/utils.py`), called at the top of `execute_cell` and `execute_code` before starting a new execution. `is_kernel_busy()` checks `kernel._client.is_alive()`, but the installed `KernelClient` (`jupyter_kernel_client`) has no `_client` attribute, so the check always falls into `except Exception: return False`. It is a silent no-op, so a subsequent call never actually waits for a previous timed-out execution to finish, and can start a second `notebook.execute_cell()` against the same kernel/notebook connection while the first is still running in the background.

### Reproduction

Verified against HEAD `d33f0b6`, editable install in a clean `python:3.12-slim` container, with an in-memory fake kernel/notebook (same style as `tests/test_execute_cell_stream.py`):

1. Call `ExecuteCellTool().execute(..., stream=True, timeout_seconds=0)` with an `execute_cell` implementation that sleeps for 0.3s. The tool returns its `[TIMEOUT ...]` result almost instantly.
2. Immediately after, `is_kernel_busy(kernel)` returns `False`, even though the fake `execute_cell` is still asleep in its background thread for another ~0.3s.
3. `wait_for_kernel_idle(kernel)` returns immediately for the same reason, instead of waiting for the orphaned execution.
4. The same holds for the non-streaming path, `execute_cell_with_forced_sync()` in `jupyter_mcp_server/utils.py`.

This is a plausible mechanism for issue #298's "stray output lands in the notebook after the client already timed out," though I have not confirmed that report shares this exact root cause.

### Expected behavior

`is_kernel_busy()` should reflect that a previous execution's background thread is still running until it actually finishes, so `wait_for_kernel_idle()` makes a subsequent call wait instead of racing a still-running one on the same kernel/notebook connection.

AmirF194:
### Root cause

`execute_cell` runs `notebook.execute_cell(cell_index, kernel)` in a background task wrapping `asyncio.to_thread(...)` (both the streaming branch in `execute_cell_tool.py` and `execute_cell_with_forced_sync` in `utils.py`). On timeout it calls `.cancel()` and `kernel.interrupt()`, then returns the timeout result without ever awaiting the task.

`asyncio.Task.cancel()` on a task wrapping `asyncio.to_thread()` cannot stop the underlying OS thread once it has started: the cancellation only takes effect at the next `await` inside the coroutine, and the wrapped call has none. The thread keeps running `notebook.execute_cell()`, and keeps mutating the shared notebook/kernel state, after the caller already received a `TimeoutError`.

The intended guard against a second call racing that orphaned execution is `wait_for_kernel_idle()` / `is_kernel_busy()`, called before every `execute_cell`/`execute_code`. `is_kernel_busy()` checked `kernel._client.is_alive()`, but `KernelClient` (`jupyter_kernel_client`) has no `_client` attribute, so it always fell into `except Exception: return False`. It was a silent no-op: a subsequent call never waited for a previous timed-out execution to finish.

### Fix

`track_pending_execution(kernel, task)` (new, in `utils.py`) records the background task on the kernel when it is created, and clears it via a done callback once the task actually finishes, whichever way. `is_kernel_busy()` now reports that task's state instead of the nonexistent `_client` attribute. `wait_for_kernel_idle()` was otherwise unchanged, so it now correctly blocks a subsequent call until a previous timed-out execution finishes, instead of racing it on the same kernel/notebook connection.

### Verified

In a clean `python:3.12-slim` container, editable install, HEAD `d33f0b6`:

- New `tests/test_execute_orphaned_timeout.py` (3 tests, fake kernel/notebook, same style as `tests/test_execute_cell_stream.py`): fail on unfixed `utils.py`/`execute_cell_tool.py` (`is_kernel_busy` stays `False` right after a timeout; `wait_for_kernel_idle` returns in ~0s instead of waiting for the still-running background thread), pass on this branch. `__file__`-checked on both sides to confirm the installed package pointed at the file under test.
- `TEST_MCP_SERVER=true TEST_JUPYTER_SERVER=false pytest` and `TEST_MCP_SERVER=false TEST_JUPYTER_SERVER=true pytest` (the two `make test-*` targets): same pass/fail counts as main plus these 3, no new failures. The pre-existing `test_span_log_summary` failure and the ~55 `JupyterLab failed to start` errors (tests needing a live server) reproduce identically on unmodified main, unrelated to this change.
- `ruff check` / `mypy` on the two changed source files against a pristine `origin/main` worktree already return hundreds of pre-existing findings unrelated to this diff (matches this repo's known `lint.sh` local/CI mismatch); mypy's error count on these two files is identical before and after this change (26), and none of ruff's new findings touch the lines this PR adds.
- Bare-install smoke test from `test.yml` (`pip install .`, import every submodule, `jupyter-mcp-server --help`): clean.

Not verified: a live Jupyter server with a real kernel, end to end. The fake-kernel tests exercise the exact control-flow path (`.cancel()` without awaiting, then a subsequent `is_kernel_busy`/`wait_for_kernel_idle` check) that both call sites share with the real kernel client.

Fixes #307

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -23,6 +23,7 @@ from jupyter_mcp_server.utils import (
     wait_for_kernel_idle,
     safe_extract_outputs,
     execute_cell_with_forced_sync,
+    track_pending_execution,
     extract_output
 )
 
@@ -303,6 +304,7 @@ class ExecuteCellTool(BaseTool):
                     execution_task = asyncio.create_task(
                         asyncio.to_thread(notebook.execute_cell, cell_index, kernel)
                     )
+                    track_pending_execution(kernel, execution_task)
 
                     start_time = time.time()
                     last_output_count = 0
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -502,17 +502,38 @@ def ensure_kernel_alive(notebook_manager, current_notebook, create_kernel_fn):
     return notebook_manager.ensure_kernel_alive(current_notebook, create_kernel_fn)
 
 
+def track_pending_execution(kernel, task):
+    """Remember a background execute_cell task on the kernel so is_kernel_busy
+    can see it, and forget it once the task actually finishes.
+
+    asyncio.Task.cancel() on a task wrapping asyncio.to_thread() cannot stop the
+    underlying OS thread: the thread keeps running notebook.execute_cell() (and
+    mutating the shared notebook/kernel state) until it returns on its own,
+    regardless of the cancellation request. Recording the task here lets a
+    later call see that a previous execution is still in flight instead of
+    assuming the kernel is free the moment a timeout is raised.
+    """
+    kernel._mcp_pending_execution = task
+
+    def _clear(finished_task, kernel=kernel):
+        if getattr(kernel, "_mcp_pending_execution", None) is finished_task:
+            kernel._mcp_pending_execution = None
+
+    task.add_done_callback(_clear)
+
+
 async def execute_cell_with_forced_sync(notebook, cell_index, kernel, timeout_seconds = 300):
     """Execute cell with forced real-time synchronization."""
     from jupyter_mcp_server.log import logger
-    
+
     start_time = time.time()
-    
+
     # Start execution
     execution_future = asyncio.create_task(
         asyncio.to_thread(notebook.execute_cell, cell_index, kernel)
     )
-    
+    track_pending_execution(kernel, execution_future)
+
     last_output_count = 0
     
     while not execution_future.done():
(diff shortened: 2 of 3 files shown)


echarles:
Fails on macos with 

FAILED tests/test_execute_orphaned_timeout.py::test_wait_for_kernel_idle_blocks_until_orphaned_stream_task_finishes - assert 6.9141387939453125e-06 >= 0.3

I have relauched, if reproducible, there is something to be instigated with async python on mac

AmirF194:
You're right, and I think I found why: the monitoring loop in execute_cell_tool.py polls in whole-second increments (`await asyncio.sleep(1)`), so a timed-out call can burn a real second or more before returning control to the caller. The original background sleep was only 0.3s, so on a slower or more loaded runner that window can already be over by the time the test takes its first measurement, which matches the near-zero elapsed you saw.

Pushed a fix that gives the background work real margin over that loop overhead (2.0s instead of 0.3s) and adjusted the dependent waits/assertions to match. Ran the updated test in a loop, 12/12 clean on Linux, but I don't have a macOS runner to reproduce your original failure directly, so this is inference from reading the code and the CI log, not a confirmed macOS repro. Let me know if it still flakes there.

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for confirming and merging. Appreciate you catching the macOS flake report so quickly, that made the fix straightforward.

Issue #298, opened 2026-07-23
When the execution time of the cell exceeds 3 minutes, the client will time out and output to the output of the notebook


SweetMojito:

echarles:
See https://jupyter-mcp-server.datalayer.tech/reference/configuration/#command-line-options

```
  --execution-timeout INTEGER RANGE   Default timeout in seconds for code execution when a tool call passes no timeout (default: 120)
  --max-execution-timeout INTEGER RANGE
                                      Maximum timeout in seconds a tool call may request (default: 3600)
```

PR #306, opened 2026-07-23
feat: code sandboxes

echarles:
Fixes https://github.com/datalayer/jupyter-mcp-server/issues/303
Fixes https://github.com/datalayer/jupyter-mcp-server/issues/291

PR #309, opened 2026-07-23
fix(execute): MCP progress keepalive for long cell runs (#298)

stantheman0128:
## Summary

Long-running `execute_cell` / `execute_code` calls were silent on the MCP wire until they finished. Raising `--execution-timeout` / `JUPYTER_MCP_EXECUTION_TIMEOUT` alone does not send protocol traffic, so many MCP clients idle-timeout around a few minutes while the kernel keeps writing notebook outputs afterward (issue #298).

This change:

- Emits MCP `report_progress` + `info` keepalives on a configurable `progress_interval` during `execute_cell` (stream and non-stream), `insert_execute_code_cell`, and `execute_code`
- After a timeout + kernel interrupt, briefly settles the background `to_thread` work (without cancelling the asyncio Task) so the tool response can include notebook outputs that land in that window, instead of leaving them as stray notebook-only writes
- Documents the keepalive behavior next to the existing timeout config

Related: #308 already tracks orphaned post-timeout tasks for `is_kernel_busy`; this builds on that for the client-visible long-cell path.

## AI assistance

This PR was prepared with AI assistance (Cursor/Grok). I reviewed the change, ran the tests below, and am responsible for the submission.

## Evidence

`	ext
python -m pytest tests/test_execute_progress_keepalive.py tests/test_execute_orphaned_timeout.py tests/test_execute_cell_stream.py tests/test_execute_code_timeout.py -v --tb=line

...
13 passed in ~54s
`

Long-cell path covered by `test_stream_path_emits_progress_callback_during_long_cell`, `test_forced_sync_path_emits_progress_callback_during_long_cell`, and `test_execute_code_emits_progress_callback_during_long_run` (progress callback invoked while a multi-second fake cell is still running). Timeout + notebook output sibling covered by `test_timeout_settle_includes_late_notebook_output_in_stream_result` plus the existing orphaned-timeout suite.

## What was not tested

- Live Cursor/Claude Desktop idle-timeout against a real 3+ minute kernel cell (depends on client progressToken support; keepalive no-ops if the client omits a progress token, though `info` logs still go out)
- Full JupyterLab UI click-through of notebook outputs after a live timeout

## How to use

`ash
# raise the server-side ceiling (already supported)
jupyter-mcp start --execution-timeout 600 --max-execution-timeout 3600
# or
export JUPYTER_MCP_EXECUTION_TIMEOUT=600
`

Call `execute_cell` with a matching `timeout` (or `0` for the config default). Progress keepalives fire every `progress_interval` seconds (default 5) even when `stream=false`.

### DIFF
--- a/docs/docs/reference/tools/index.mdx
+++ b/docs/docs/reference/tools/index.mdx
@@ -196,7 +196,7 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
 - Input:
   - `cell_index`(int): Index of the cell to execute (0-based)
   - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s via `JUPYTER_MCP_MAX_EXECUTION_TIMEOUT`)
-  - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: false)
+  - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: true)
   - `progress_interval`(int, optional): Seconds between MCP progress keepalive updates (and optional stream log lines). Default: 5. Progress is emitted even when `stream=false` so MCP clients do not idle-timeout during long cells.
 - Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. Supports streaming mode with progress updates for long-running cells.
 
@@ -208,7 +208,7 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
   - `cell_index`(int): Index of the cell to insert and execute (0-based)
   - `cell_source`(string): Code source for the cell
   - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s)
-  - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: false)
+  - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: true)
   - `progress_interval`(int, optional): Seconds between MCP progress keepalive updates (default: 5)
 - Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. Returns both insertion confirmation and execution results.
 
--- a/docs/docs/reference/tools/index.mdx
+++ b/docs/docs/reference/tools/index.mdx
@@ -198,7 +198,7 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
   - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s via `JUPYTER_MCP_MAX_EXECUTION_TIMEOUT`)
   - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: true)
   - `progress_interval`(int, optional): Seconds between MCP progress keepalive updates (and optional stream log lines). Default: 5. Progress is emitted even when `stream=false` so MCP clients do not idle-timeout during long cells.
-- Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. Supports streaming mode with progress updates for long-running cells.
+- Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. With `stream=true` the cell outputs come first, unchanged, followed by timeline entries (`[PROGRESS: ...]`, `[COMPLETED in ...]`, `[TIMEOUT at ...]`).
 
 #### 17. `insert_execute_code_cell`
 
@@ -210,7 +210,7 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
   - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s)
   - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: true)
   - `progress_interval`(int, optional): Seconds between MCP progress keepalive updates (default: 5)
-- Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. Returns both insertion confirmation and execution results.
+- Returns: `list[Union[str, ImageContent]]` - List of outputs from the executed cell including text, HTML, and images. Returns both insertion confirmation and execution results. With `stream=true` the cell outputs come first, unchanged, followed by timeline entries.
 
 #### 18. `read_cell`
 
--- a/docs/docs/reference/tools/index.mdx
+++ b/docs/docs/reference/tools/index.mdx
@@ -195,9 +195,9 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
 - Execute a cell from the currently activated notebook with timeout and return it's outputs
 - Input:
   - `cell_index`(int): Index of the cell to execute (0-based)
-  - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s)
+  - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s via `JUPYTER_MCP_MAX_EXECUTION_TIMEOUT`)
   - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: false)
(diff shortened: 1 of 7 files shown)


docs/docs/reference/tools/index.mdx
```diff
@@ -143,6 +143,8 @@ When `use_notebook` starts a new kernel for a notebook, the kernel is started wi
   - `cell_index`(int): Index of the cell to insert and execute (0-based)
   - `cell_source`(string): Code source for the cell
   - `timeout`(int, optional): Maximum seconds to wait for execution (0 = use config default, default: 0). Configurable via `JUPYTER_MCP_EXECUTION_TIMEOUT` env var (default: 120s, max: 3600s)
+  - `stream`(bool, optional): Enable streaming progress updates for long-running cells (default: false)
```
echarles: could stream be set true by default?

stantheman0128:
CI follow-up for the red `test (windows-latest, 3.10)` job.

`test_timeout_settle_includes_late_notebook_output_in_stream_result` failed because `assert any("[TIMEOUT at" ...)` was False: on Windows with a coarse `time.time()` tick, the first monitor poll can see `elapsed == 0.0`, so `elapsed > timeout_seconds` (with `timeout_seconds=0`) does not fire. The loop then sleeps a full second, the 0.3s fake cell finishes, and the tool returns COMPLETED instead of TIMEOUT (settle never runs).

Fix in `1fd5320`:
- measure elapsed with `time.perf_counter()`
- treat timeout with `elapsed >= timeout_seconds` (so `0` means immediate timeout)
- sleep only the remaining timeout budget instead of always 1s
- same comparison on the forced-sync and `execute_code` paths
- added a frozen-clock regression test

Local Evidence: `pytest tests/test_execute_progress_keepalive.py tests/test_execute_orphaned_timeout.py -v` -> 8 passed (Win / Python 3.13).

docs/docs/reference/tools/index.mdx
stantheman0128: The keepalive and the stream flag are two separate things here, so a quick clarification on what a true default would change.

The MCP progress keepalive that fixes #298 is already on by default in both modes. Even with `stream=false`, execute_cell emits a progress notification every `progress_interval` seconds (`execute_cell_with_forced_sync` in utils.py), so clients no longer hit idle timeouts on long cells.

What `stream=true` changes is the return payload. Instead of the plain cell outputs, the tool returns a timestamped execution log: `[12.3s] <output>`, `[PROGRESS: ...]`, `[COMPLETED in 41.2s]`. I kept the default at false so existing clients and tests keep getting the output shape they expect from `safe_extract_outputs`.

If you prefer the annotated log as the default, I can flip it to true in this PR and update the docs and tests together.

echarles:

stantheman0128:
Rebased onto v1.1.0. No problem at all about the conflict.

The `SandboxKernel` adapter kept the `KernelClient` surface this PR relies on (`interrupt()`, `execute()`), so nothing in the keepalive or timeout-settle logic needed rewriting. Both test files are byte-identical to the pre-rebase versions and still pass:

```
tests/test_execute_progress_keepalive.py  5 passed
tests/test_execute_orphaned_timeout.py    3 passed
======================== 8 passed in 23.43s ========================
```

The remaining failures in the wider suite (`test_cli_typer_config`, `test_span_log_summary`, and the `tests/test_tools.py` errors) reproduce identically on unmodified `main` in my environment, so they are local rather than introduced by the rebase.

On your earlier question about `stream`: the keepalive that fixes #298 is already on by default in both modes, so making `stream=true` the default would only change the return payload shape for existing clients. Happy to flip it if you prefer the annotated log as the default, and I would update the docs and tests in the same push.

AmirF194 (commented):
Nice fix for the client idle-timeout problem, and using `settle_timed_out_execution` + `track_pending_execution` instead of cancelling the task avoids reintroducing the CancelledError-discards-the-log bug from #276.

One redundancy in the non-streaming path: `execute_cell_with_forced_sync` (utils.py:679-690) already calls `kernel.interrupt()` and awaits `settle_timed_out_execution(execution_future)` before raising `TimeoutError`. The caller's `except asyncio.TimeoutError` block in `execute_cell_tool.py` (lines 488-497) then repeats both on the same task via `kernel._mcp_pending_execution`.

I ran this at PR head (3a8c971) with a fake kernel/cell that sleeps 5s and `timeout_seconds=0`: the tool call takes ~2.0s to return and `kernel.interrupt()` is called twice, instead of the single ~1s settle window `TIMEOUT_OUTPUT_SETTLE_SECONDS` implies.

```
elapsed: 2.0038020538631827
interrupt_calls: 2
result: ['[TIMEOUT ERROR: Execution exceeded 0 seconds]']
```

Not blocking, nothing breaks, but it doubles the settle wait and sends a duplicate interrupt on every non-stream timeout. Might be worth having `execute_cell_with_forced_sync` raise without settling and let the caller settle once, or dropping the caller's redundant interrupt/settle since the forced-sync function already did both.

(Disclosure: I used AI assistance to review this and ran the check above myself.)

docs/docs/reference/tools/index.mdx
echarles: Thx for the clarification. As a user, I find always nice to see outputs if there are, so stream being on be default makes sense to me. Yes, set it on by default, please

stantheman0128: Done in 1c6828a: `execute_cell` and `insert_execute_code_cell` now default `stream=true` (tool docs updated). Callers can still pass `stream=false` for the forced-sync path; MCP progress keepalives continue either way.

Re-ran the execute keepalive/orphaned/stream suite locally: 14 passed.

stantheman0128: Pushed in 1c6828a on this branch — thanks for confirming.

stantheman0128:
@AmirF194 Thanks for the careful non-streaming timeout review. Agreed on the double interrupt/settle.

On current PR HEAD the non-stream `TimeoutError` path no longer re-settles after `execute_cell_with_forced_sync` (it already interrupted + settled before raising). Streaming path is unchanged and still settles once in its own timeout branch. `stream=true` remains the default per echarles.

Stan Shih (stantheman0128)

stantheman0128:
Rebased onto current `main` and force-pushed.

- Merged upstream `code_sandboxes` / `sandbox_client.execute` path with our MCP progress keepalive + `settle_timed_out_execution` timeout handling in `execute_code_tool.py`.
- Conflict was only in the execute-code path; cell/stream paths unchanged on this branch.

CI should re-run shortly. Let me know if anything else is needed before review.

echarles:
>  @AmirF194 Thanks for the careful non-streaming timeout review. Agreed on the double interrupt/settle.
> On current PR HEAD the non-stream TimeoutError path no longer re-settles after execute_cell_with_forced_sync (it already interrupted + settled before raising). Streaming path is unchanged and still settles once in its own timeout branch. stream=true remains the default per echarles.

@AmirF194 Do you want to give another try, otherwise I am good to merge.

stantheman0128:
Rebased onto current `main` again and force-pushed (`37ffd91`).

- Resolved the new conflict in `execute_cell_tool.py` / `utils.py` by keeping upstream `execution_count_out` capture together with our MCP progress keepalive (`progress_callback` / `progress_interval`).
- Branch should be mergeable again; CI is re-running.

@AmirF194 @echarles — happy to take any further timeout-path feedback; otherwise this is ready from my side.

Stan Shih (stantheman0128)

stantheman0128:
CI follow-up: the failing `Smoke test (bare install)` job is the same `ModuleNotFoundError: No module named 'mcp.server.fastmcp'` import in `jupyter_cite_prompt.py` on current `main` (see main branch runs for the `bump: 1.1.3` push). Not introduced by this PR's execute/timeout diff.

Mergeable again after the rebase; happy to re-run once upstream fixes the mcp 2.0 import path.

Stan Shih (stantheman0128)

echarles:
Some test are still failing.

More PR on their way to be merge, with potential conflicts, sorry.

stantheman0128:
Thanks for the heads up, and no problem about the conflicts.

Rebased onto `main` (fcb5be4). The PR is mergeable again.

On the failing tests: every red `test` job and `test_sdist` on the previous run failed in the "Smoke test (bare install)" step with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`. That was pip resolving `mcp` 2.0.0 against my stale base (v1.1.3), and it is fixed upstream by your `deps: constraint mcp` (#328), which this branch now inherits.

Conflict resolution: `main` wrapped `execute_via_execution_stack` in `execute_cell_tool.py` with the kernel-not-found retry. I kept that retry and passed `progress_callback` / `progress_interval` through both the initial call and the retry call, in the RTC and the file-based branch.

Local verification on the rebased branch (Windows, Python 3.13):

```
pytest tests/test_execute_progress_keepalive.py tests/test_execute_orphaned_timeout.py
10 passed in 29.44s

pytest tests/test_execute_cell_stream.py tests/test_execute_cell_output_fidelity.py \
  tests/test_execute_cell_kernel_execution_count.py tests/test_execute_code_timeout.py \
  tests/test_execute_code_orphaned_timeout.py tests/test_extract_output_fidelity.py \
  tests/test_cell_tools_explicit_notebook_target.py
37 passed in 13.69s
```

One thing worth flagging separately: the Test workflow on `main` itself is currently red at fcb5be4, on `tests/test_auth.py::TestPasswordAuthE2E::test_password_auth_notebook_operations` (`expected product 335121091 not found in cell`). That looks unrelated to this PR, so it may show up here too.

Happy to rebase again whenever the queued PRs land. Just ping me.

echarles (approved):
LGTM Thx @stantheman0128

PR #310, opened 2026-07-23
fix(mcpb): sync bundle version numbers with the package version

AmirF194:
Root cause: `mcpb/manifest.json`'s `version` and `mcpb/pyproject.toml`'s `version` are independent, hand-maintained fields. Every release since #232 bumps `jupyter_mcp_server/__version__.py` only, so the three numbers have drifted apart: `__version__.py` is `1.0.7`, `mcpb/pyproject.toml` is `1.0.0`, and `mcpb/manifest.json` is `0.22.1`.

Fix: `scripts/sync_mcpb_version.py` reads `__version__.py` as the source of truth and rewrites the two mcpb files to match it; this PR runs it once to close the current gap. Its `--check` mode is wired into the `test_lint` CI job so a future release that forgets to run it fails the build instead of drifting again.

Verified in a clean Docker container (`python:3.10-slim`) against HEAD `38dbcb0`: `tests/test_mcpb_version_sync.py` fails on main (`0.22.1 != 1.0.7`, `1.0.0 != 1.0.7`) and passes after the fix, and `scripts/sync_mcpb_version.py --check` exits 0 on the fixed tree. Did not run the full `make test-mcp-server`/`make test-jupyter-server` suite, since it needs a live Jupyter/kernel stack this change does not touch; `pytest --collect-only` over the whole `tests/` directory confirms the new test file collects cleanly alongside the existing 287 tests, with no naming or fixture conflicts.

Fixes #236

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,6 +63,9 @@ jobs:
       - name: Run Linters
         run: |
           bash ./.github/workflows/lint.sh
+      - name: Check mcpb version sync
+        run: |
+          python scripts/sync_mcpb_version.py --check
 
   test_sdist:
     needs: build
--- a/mcpb/manifest.json
+++ b/mcpb/manifest.json
@@ -2,7 +2,7 @@
   "manifest_version": "0.4",
   "name": "jupyter-mcp-server",
   "display_name": "Jupyter MCP Server",
-  "version": "0.22.1",
+  "version": "1.0.7",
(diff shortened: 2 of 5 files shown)


echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging. Glad the sync script keeps the mcpb version from drifting further.

Issue #160, opened 2025-11-04
The `insert_execute_code_cell` tool stuck when `PYTHONDEVMODE=1` environment variable is set


Atry:
## Description
The `insert_execute_code_cell` tool stuck when `PYTHONDEVMODE=1` environment variable is set
## Reproduce

1. Create the following `.mcp.json`
    ```
    {
      "mcpServers": {
        "jupyter": {
          "command": "uvx",
          "args": [ "jupyter-mcp-server" ],
          "env": {
            "JUPYTER_URL": "http://localhost:8888",
            "JUPYTER_TOKEN": "my token",
            "ALLOW_IMG_OUTPUT": "true",
            "PYTHONDEVMODE": "1"
          }
        }
      }
    }
    ```
2. Run `claude` cli
3. Ask the agent to use `insert_execute_code_cell`

Then `claude` cli stuck.

## Expected behavior

`claude` cli should not get stuck.

## Context

Workaround: unset `PYTHONDEVMODE`

echarles:
Thx for opening the issue. Are there other tools that get stuck? (like execute_code, or insert_cell...)?

Atry:
`execute_code`, `insert_cell` and `use_notebook` work

`execute_cell` also gets stuck

echarles:
will try to reproduce. which python version? Any logs appearing?

Atry:
```
$ python --version
Python 3.13.8
```

Atry:
> Any logs appearing?

I did not see any abnormal logs from `claude` cli. I guess `jupyter-mcp-server` accedentally prints some logs to stdout when `PYTHONDEVMODE=1` but I cannot directly see `jupyter-mcp-server`'s logs from `claude` cli.

However, I do see the following logs from jupyter lab when `claude` cli get stuck.
```
[E 2025-11-04 02:06:09.192 ServerApp] Failed to write message
    Traceback (most recent call last):
      File "/nix/store/j8p98639hb4xm9g63hgl6f5pwmsq07al-agentic-sql-dev-env/lib/python3.13/site-packages/jupyter_server_ydoc/handlers.py", line 282, in send
        self.write_message(message, binary=True)
        ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #311, opened 2026-07-24
fix(notebook): bound the notebook disconnect so a stuck teardown cannot hang a tool call forever

AmirF194:
`insert_execute_code_cell`/`execute_cell` can hang forever when the MCP server process runs under `PYTHONDEVMODE=1`, as reported in #160.

## Root cause

`NotebookConnection.__aexit__` (`jupyter_mcp_server/notebook_manager.py`) awaits `NbModelClient.__aexit__()` with no timeout of its own. Under `PYTHONDEVMODE=1` I could reliably reproduce a case where the client's teardown (`jupyter_nbmodel_client/client.py`, `NbModelClient.stop()`) gets past its own bounded step (propagating queued changes, bounded by `REQUEST_TIMEOUT`, 10s default) and then stalls indefinitely cancelling its background sender task. Since every tool call that opens a notebook connection goes through `async with notebook_manager.get_current_connection() as notebook:`, Python's `async with` protocol has to wait for `__aexit__` to finish before the call can return (or propagate its exception); the stall there hangs the whole tool call with no response ever reaching the MCP client, matching the reported symptom exactly.

I did not track down why the client's own cancellation stalls specifically under dev mode (that path lives in the `jupyter-nbmodel-client` dependency, not in this repo), but the fix here does not depend on knowing that: no matter the cause, a stuck disconnect should never be able to block a tool call forever.

## Fix

Bound the disconnect with `asyncio.wait_for(..., timeout=DISCONNECT_TIMEOUT)` (20s). On timeout, log a warning and let the original result/exception continue propagating normally instead of blocking on cleanup.

## Verification

Reproduced end to end in a clean Docker container (Python 3.13, matching the reporter's version): a real JupyterLab server plus `jupyter-mcp-server` run over the actual `stdio` transport, driven by a scripted MCP client calling `insert_execute_code_cell`.

- On HEAD (`38dbcb0`), with `PYTHONDEVMODE=1`: the tool call never returns (60s+ with no response); without it, the same call completes in ~1.7s.
- With this fix, under `PYTHONDEVMODE=1`: the call now returns within the 20s bound, as an error result (`isError=True`) instead of hanging.
- Added `tests/test_notebook_connection_disconnect_timeout.py`, which stands in a fake client whose `__aexit__` never completes, so the fix is exercised deterministically without needing `PYTHONDEVMODE` or the underlying race. Confirmed it fails on unpatched `main` (the connection's own `__aexit__` hangs past the test's outer timeout) and passes on this branch.
- Ran the full existing suite (`pytest tests/`); all relevant execute/notebook-manager tests pass, plus the real end-to-end `test_execute_code[mcp_server]` integration test.

Not verified: the exact mechanism inside `jupyter-nbmodel-client`/`pycrdt` that stalls the cancellation under dev mode. This fix bounds the symptom in this repo; I have not opened an issue upstream since I could not pin the precise root cause there.

### DIFF
--- a/jupyter_mcp_server/notebook_manager.py
+++ b/jupyter_mcp_server/notebook_manager.py
@@ -9,6 +9,8 @@ This module provides centralized management for Jupyter notebooks and kernels,
 replacing the scattered global variable approach with a unified architecture.
 """
 
+import asyncio
+import logging
 from typing import Dict, Any, Optional, Callable, Union
 from types import TracebackType
 
@@ -16,6 +18,15 @@ from jupyter_nbmodel_client import NbModelClient, get_notebook_websocket_url
 from jupyter_kernel_client import KernelClient
 
 from .config import get_config
+
+logger = logging.getLogger(__name__)
+
+# Bounds NbModelClient.__aexit__ (the notebook websocket disconnect). The client's own
+# teardown already bounds propagating queued changes to REQUEST_TIMEOUT (10s default),
+# but a stuck cancellation of its background sender task past that point has no further
+# bound of its own, so a hang there was blocking every tool call that opened this
+# connection forever with no response ever reaching the MCP client (#160).
+DISCONNECT_TIMEOUT = 20
 
 
 class NotebookConnection:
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for the quick review and merge. Glad the timeout bound closes that hang for good.

PR #312, opened 2026-07-24
fix(tools): let cell-level tools target an explicit notebook by name

### LINKED ISSUE #256: two agents working with same instance of Jupyter MCP server - is it a supported use case ?
I have two instances of Claude Code connected to a single instance of Jupyter lab with MCP server extension.
The agents were working on different notebooks, and I noticed that at least one cell insertion operation went to wrong notebooks.
I wonder if this setup is supported at all?

PS. I am sorry I do not have time to write a proper bug report, will have more time in couple weeks...

AmirF194:
### Root cause

`NotebookManager` tracks one process-wide `_current_notebook` pointer (`notebook_manager.py:76`). Every cell-level tool (`insert_cell`, `overwrite_cell_source`, `edit_cell_source`, `read_cell`, `delete_cell`, `clear_cell_output`, `move_cell`) resolves its target notebook only through that pointer, via `get_current_notebook_context()` (JUPYTER_SERVER mode) or `get_current_connection()` (MCP_SERVER mode). Neither path accepts an explicit target, unlike `read_notebook`, `restart_notebook`, and `unuse_notebook`, which already take a `notebook_name` argument.

Two MCP clients sharing one server race this pointer: client 1 calls `use_notebook(A)`, client 2 calls `use_notebook(B)`, and any cell-level call from client 1 issued after that silently lands on B instead of A, because both clients read the same `_current_notebook` field. There is no way for a caller to opt out of the shared pointer and address a notebook directly.

As discussed on #256, the complete fix is a session-identifier redesign that would touch a large part of the tool surface. This PR is a smaller, backward-compatible step: it does not solve session isolation, it gives callers an explicit-targeting escape hatch so a client that already knows which notebook it means never has to depend on the shared pointer being correct.

### Fix

Each of the seven cell-level tools gains an optional `notebook_name` parameter, defaulting to `None`. Two small resolver helpers in `utils.py` mirror `get_current_notebook_context`:

- `resolve_notebook_path(notebook_manager, notebook_name)` for JUPYTER_SERVER mode: returns the explicit notebook's path/kernel id when `notebook_name` is given (raising if it is not a connected notebook), otherwise falls back to `get_current_notebook_context`.
- `resolve_notebook_connection(notebook_manager, notebook_name)` for MCP_SERVER mode: same fallback, over `get_notebook_connection` / `get_current_connection`.

Both are reused by the existing per-notebook accessors (`NotebookManager.get_notebook_path`, `get_kernel_id`, `get_notebook_connection`), so this does not introduce a second way of tracking notebook state, only a second way of choosing which existing entry to read. Omitting `notebook_name` is a no-op: behavior for every existing caller is unchanged.

`execute_cell` and `insert_execute_code_cell` are deliberately left out of this PR. Their JUPYTER_SERVER path resolves a kernel and a live YDoc/file_id alongside the notebook path, which is a materially larger change than the read/write tools above; targeting execution correctly needs the kernel side of the redesign echarles described, not just the notebook side.

### Verified

In a clean `python:3.13-slim` container (non-root user, dependencies from `pyproject.toml`'s core list installed, source bind-mounted so both `main` and this branch run the same image), HEAD `cffe9b1`:

- New `tests/test_cell_tools_explicit_notebook_target.py`: a fake `NotebookManager` with two independent notebooks, "A" (current) and "B". 9 tests call each of the 7 tools with `notebook_name="B"` while "A" stays current, plus one default-omitted control and one unknown-name error case. On unpatched `main`: 8 of 9 fail (`notebook_name` is silently absorbed by `**kwargs` and every op still lands on the current notebook "A" instead of "B"; only the default-omitted control passes). On this branch: all 9 pass.
- `ruff check` on the 9 changed source files: 625 findings on the branch vs. 605 on the same files at `main`; every new finding is `E501`/`UP045`, matching this repo's own pre-existing style in these files (the repo's `Optional[X]` annotations already trip `UP045` 44 times on `main`), and `ruff` is confirmed non-gating in practice on this repo (2628 pre-existing errors on `main` overall). `mypy` reports the identical 123 pre-existing module-resolution errors on `main` and on this branch (unrelated to this change, and unchanged by it).

Not verified: end to end against a real two-client race (two live MCP sessions hitting one server). The fake-manager tests exercise the exact resolution path (`resolve_notebook_path` / `resolve_notebook_connection`) that the real `NotebookManager` shares between both code paths, but I have not driven this through a live JupyterLab + two MCP stdio clients.

Fixes #256

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -588,6 +588,12 @@ async def insert_cell(
     ],
     cell_type: Annotated[Literal["code", "markdown"], Field(description="Type of cell to insert")],
     cell_source: Annotated[str, Field(description="Source content for the cell")],
+    notebook_name: Annotated[
+        str | None,
+        Field(
+            description="Target this specific connected notebook instead of the currently activated one. Use when multiple clients share this server, to avoid racing the shared 'current notebook' pointer. Omit to use the currently activated notebook."
+        ),
+    ] = None,
 ) -> Annotated[
     str, Field(description="Success message and the structure of its surrounding cells")
 ]:
@@ -602,6 +608,7 @@ async def insert_cell(
             cell_index=cell_index,
             cell_source=cell_source,
             cell_type=cell_type,
+            notebook_name=notebook_name,
         )
     )
 
@@ -616,6 +623,12 @@ async def insert_cell(
 async def overwrite_cell_source(
     cell_index: Annotated[int, Field(description="Index of the cell to overwrite (0-based)", ge=0)],
     cell_source: Annotated[str, Field(description="New complete cell source")],
+    notebook_name: Annotated[
+        str | None,
+        Field(
+            description="Target this specific connected notebook instead of the currently activated one. Use when multiple clients share this server, to avoid racing the shared 'current notebook' pointer. Omit to use the currently activated notebook."
+        ),
+    ] = None,
 ) -> Annotated[str, Field(description="Success message with diff showing changes made")]:
     """Replace the entire source of a cell in the currently activated notebook.
     Returns a diff showing the changes made.
@@ -631,6 +644,7 @@ async def overwrite_cell_source(
             notebook_manager=notebook_manager,
             cell_index=cell_index,
             cell_source=cell_source,
+            notebook_name=notebook_name,
         )
     )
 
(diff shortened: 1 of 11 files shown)


echarles:

AmirF194:

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging, and for the heads-up on the sandboxes refactor conflict, the rebase went smoothly once I knew what caused it.

PR #314, opened 2026-07-24
fix(execute): use the kernel's own execution_count instead of a file-scan guess

### LINKED ISSUE #313: execute_via_execution_stack discards the kernel's own execution_count, diverging from it after a kernel restart
`execute_via_execution_stack` (JUPYTER_SERVER file mode, no live YDoc) reads `result["outputs"]` from the `ExecutionStack.get()` reply but never reads the same reply's top-level `execution_count`, even though `jupyter-server-nbmodel`'s own `_execute_snippet` returns it alongside `outputs` (`{"status": ..., "execution_count": reply_content.get("execution_count"), "outputs": ...}`).

`_write_outputs_to_cell` then re-derives the cell's `execution_count` with a file-scan heuristic:

```python
max_count = 0
for c in notebook.cells:
    if c.cell_type == 'code' and c.execution_count:
        max_count = max(max_count, c.execution_count)
cell.execution_count = max_count + 1
```

This only agrees with the kernel's real counter when the two never diverge. A kernel restart resets the kernel's own counter without touching the notebook file on disk, so the next execution the kernel reports as `execution_count=1` gets persisted as `max(existing cells)+1` instead, and the `execute_result` output's own `execution_count` field (copied from `cell.execution_count`) is wrong too.

Reproduced with a unit test against `_write_outputs_to_cell` directly: a notebook where cell 0 already has `execution_count=5`, cell 1 executed with a kernel reply carrying `execution_count=1` (post-restart), currently persists `cell.execution_count == 6` instead of `1`.

Fault surface:
- `jupyter_mcp_server/utils.py`, `execute_via_execution_stack` (drops `result["execution_count"]`)
- `jupyter_mcp_server/tools/execute_cell_tool.py:107-117`, `_write_outputs_to_cell` (the `max_count + 1` heuristic)

Only the JUPYTER_SERVER file-mode path (no live YDoc) is affected; the RTC path writes `execution_count` into the Yjs cell directly on the `jupyter-server-nbmodel` side, so it's unaffected.

AmirF194:
## Root cause

`execute_via_execution_stack` (JUPYTER_SERVER file mode, no live YDoc) reads `result["outputs"]` from `ExecutionStack.get()` but never reads the same reply's top-level `execution_count`, even though `jupyter-server-nbmodel`'s `_execute_snippet` returns it right alongside `outputs`. `_write_outputs_to_cell` then re-derives the cell's `execution_count` by scanning the notebook's other cells (`max(existing) + 1`), which only matches the kernel's real counter when the two never diverge. A kernel restart resets the kernel's counter without touching the file on disk, so the next execution gets stamped with the file-scan guess instead of the kernel's own reported count.

Fixes #313.

## Invariant

`cell.execution_count` (and the `execute_result` output's own `execution_count`) reflects the kernel's own counter whenever the kernel reports one, not a value re-derived from the notebook file.

## Fix

- `execute_via_execution_stack` gains an optional `execution_count_out` out-param (same shape as the existing `raw_outputs` one) and appends the reply's `execution_count` to it once, for both the success and error branches (a kernel increments its counter either way).
- `_write_outputs_to_cell` gains an optional `kernel_execution_count` parameter; when given, it is used directly instead of the file-scan heuristic.
- The one file-mode call site now threads the kernel's reported count through. The RTC call site is untouched: `jupyter-server-nbmodel` already writes `execution_count` into the Yjs cell directly in that path.
- Without a kernel count (e.g. the timeout path, which only has formatted strings), the old heuristic is unchanged.

## Verification

Docker (`python:3.10-slim`, matching the CI matrix's lower bound), two isolated checkouts (branch and a `git worktree` of `main`) so the differential's provenance is real, confirmed via `jupyter_mcp_server.__file__` on each side:

- New regression test (`tests/test_execute_cell_kernel_execution_count.py`): a notebook where cell 0 already ran as `execution_count=5`, cell 1 executed with a simulated post-restart kernel reply reporting `execution_count=1`. FAILS on `main` (`TypeError: unexpected keyword argument 'kernel_execution_count'`), PASSES on the branch (`cell.execution_count == 1`).
- Full existing suite (`tests/`) run on both sides for comparison: identical baseline of 58 pre-existing errors (JupyterLab fails to start in this container) and one pre-existing unrelated failure (`test_otel_integration.py::test_span_log_summary`) on both `main` and the branch. No new failures from this change.
- Smoke test from `build.yml` (bare `pip install .`, import every submodule, `jupyter-mcp-server --help`, extension listing) and `scripts/sync_mcpb_version.py --check`: both pass.

Not verified: a live kernel-restart end-to-end run against a real JupyterLab server (this container's JupyterLab fails to start regardless of this change, a pre-existing environment gap, not something this PR touches).

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -38,6 +38,7 @@ class ExecuteCellTool(BaseTool):
         cell_index: int,
         outputs: list[str | ImageContent],
         raw_outputs: list[dict] | None = None,
+        kernel_execution_count: int | None = None,
     ):
         """Write execution outputs back to a notebook cell.
 
@@ -45,6 +46,12 @@ class ExecuteCellTool(BaseTool):
         persisted, which keeps each output's real output_type. Without them only
         the formatted strings are available (the timeout path, for example), and
         the output type can only be guessed from the string form.
+
+        When kernel_execution_count is given, it is used as the cell's
+        execution_count directly. Without it, the count is re-derived by
+        scanning the notebook's other cells, which diverges from the kernel's
+        own counter whenever the two disagree (most commonly after a kernel
+        restart mid-session).
         """
 
         with open(notebook_path, encoding="utf-8") as f:
@@ -111,12 +118,18 @@ class ExecuteCellTool(BaseTool):
         # (the raw outputs above come straight from the kernel).
         clean_notebook_outputs(notebook)
 
-        # Update execution count
-        max_count = 0
-        for c in notebook.cells:
-            if c.cell_type == "code" and c.execution_count:
-                max_count = max(max_count, c.execution_count)
-        cell.execution_count = max_count + 1
+        # Update execution count. Prefer the kernel's own reported count; it is
+        # the authoritative source and can diverge from a file-scan heuristic
+        # (e.g. after a kernel restart resets the kernel's counter below the
+        # notebook's existing max).
+        if kernel_execution_count is not None:
+            cell.execution_count = kernel_execution_count
+        else:
+            max_count = 0
+            for c in notebook.cells:
+                if c.cell_type == "code" and c.execution_count:
+                    max_count = max(max_count, c.execution_count)
+            cell.execution_count = max_count + 1
 
         # An execute_result carries the execution count of the cell that produced it.
         for output in cell.outputs:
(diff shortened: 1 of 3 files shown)


echarles:

AmirF194:

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging. Good catch flagging the sandboxes conflict ahead of time, saved me from chasing a confusing diff.

PR #316, opened 2026-07-24
fix(tools): report real kernel liveness for list_notebooks in JUPYTER_SERVER mode

### LINKED ISSUE #315: list_notebooks always reports kernel_status=dead in JUPYTER_SERVER (extension) mode
`list_notebooks` reports `kernel_status: dead` for every notebook when the server is
running as a Jupyter Server extension (`JUPYTER_SERVER` mode), even when the kernel is
alive.

## Root cause

`NotebookManager.list_all_notebooks()` (`jupyter_mcp_server/notebook_manager.py:398`)
decides liveness with:

```python
kernel_status = "alive" if hasattr(kernel, 'is_alive') and kernel.is_alive() else "dead"
```

`NotebookManager` stores two different shapes under `"kernel"` depending on the
deployment mode: in `MCP_SERVER` mode it's a real `jupyter_kernel_client.KernelClient`
(has `.is_alive()`); in `JUPYTER_SERVER` (extension) mode it's a plain
`{"id": kernel_id}` dict (see `use_notebook_tool.py`, `restart_notebook_tool.py`'s
`_reprovision_kernel`, and the on-demand kernel-start paths in
`execute_cell_tool.py`/`execute_code_tool.py`). A dict has no `is_alive` attribute, so
`hasattr(...)` is unconditionally `False` in that mode, and `kernel_status` is always
`"dead"` no matter what the real kernel state is.

`list_notebooks_tool.py` calls `list_all_notebooks()` unmodified in both modes, and the
`list_notebooks` MCP tool takes zero parameters, so every call in `JUPYTER_SERVER` mode
hits this.

## Reproduction

```python

nm = NotebookManager()
nm.add_notebook("nb", {"id": "kernel-123"}, server_url="local")  # JUPYTER_SERVER shape

print(nm.list_all_notebooks())
# {'nb': {'path': '', 'kernel_status': 'dead', 'is_current': True}}
# ...even though "kernel-123" is never checked against anything and could be alive.
```

## Fix

The rest of the codebase already has the right pattern for checking a JUPYTER_SERVER
kernel's liveness: `kernel_id in kernel_manager` against the server's real kernel
manager (`restart_notebook_tool.py:94`, `use_notebook_tool.py:234`).
`list_all_notebooks()` just needs that same kernel_manager threaded in so it can use it
for the dict-shaped case instead of the KernelClient-only `hasattr` check.

PR incoming with a regression test.

AmirF194:
## Root cause

`NotebookManager.list_all_notebooks()` (`jupyter_mcp_server/notebook_manager.py:398`)
decides kernel liveness with `hasattr(kernel, 'is_alive') and kernel.is_alive()`. That
works in `MCP_SERVER` mode, where `kernel` is a real `jupyter_kernel_client.KernelClient`,
but in `JUPYTER_SERVER` (extension) mode `kernel` is a plain `{"id": kernel_id}` dict
(see `use_notebook_tool.py`, `restart_notebook_tool.py`'s `_reprovision_kernel`, and the
on-demand kernel-start paths in `execute_cell_tool.py`/`execute_code_tool.py`). A dict
has no `is_alive`, so the check is unconditionally `False` and `kernel_status` reports
`"dead"` for every notebook regardless of the kernel's real state. `list_notebooks`
takes no parameters, so every call in extension mode hits this.

## Fix

Thread the server's real `kernel_manager` through (`server.py`'s `list_notebooks` ->
`ListNotebooksTool.execute` -> `list_all_notebooks`) and, for a dict-shaped kernel, check
liveness the way the rest of the codebase already does for `JUPYTER_SERVER` mode:
`kernel_id in kernel_manager` (same pattern as `restart_notebook_tool.py:94` and
`use_notebook_tool.py:234`). `MCP_SERVER` mode is untouched. When no `kernel_manager` is
passed, a dict-shaped kernel now reports `"unknown"` instead of the previous incorrect
`"dead"`.

The same `hasattr` pattern also exists at `server.py:286` (the standalone `/api/healthz`
route). I left it alone: that route only runs in `MCP_SERVER` mode, where the kernel is
always a real `KernelClient`, so it never sees a dict-shaped kernel and isn't part of
this bug.

## Verification

Added `tests/test_notebook_manager_list_all_notebooks_kernel_status.py`, constructing a
`NotebookManager` directly (no live Jupyter server needed):
- MCP_SERVER-mode kernel (KernelClient-shaped): unaffected, still uses `is_alive()`.
- JUPYTER_SERVER-mode kernel present in a fake kernel manager: now reports `alive`.
- JUPYTER_SERVER-mode kernel absent from the kernel manager: reports `dead`.
- JUPYTER_SERVER-mode kernel with no kernel_manager passed: reports `unknown`.

Ran on `main` (`a5c5684`): 3 of the 4 new tests fail, confirming the bug. Ran on this
branch: all 4 pass. Also ran the full non-integration test suite (everything except the
tests that spin up a real `jupyter lab` process, which needs a non-root user to start):
172 passed, no regressions, plus the existing `test_restart_notebook_tool.py` suite
green.

Fixes #315

### DIFF
--- a/jupyter_mcp_server/notebook_manager.py
+++ b/jupyter_mcp_server/notebook_manager.py
@@ -382,9 +382,15 @@ class NotebookManager:
             return self._notebooks[current]["notebook_info"].get("path")
         return None
 
-    def list_all_notebooks(self) -> dict[str, dict[str, Any]]:
+    def list_all_notebooks(self, kernel_manager: Any | None = None) -> dict[str, dict[str, Any]]:
         """
         Get information about all managed notebooks.
+
+        Args:
+            kernel_manager: Jupyter server's kernel manager (JUPYTER_SERVER mode only).
+                In that mode `kernel` is a plain dict ({"id": kernel_id}), not a
+                KernelClient, so liveness has to be checked against the real kernel
+                manager instead of via `is_alive()`.
 
         Returns:
             Dictionary with notebook names as keys and their info as values
@@ -398,9 +404,11 @@ class NotebookManager:
             kernel_status = "unknown"
             if kernel:
                 try:
-                    kernel_status = (
-                        "alive" if hasattr(kernel, "is_alive") and kernel.is_alive() else "dead"
-                    )
+                    if hasattr(kernel, "is_alive"):
+                        kernel_status = "alive" if kernel.is_alive() else "dead"
+                    elif isinstance(kernel, dict) and kernel_manager is not None:
+                        kernel_id = kernel.get("id")
+                        kernel_status = "alive" if kernel_id and kernel_id in kernel_manager else "dead"
                 except Exception:
                     kernel_status = "error"
             else:
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -469,6 +469,7 @@ async def list_notebooks() -> (
     return await ListNotebooksTool().execute(
         mode=server_context.mode,
         notebook_manager=notebook_manager,
+        kernel_manager=server_context.kernel_manager,
     )
 
 
--- a/jupyter_mcp_server/tools/list_notebooks_tool.py
+++ b/jupyter_mcp_server/tools/list_notebooks_tool.py
@@ -41,7 +41,7 @@ class ListNotebooksTool(BaseTool):
             return "No notebook manager available."
 
         # Get all managed notebooks
-        managed_notebooks = notebook_manager.list_all_notebooks()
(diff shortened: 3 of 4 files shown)


echarles:

AmirF194:

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging this one too. Appreciate the quick turnaround on all three.

PR #322, opened 2026-07-25
fix(tools): forward reconnect_interval/timeout in use_notebook's MCP_SERVER kernel creation

### LINKED ISSUE #321: use_notebook's MCP_SERVER kernel creation never applies the configured --reconnect-interval
`use_notebook`'s MCP_SERVER kernel creation never applies the configured `--reconnect-interval`

`UseNotebookTool.execute()`'s MCP_SERVER branch creates the kernel with:

```python
kernel = create_jupyter_sandbox_kernel(
    server_url=runtime_url,
    token=runtime_token,
    kernel_id=kernel_id,
    path=notebook_path,
    logger=logger,
)
```

This omits `reconnect_interval` and `timeout`. The two other call sites that create a
kernel the same way both pass them through:

- `utils.py`, `create_kernel()`: `reconnect_interval=getattr(config, "reconnect_interval", 0) or 0`
- `tools/execute_code_tool.py`, `_connect_to_kernel()`: same pattern

So a server started with `--reconnect-interval` gets it applied when `execute_code` opens a
kernel, but not when `use_notebook` does, which is the primary path that actually creates the
kernel connection in MCP_SERVER mode. The configured value is silently dropped on that path.

Repro (no live server needed): monkeypatch `create_jupyter_sandbox_kernel` and set
`reconnect_interval=5` in config, then call `UseNotebookTool().execute(mode=ServerMode.MCP_SERVER, ...)`
with `use_mode="create"`. The captured kwargs never include `reconnect_interval`.

Fault location: `jupyter_mcp_server/tools/use_notebook_tool.py`, `UseNotebookTool.execute()`,
the `create_jupyter_sandbox_kernel(...)` call in the `MCP_SERVER` branch (around line 246).

I have a fix + regression test ready and will open a PR shortly.

AmirF194:
Fixes #321.

## Root cause

`UseNotebookTool.execute()`'s MCP_SERVER branch calls `create_jupyter_sandbox_kernel(...)`
without `reconnect_interval` or `timeout`. The two sibling call sites that create a kernel
the same way, `utils.create_kernel()` and `execute_code_tool.py`'s `_connect_to_kernel`,
both forward `reconnect_interval=getattr(config, "reconnect_interval", 0) or 0` and
`timeout=getattr(config, "execution_timeout", None)`. `use_notebook` is the call site that
actually establishes the kernel connection in MCP_SERVER mode, so a configured
`--reconnect-interval` was silently inert there.

## Fix

Forward both values the same way the sibling call sites already do, so all three
`create_jupyter_sandbox_kernel` call sites stay consistent.

## Verified

Docker (python:3.11-slim, current main `1c52823`): the new regression test fails on main
(`KeyError: 'reconnect_interval'`) and passes on this branch, confirmed the loaded module
is the branch's own file via `__file__`. Also ran the existing `use_notebook`/`restart_notebook`
unit-test files that don't need a live server (`test_use_notebook_create_local.py`,
`test_use_notebook_ui_open.py`, `test_restart_notebook_tool.py`) alongside the new test, all
green. Checked `ruff`/`mypy` on the touched files: both report only pre-existing errors
elsewhere in the file, none on the changed lines.

I did not run the full live-server suite (`tests/conftest.py`'s session-scoped `jupyter_server`
fixture), since this change is scoped to a mocked, no-live-server code path and the fixture
mutates tracked notebook content as a side effect.

### DIFF
--- a/jupyter_mcp_server/tools/use_notebook_tool.py
+++ b/jupyter_mcp_server/tools/use_notebook_tool.py
@@ -243,12 +243,17 @@ class UseNotebookTool(BaseTool):
                 # Ensure the kernel is started with the same path as the notebook.
                 # Routed through code-sandboxes (jupyter variant) rather than a
                 # direct KernelClient; the sandbox creates and starts the kernel.
+                from jupyter_mcp_server.config import get_config
+
+                config = get_config()
                 kernel = create_jupyter_sandbox_kernel(
                     server_url=runtime_url,
                     token=runtime_token,
                     kernel_id=kernel_id,
                     path=notebook_path,
                     logger=logger,
+                    timeout=getattr(config, "execution_timeout", None),
+                    reconnect_interval=getattr(config, "reconnect_interval", 0) or 0,
                 )
 
                 info_list.append(f"[INFO] Connected to kernel '{kernel.id}'.")
--- a/tests/test_use_notebook_reconnect_interval.py
+++ b/tests/test_use_notebook_reconnect_interval.py
@@ -1,0 +1,105 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""use_notebook's MCP_SERVER kernel-creation path must forward the configured
+reconnect_interval/execution_timeout to create_jupyter_sandbox_kernel, the same
+way its sibling call sites (utils.create_kernel, execute_code_tool) already do.
+"""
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for the quick review and merge!

PR #327, opened 2026-07-29
deps: migrate to mcp version 2

echarles:
- [ ] Needs also to monitor other MCP Servers migrations to avoid having incompatible MCP version 1 and 2 in the same python environment
- [ ] Monitor progress on https://github.com/pydantic/pydantic-ai/pull/6738 to avoid upstream consumers breaking.
- [ ] Should Jupyter MCP Version bump to version 2 to clearly indicate breaking MCP client version?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
   ~ BSD 3-Clause License
 -->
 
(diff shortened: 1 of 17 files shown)


echarles:
Closing in favor of https://github.com/datalayer/jupyter-mcp-server/pull/405

PR #333, opened 2026-08-01
fix(auth): fetch anonymous XSRF cookie when no password or token is set

### LINKED ISSUE #183: Support XSRF-protected Jupyter deployments without Bearer tokens

AmirF194:
## Root cause

`ServerContext._init_mcp_server_mode` only ever sets up an `_xsrf` cookie when
`code_sandbox_password` is configured (added in #250). When neither a password
nor a token is set, the `else` branch just builds a `JupyterServerClient` and
`code_sandbox_auth_headers` stays permanently `{}` (`server_context.py:226`),
so every state-changing request (kernel management, the collaboration API)
fails with `'_xsrf' argument missing from POST` on any deployment where auth
is handled externally: `--IdentityProvider.token=''`, JupyterHub single-user
servers, or managed environments (SageMaker Studio, Colab Enterprise). #250's
own body says it "partially addresses #183" and does not cover this case.

## Fix

Added `JupyterAnonymousAuth` (`auth.py`), a `JupyterPasswordAuth` subclass
that performs only the anonymous `GET /login` step to pick up the `_xsrf`
cookie, skipping the password `POST` and its verification. Wired into
`_init_mcp_server_mode`'s existing no-password branch, only when
`code_sandbox_token` is also unset, so the token-authenticated path is
unchanged. Every existing reader of `_code_sandbox_password_auth`
(`code_sandbox_auth_headers`, `relogin_code_sandbox`, the document-auth
reuse when `document_url` matches `code_sandbox_url`) already goes through
`get_headers()` / `relogin()` / `close()`, so no other call site needed to
change.

Scoped to the reported case (code sandbox with no credentials, document
server at the same URL or unset). A document server configured at a
genuinely different URL with no credentials of its own is unchanged, still
silently unauthenticated as before this PR; not attempted here to keep the
diff to the no-token-no-password branch.

**Update:** CI on this PR caught a regression from the first commit. The
anonymous fetch ran eagerly at init time, and several existing tests (e.g.
`test_create_kernel_passes_reconnect_interval`) go through
`ServerContext.get_instance()` with the process-global default config, which
has no token and no password. That combination previously did zero network
I/O; after the first commit it made a real GET and raised when the target
wasn't reachable, failing 6 tests across the matrix. Second commit extracts
the fetch into `_try_anonymous_auth`, which degrades to no cookie on a
connection error instead of raising (explicit password auth is untouched and
still hard-fails as before).

## Verification

- `tests/test_auth.py::TestInitMcpServerModeAnonymousXsrf::test_no_password_no_token_fetches_xsrf`
  fails on unmodified main (`_code_sandbox_password_auth` stays `None`) and
  passes on this branch; a sibling test pins that a configured
  `code_sandbox_token` still skips the anonymous fetch entirely.
- `test_unreachable_server_does_not_abort_init` (added in the second commit)
  fails against the first commit and passes against the second; confirmed the
  6 CI-failing tests (`test_create_kernel_passes_reconnect_interval`,
  `test_create_kernel_no_reconnect_by_default`, all 4 in
  `test_execute_code_timeout.py`) pass locally in `python:3.11-slim` Docker.
- `python:3.11-slim` Docker: full `tests/` (excluding the live-server E2E/
  session-expiry tests, which need a non-root Jupyter Lab and fail identically
  on unmodified main in this container) shows no new failures or errors versus
  main, only the same pre-existing ones.
- `ruff check` on the two changed files: identical finding set to unmodified
  main, nothing new on the added code.
- Not verified: end-to-end against a real Jupyter server started with
  `--IdentityProvider.token=''` (no such server available in this session).

Fixes #183

### DIFF
--- a/jupyter_mcp_server/auth.py
+++ b/jupyter_mcp_server/auth.py
@@ -195,3 +195,36 @@ class JupyterPasswordAuth:
             return
         for name, value in self._live_cookies().items():
             session.cookies.set(name, value)
+
+
+class JupyterAnonymousAuth(JupyterPasswordAuth):
+    """Obtains only the anonymous `_xsrf` cookie, for deployments with no
+    password and no bearer token (SSO/reverse-proxy auth, JupyterHub
+    single-user servers, `--IdentityProvider.token=''`).
+
+    XSRF-protected POST/PUT/DELETE requests need the `_xsrf` cookie Tornado's
+    login handler sets on any GET, even with no credentials being presented,
+    so this performs that GET and skips the password POST and verification
+    entirely. If the server does not enforce XSRF, no cookie is set and
+    `get_headers()` returns an empty dict, same as before this class existed.
+    """
+
+    def __init__(self, server_url: str):
+        super().__init__(server_url, password="")
+
+    def login(self, timeout: float = 10.0) -> None:
+        session = requests.Session()
+        self._session = session
+        try:
+            self._do_request(
+                "GET", f"{self.server_url}/login",
+                stage="anonymous XSRF fetch", timeout=timeout,
+                allow_redirects=False,
+            )
+            self._xsrf_token = session.cookies.get("_xsrf", "")
+            self._authenticated = True
+            logger.info(f"Anonymous XSRF cookie fetch complete for {self.server_url}")
+        except BaseException:
+            session.close()
+            self._session = None
+            raise
--- a/jupyter_mcp_server/server_context.py
+++ b/jupyter_mcp_server/server_context.py
@@ -93,7 +93,7 @@ class ServerContext:
 
         logger.info(f"Initializing MCP_SERVER mode with code_sandbox_url: {code_sandbox_url}")
 
-        from jupyter_mcp_server.auth import JupyterPasswordAuth
+        from jupyter_mcp_server.auth import JupyterAnonymousAuth, JupyterPasswordAuth
 
         try:
             # Code Sandbox auth — password takes precedence over token
@@ -107,6 +107,13 @@ class ServerContext:
                 self._install_code_sandbox_auth_retry(self._server_client)
             else:
                 self._server_client = JupyterServerClient(base_url=code_sandbox_url, token=config.code_sandbox_token)
+                if not config.code_sandbox_token:
+                    # No password and no token, but the server may still require the
+                    # anonymous `_xsrf` cookie on state-changing requests (SSO/reverse-proxy
+                    # auth, JupyterHub single-user servers, --IdentityProvider.token='').
+                    self._code_sandbox_password_auth = JupyterAnonymousAuth(code_sandbox_url)
+                    self._code_sandbox_password_auth.login()
+                    self._code_sandbox_password_auth.inject_into_session(self._server_client.http_client.session)
 
             # Document auth — only needed when the document server is explicitly
             # different from the code sandbox server. When URLs match (or document_url
--- a/jupyter_mcp_server/server_context.py
+++ b/jupyter_mcp_server/server_context.py
@@ -66,6 +66,29 @@ class ServerContext:
         if document is not None and document is not code_sandbox:
             document.close()
 
+    @staticmethod
+    def _try_anonymous_auth(code_sandbox_url):
+        """Fetch the anonymous `_xsrf` cookie.
+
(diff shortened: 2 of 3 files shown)


echarles:
CI Test were failing, I have relaunched, failure again - sounds like intermediate bad state of the github ci worker. TBC

AmirF194:
Not a CI worker flake, the first commit had a real bug: the anonymous XSRF fetch ran eagerly during server init, and in the test suite that hit `ServerContext.get_instance()` with no token/password configured, it made a real network call that failed on connection-refused. Fixed in 15dfeef by degrading to no cookie when the fetch fails instead of raising, with a regression test that fails without the fix. Full matrix is green now, all 15 legs.

echarles:
@AmirF194 This PR has now conflicts further other merges.

AmirF194:
Rebased onto main, conflicts resolved (both new test classes from #343/#346/#349 are now separate, kept independently). CI running now.

echarles (approved):
LGTM Thx @AmirF194

PR #341, opened 2026-08-03
fix(tools): use the authenticated server_client in list_files MCP_SERVER mode

### LINKED ISSUE #340: list_files (MCP_SERVER mode) 403s on password-auth deployments: builds its own unauthenticated client instead of reusing server_client
## Bug

In `MCP_SERVER` mode, `list_files` builds its own `JupyterServerClient` from `get_config()` instead of using the `server_client` that `server.py` already passes into every tool. That fresh client carries no session cookies, so on a deployment secured by `code_sandbox_password` (no static token, the whole point of password auth) it gets a 403 while every other tool succeeds through the shared, authenticated client.

`jupyter_mcp_server/tools/list_files_tool.py:203-209`:

```python
elif mode == ServerMode.MCP_SERVER:
    config = get_config()
    server_client = JupyterServerClient(
        base_url=config.code_sandbox_url, token=config.code_sandbox_token
    )
    all_files = _list_files_mcp(server_client, path, 0, None, max_depth)
```

`server.py:356` already passes the authenticated client in as `server_client=server_context.server_client`, and `server_context.py:96-108` is where that client is built and logged in via `JupyterPasswordAuth` when a password is configured. `list_files_tool.py` shadows the parameter with a new, unauthenticated one instead of using it.

## Reproduction

Started a real `jupyter_server` with `--ServerApp.password` set and no token (password-only auth, the deployment class `code_sandbox_password` exists for). Called `list_files` through the MCP server in `MCP_SERVER` mode: it returns a single `error` row, `Error: 403: Forbidden`. Every other tool (`list_notebooks`, `use_notebook`, `execute_code`, ...) succeeds against the same server, because they receive and reuse the shared authenticated client while `list_files` builds its own.

## Expected

`list_files` should reuse the authenticated `server_client` like every other MCP_SERVER-mode tool, and list files successfully on password/XSRF-only deployments.

AmirF194:
## Root cause

`ListFilesTool.execute()`'s `MCP_SERVER` branch built a fresh `JupyterServerClient(base_url=config.code_sandbox_url, token=config.code_sandbox_token)` from `get_config()` instead of using the `server_client` parameter already passed in from `server.py:356` (`server_client=server_context.server_client`). That fresh client carries no session cookies, so on any deployment secured by `code_sandbox_password` (no static token, the case password auth exists for) it 403s while every other MCP_SERVER tool succeeds through the shared, authenticated client built in `server_context.py`.

## Fix

Reuse the passed-in `server_client` (matching every other tool's wiring) and drop the now-unused `get_config()` construction.

## Verification

- Added `test_password_auth_list_files` to `tests/test_auth.py::TestPasswordAuthE2E`, which fails on main (`Error: 403: Forbidden`) and passes with this fix, run against a real password-protected `jupyter_server`.
- Full `tests/test_auth.py` suite passes (42/42).
- `ruff check` and `mypy` on the changed files show the same pre-existing findings before and after this diff; no new issues introduced.
- Did not exercise the `contents_manager`/`JUPYTER_SERVER` branch or the rest of the MCP tool surface; this change only touches the `MCP_SERVER` branch of `list_files`.

Fixes #340

### DIFF
--- a/jupyter_mcp_server/tools/list_files_tool.py
+++ b/jupyter_mcp_server/tools/list_files_tool.py
@@ -10,7 +10,6 @@ from typing import Any
 from jupyter_core.utils import ensure_async
 from jupyter_server_client import JupyterServerClient
 
-from jupyter_mcp_server.config import get_config
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 from jupyter_mcp_server.utils import format_TSV
 
@@ -200,12 +199,9 @@ class ListFilesTool(BaseTool):
         if mode == ServerMode.JUPYTER_SERVER and contents_manager is not None:
             # Local mode: use contents_manager directly
             all_files = await _list_files_local(contents_manager, path, max_depth)
-        elif mode == ServerMode.MCP_SERVER:
-            # Remote mode: use HTTP client
-            config = get_config()
-            server_client = JupyterServerClient(
-                base_url=config.code_sandbox_url, token=config.code_sandbox_token
-            )
+        elif mode == ServerMode.MCP_SERVER and server_client is not None:
+            # Remote mode: reuse the shared, authenticated HTTP client
+            # (a client built fresh here carries no session cookies).
             all_files = _list_files_mcp(server_client, path, 0, None, max_depth)
         else:
             raise ValueError(f"Invalid mode or missing required clients: mode={mode}")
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @AmirF194

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for the quick review, appreciate it.

PR #343, opened 2026-08-04
fix(auth): retry code sandbox requests after a session-cookie expiry

### LINKED ISSUE #342: code sandbox HTTP calls have no retry-on-session-expiry, unlike the document/collaboration path
`ServerContext.relogin_code_sandbox()` (`server_context.py:259`) re-authenticates the code sandbox password session and re-injects fresh cookies, but nothing calls it directly: its only caller is `relogin_document()`'s shared-auth branch (`server_context.py:288-289`), which only fires when `document_url` and `code_sandbox_url` point at the same server.

The document/collaboration path already survives a session expiry: `NotebookConnection._get_ws_url` (`notebook_manager.py:84-114`) catches a 401/403 from the collaboration session request, calls `relogin_document()`, and retries once.

Every tool call that goes through `server_client` directly has no equivalent protection: `kernels.list_kernels()` / `kernelspecs.list_kernelspecs()` (`execute_code_tool.py:65`, `list_kernels_tool.py:22,28`), `contents.list_directory()` / `create_notebook()` (`list_files_tool.py:50`, `use_notebook_tool.py:79,81,171,252,257`). A password-authenticated session that expires or is otherwise invalidated mid-run makes any of these fail with an uncaught `AuthenticationError`/`ForbiddenError` from `jupyter_server_client`, instead of transparently re-authenticating the way the document path does.

Reproduced with a stubbed `http_client.request` that raises `AuthenticationError` on the first call: the exception propagates straight through `list_kernels_tool.py`'s `except Exception as e: raise RuntimeError(...)` wrapper with no retry attempt.

AmirF194:
`relogin_code_sandbox()` re-authenticates the code sandbox password session, but the only thing that ever calls it is `relogin_document()`'s shared-auth branch, which fires only when `document_url` and `code_sandbox_url` are the same server. Every tool call that goes through `server_client` directly (`list_kernels`, `list_files`, `create_notebook`, kernelspecs, get_status) has no protection against an expired or invalidated password-auth session, unlike the document/collaboration path, which already retries once on a 401/403 (`NotebookConnection._get_ws_url`).

`JupyterServerClient`'s `kernels`/`contents`/`kernelspecs`/`sessions` managers all funnel through one method, `http_client.request()`. This wraps that single choke point so a 401/403 triggers one `relogin_code_sandbox()` call and a retry, installed once when the password-authenticated `server_client` is built. No call site changes needed, and it covers every current and future caller through `server_client`, not just the five identified in the issue.

- New `TestCodeSandboxAuthRetry` (retry-then-succeed, persistent-failure-propagates, no-retry-when-healthy) fails with `AttributeError` on unmodified `main` and passes on this branch (`python:3.11-slim`, `jupyter-server-client==0.1.1`).
- Full `tests/test_auth.py`: 45 passed, same container.
- `ruff check` + `mypy` on both changed files: identical error count to unmodified `main` (22 ruff / the same missing-stubs class on `jupyter_server_client`), so nothing new; `header-license-fix` clean except the pre-existing `.github/copilot-instructions.md` failure.
- Not verified: a genuine end-to-end 401 through a real password-authed server. The short-cookie fixture this suite uses elsewhere can't trigger one here, because `inject_into_session` copies cookies by name/value only, so the copy in `server_client.http_client.session` never carries the original cookie's expiry; the mocked-exception tests above are what stand behind this fix instead.

Fixes #342

### DIFF
--- a/docs/docs/reference/security/index.mdx
+++ b/docs/docs/reference/security/index.mdx
@@ -182,6 +182,10 @@ This is useful when:
 3. These cookies are injected into all HTTP requests (API calls, kernel operations) and WebSocket connections (notebook collaboration)
 4. The `X-XSRFToken` header is automatically included in requests that require XSRF protection
 
+### Session Expiry Recovery
+
+Password-authenticated sessions can expire or be invalidated while the MCP server is running (server restart, cookie TTL, etc). When a request to the code sandbox server returns 401 or 403, the MCP server performs the `/login` flow again to obtain a fresh session, then retries the original request once. This only re-runs the login (a new session, not the old expired one), so it recovers automatically instead of surfacing an auth error on every subsequent tool call. The document/collaboration path has the equivalent behavior on its own WebSocket connection.
+
 ### Configuration
 
 #### Simplified (Same Password for Both Servers)
--- a/jupyter_mcp_server/server_context.py
+++ b/jupyter_mcp_server/server_context.py
@@ -104,6 +104,7 @@ class ServerContext:
                     logger.warning("Both code_sandbox_password and code_sandbox_token are set. Password auth takes precedence.")
                 self._server_client = JupyterServerClient(base_url=code_sandbox_url, token=None)
                 self._code_sandbox_password_auth.inject_into_session(self._server_client.http_client.session)
+                self._install_code_sandbox_auth_retry(self._server_client)
             else:
                 self._server_client = JupyterServerClient(base_url=code_sandbox_url, token=config.code_sandbox_token)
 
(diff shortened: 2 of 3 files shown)


echarles:
CI is failing for some tests.

@AmirF194 Can you remind here why we needed that `relogin` (if a token is expired or becomes invalid, trying a second time will not change that, right?) - Worth documented that on https://jupyter-mcp-server.datalayer.tech/reference/security/ as part of this PR

Thank you so much!

AmirF194:
The failures are environmental, not from this diff: `main` at 6d8f5bd3 (2026-08-04) has the identical CI failure, all six matrix legs, same root cause (`RuntimeError: UndoManager must be created with doc or scopes`, 476 occurrences in the ubuntu/3.10 log). This PR only touches `http_client.request()` in the auth module and doesn't change anything on the collaboration/UndoManager path.

On the relogin question: you're right that retrying the same expired session changes nothing, but `relogin_code_sandbox()` doesn't retry with the old token, it re-authenticates first. It re-runs the password login flow against `code_sandbox_url` and installs a fresh session cookie into `server_client.http_client.session` before the retried request goes out, so the retry uses a new, valid session rather than the one that just 401'd. Same shape as the document/collaboration path already has (`NotebookConnection._get_ws_url`'s one retry on 401/403), just applied to the direct `server_client` calls that had no equivalent protection.

Added a line to the security reference doc explaining this.

echarles:
> The failures are environmental, not from this diff

Is CI failing also now on main ?

AmirF194:
Yes, still failing on main right now: HEAD is still 6d8f5bd3 (no new commits since), and its own Test workflow run is red on every matrix leg (ubuntu, windows, macos, both Python versions), same UndoManager error. Nothing has changed there since my last comment.

echarles:
mmh CI for latest merge was green https://github.com/datalayer/jupyter-mcp-server/pull/341/checks

AmirF194:
That checks page is showing #341's pre-merge run against its own branch head (`1a26a55d`, tested 2026-08-03T18:46Z), not the merge commit. The merge itself (`6d8f5bd3`, merged 2026-08-04T06:31:39Z) triggered a separate post-merge Test run about 12 hours later, and that one is red on all six matrix legs: https://github.com/datalayer/jupyter-mcp-server/actions/runs/30884306512

So there's a real gap between the branch that was tested and what actually landed on main, most likely a floating dependency drifting in that 12-hour window (the failures are all `UndoManager must be created with doc or scopes` from the collaboration/Yjs path, nothing #341 or this PR touch). Main is still on `6d8f5bd3` right now with no new commits, so this is reproducible: `git checkout 6d8f5bd3 && make test-mcp-server` should show it directly.

AmirF194:
The UndoManager failures across our recent PRs turned out to be a real upstream break: jupyter-collaboration 5.0.0 (released Aug 3) versus this repo's pin. #347 fixed that by pinning back to <5.

Rebased this branch onto main to pick it up. CI is running now: https://github.com/datalayer/jupyter-mcp-server/pull/343/checks. Ran tests/test_auth.py locally first, 36/36 of the non-E2E tests pass (the 9 E2E ones need a live jupyter binary on PATH, an environment gap in my container, not this repo's CI).

echarles:
In the meantime https://github.com/datalayer/jupyter-mcp-server/pull/349 is merged and is using latest jupyter-collaboration. Will be great to rebase on latest main.

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging, and for pulling in the #347/#349 pin fix along the way. Appreciate the back and forth on tracking down the CI flakiness.

PR #346, opened 2026-08-05
fix(use_notebook): reject create at a path that already exists

### LINKED ISSUE #345: use_notebook(use_mode="create") silently overwrites an existing notebook file
`use_notebook(use_mode="create")` never checks whether `notebook_path` already exists before writing a fresh notebook there, in both server modes.

`UseNotebookTool._check_path_local`/`_check_path_http` only compute `file_exists` when `use_mode == "connect"`; for `"create"` they verify only that the parent directory exists and return `(True, None)` unconditionally. Execution then goes straight to `contents_manager.new(model=..., path=notebook_path)` (JUPYTER_SERVER) or `server_client.contents.create_notebook(notebook_path, content=content)` (MCP_SERVER), both of which are plain overwrite operations on jupyter_server's own contents manager when the path already resolves to a file.

In practice this means a second, independent `use_notebook(create)` call at a path some other session already has real work in (the tool has no way to know, since the "already created" guard above it only checks its own in-memory `NotebookManager`, not the filesystem) destroys that work with no warning and no error.

Reproduced against a real `jupyter_server.services.contents.filemanager.AsyncFileContentsManager` (not a stand-in): a notebook containing one code cell is written directly, then `UseNotebookTool().execute(mode=JUPYTER_SERVER, use_mode="create", notebook_path=<same path>, ...)` is called from a fresh `NotebookManager`. The original cell is gone afterward, replaced by the tool's "New Notebook Created by Jupyter MCP Server" skeleton.

```
$ python -m pytest tests/test_use_notebook_create_overwrite_repro.py -v
...
AssertionError: use_notebook(create) silently overwrote an existing notebook: cells on disk are now ['New Notebook Created by Jupyter MCP Server']
```

Will open a PR mirroring the existing-path check the `"connect"` branch already has.

AmirF194:
`use_notebook(use_mode="create")` never checks whether `notebook_path` already exists before writing to it, in either server mode. `_check_path_local`/`_check_path_http` only compute `file_exists` for `use_mode == "connect"`; the `"create"` branch verifies only that the parent directory exists and returns `(True, None)` unconditionally. Execution then goes straight to `contents_manager.new(...)` (JUPYTER_SERVER) or `server_client.contents.create_notebook(...)` (MCP_SERVER), both plain overwrites when the path already resolves to a file. The guard above that path check (`notebook_name in notebook_manager`) only catches the case where the same client already has this notebook open under its own in-memory manager, so a second, independent client creating at a path someone else's session already has real work in destroys that work with no warning.

Fix: compute `file_exists` for both modes, and for `"create"` return an error when the path is already taken, mirroring the message `"connect"` already gives for the opposite case.

## Verification

- New test `tests/test_use_notebook_create_overwrite_repro.py` drives `UseNotebookTool` against jupyter_server's own `AsyncFileContentsManager` (not a stand-in) pointed at a real temp directory: fails on `main` (a pre-seeded notebook's cell is silently replaced by the create skeleton), passes on this branch.
- Existing `tests/test_use_notebook_create_local.py` and `tests/test_use_notebook_reconnect_interval.py` still pass unchanged (their fake managers report an empty directory, so the new check is a no-op for them, as it should be).
- Ran the full `make test-mcp-server` suite: this branch reproduces the same 30 failed / 9 errors as unmodified `main` in an identical container, so nothing here regresses (or fixes) those; they are pre-existing and unrelated to this file.
- Not verified: the MCP_SERVER (HTTP) path against a live Jupyter server; the fix there follows the same pattern the existing `"connect"` check on that function already uses, exercised in this PR only via the fake-`server_client` unit tests.

Fixes #345

### DIFF
--- a/jupyter_mcp_server/tools/use_notebook_tool.py
+++ b/jupyter_mcp_server/tools/use_notebook_tool.py
@@ -80,12 +80,18 @@ class UseNotebookTool(BaseTool):
             else:
                 dir_contents = server_client.contents.list_directory("")
 
+            file_exists = any(file.name == path.name for file in dir_contents)
             if mode == "connect":
-                file_exists = any(file.name == path.name for file in dir_contents)
                 if not file_exists:
                     return (
                         False,
                         f"'{notebook_path}' not found in jupyter server, please check the notebook already exists.",
+                    )
+            elif mode == "create":
+                if file_exists:
+                    return (
+                        False,
+                        f"'{notebook_path}' already exists in jupyter server, please use a different path or use_mode='connect' to open it.",
                     )
 
             return True, None
@@ -113,12 +119,18 @@ class UseNotebookTool(BaseTool):
                 contents_manager.get(parent_path, content=True, type="directory")
             )
 
+            file_exists = any(item["name"] == path.name for item in model.get("content", []))
             if mode == "connect":
-                file_exists = any(item["name"] == path.name for item in model.get("content", []))
                 if not file_exists:
                     return (
                         False,
                         f"'{notebook_path}' not found in jupyter server, please check the notebook already exists.",
+                    )
+            elif mode == "create":
+                if file_exists:
+                    return (
+                        False,
+                        f"'{notebook_path}' already exists in jupyter server, please use a different path or use_mode='connect' to open it.",
                     )
 
             return True, None
--- a/tests/test_use_notebook_create_overwrite_repro.py
+++ b/tests/test_use_notebook_create_overwrite_repro.py
@@ -1,0 +1,78 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Repro: use_notebook(use_mode="create") against a REAL jupyter_server
+contents manager silently overwrites an existing notebook file on disk.
+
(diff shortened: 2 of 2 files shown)


echarles:
Thx for the analysis, very helpful, the source is indeed a new release of jupyter-collaboration on 3 Aug.

I have merged https://github.com/datalayer/jupyter-mcp-server/issues/348 with a temporary workaround to pin jupyter-collaboration 

I guess you can rebase and see how CI behaves now.

AmirF194:
Rebased onto main (picked up #347's pin). CI is running now: https://github.com/datalayer/jupyter-mcp-server/pull/346/checks

Ran the targeted regression tests and the mcp_server-mode unit suite in a clean container first: the `UndoManager` failures from the old jupyter-collaboration 5.0.0 are gone with jupyter-collaboration pinned back to 4.4.2, and the 5 tests for this fix still pass.

echarles:
Awesome, can you rebase again on latest main?

echarles (approved):
LGTM Thx @AmirF194

AmirF194:
Thanks for merging and for the guidance on rebasing through the collaboration pin fix.

PR #350, opened 2026-08-06
docs: generated MCP API reference (22 tools + 1 prompt) at /mcp/

antheducation:
Hi! While wiring jupyter-mcp-server into an agent I kept wanting a per-tool reference showing the exact schemas the server advertises over the wire. The handwritten page at `docs/docs/reference/tools` is genuinely good prose, but it's one long hand-maintained page — every new tool or parameter has to be re-described by hand, and there's no machine check that it still matches what `tools/list` returns.

This PR adds a **generated MCP reference** at [`/mcp/`](https://jupyter-mcp-server.datalayer.tech/mcp/), built from the server itself rather than written by hand.

### How it works

- `mcp-parser` snapshots the running server over stdio (`tools/list`, `prompts/list`, `resources/list`) into `docs/sourcey/mcp.json` — 22 tools + 1 prompt, protocol revision `2025-11-25`.
- A small generator turns that snapshot into one page per tool/prompt — parameters, defaults, annotations (read-only / destructive hints), JSON-RPC + Python call samples — and links each page to the file whose `@mcp.tool` decorator registers it.
- A generated **Configuration** page covers all 23 `JupyterMCPConfig` settings (from the pydantic model), both transports, and both serving modes.
- [Sourcey](https://www.npmjs.com/package/sourcey) renders it into a static site (with search, `llms.txt` / `llms-full.txt` for agent consumption, and an interactive "MCP Explorer" tab over the same snapshot).

### Nothing generated is committed

`docs/sourcey/` holds **source only**: the snapshot, the source map, the config dump, the generated markdown, and the pipeline. The site itself is built during the docs build — `npm run build` in `docs/` runs `sourcey/build_site.mjs` via `prebuild`, which renders into `docs/static/mcp/`, and Docusaurus copies that into its output like any other static asset. `docs/static/mcp/` is git-ignored. There is no runtime component.

OG image generation is disabled, so the reference produces no images at all — on disk or in git.

### CI keeps it honest

`.github/workflows/docs.yml` re-runs the whole generation on every PR (live stdio snapshot → source map → config dump → pages) and **fails on any diff** against what is checked in. It then builds the docs and asserts that the reference adds no images and that nothing under `docs/static/mcp` is tracked.

So the reference cannot silently drift from the code. Regenerating is four commands (`docs/sourcey/README.md`) and CI tells you when it is needed.

Only addition to the existing site is one navbar link.


echarles:
Thx for the contribution @antheducation 

Do I understand that it will add docs at runtime? I see generated content begin committed with images also. Is there a way to get those static pages being automatically generated, a bit like swagger openapi is doing for REST endpoints in other projects?

antheducation:
Thanks for taking a look @echarles!

**No runtime component** — everything under `docs/static/mcp/` is plain prebuilt files. Docusaurus copies `docs/static/` into the site build verbatim, so `/mcp/` is served like any other static asset; nothing generates at docs-serve or server-runtime time. (The images you see committed are the pages' OG/social preview cards, also prebuilt.)

The generation happens offline, in `docs/sourcey/`:
1. `snapshot.mjs` speaks real MCP over stdio to the actual `jupyter-mcp-server` binary and captures `tools/list` / `prompts/list` into `mcp.json` — so the reference shows exactly what the server advertises on the wire, not hand-written schemas;
2. `build_pages.mjs` renders all of `docs/static/mcp/` deterministically from that snapshot (+ `sourcemap.json` linking each tool to its `@mcp.tool` decorator at a pinned commit).

**And yes — it can absolutely run swagger-style instead of committing output.** Two levels, both small changes to the docs workflow:

1. **Render in CI, keep the snapshot committed:** treat `mcp.json` (+ `sourcemap.json`, `config-fields.json`) as the committed source of truth and run `build_pages.mjs` as a step before `docusaurus build`. Only needs Node in the docs job (`npm install` in `docs/sourcey`), and all ~7k lines of generated HTML/CSS/PNG drop out of the repo.
2. **Fully live:** the same job also re-runs `snapshot.mjs` against the server installed from the same commit, so the reference can never drift from the code — that's the closest analogue to what swagger does for REST.

I committed the built output in this PR mainly so it works with zero CI changes and is reviewable as-is, but I'd actually prefer option 1 (or 2) myself. Happy to push either onto this branch — just say which you'd like.

echarles:
Having the CI run the generation and fail for any diff would help.
Is it possible to disable the image generation? In general I avoid adding images in the git history when possible.
Is it possible to have the generation in the source, and have docusaurus build from source the dist so dist is not added in the git tree?
Can you share a screenshot of how it looks like?

antheducation:
Thanks @echarles — all three are done, and the branch is rebased onto current `main`. The diff is down from 93 files to 39, with no binary files.

**On the runtime question, plainly: nothing is generated at runtime.** Not by the MCP server, and not when the docs are served. The pages are produced at **docs build time** and are plain static HTML by the time anything is deployed. What changed is *where* that happens: it used to be an offline step whose output I committed, and it is now a step of the Docusaurus build itself.

**1. Image generation disabled.** Sourcey's per-page OG cards are turned off (`generateOgImages: false` in the new `docs/sourcey/build_site.mjs`), so none are produced at all — not in git, not on disk, not in the deployed site. The 26 PNGs previously under `docs/static/mcp/_og/` are gone from the branch. CI asserts the reference emits no image files.

**2. Generation in source, dist built by Docusaurus.** `docs/sourcey/` now holds source only: the protocol snapshot, the source map, the config dump, the generated markdown, and the pipeline. `build_site.mjs` renders that into `docs/static/mcp/`, which is git-ignored and produced by `npm run build` through a `prebuild` script (and `prestart`, so `npm start` works too). `sourcey` and `mcp-parser` are added as `devDependencies` in `docs/package.json`. `make build` / `make deploy` are unchanged and still emit `/mcp/` exactly as before — I ran the full `docusaurus build` locally to confirm.

**3. CI regenerates and fails on any diff.** New `.github/workflows/docs.yml`, following the style of `build.yml`: install the package and `ext/sandboxes`, take a live stdio snapshot of the server, regenerate the source map, config dump and pages, then `git diff --exit-code`. It then builds the docs and asserts no images and nothing tracked under `docs/static/mcp`.

One thing that gate forced, worth flagging since it is a behaviour change you did not ask for. For "fail on any diff" to be usable, the generated output has to be a function of the *source tree*, not of a commit id — otherwise it goes red every time anything moves. So source links now point at `main` instead of a pinned SHA, and no longer carry `#L<line>` anchors; all 19 tool decorators live in `server.py`, so with line numbers any unrelated edit above them would have failed the docs job. Pages now link the defining file and name the decorated function. If you would rather keep exact line links and accept the churn, say so and I will put them back.

The rebase also proved the gate works: re-snapshotting against current `main` picked up the `stream` and `progress_interval` parameters added to `execute_cell` / `execute_code` since I opened this, which the old committed snapshot had missed.

**Screenshots.** I cannot attach an image to a comment, and `/mcp/` is not deployed yet, so these are from the local `docusaurus build` output. They live on a **separate branch of my fork** and are deliberately not part of this PR, so no image enters this repository's history:

Landing page (`/mcp/`):

MCP reference overview

A tool page (`/mcp/tools/execute-cell/`) — note the freshly snapshotted `stream` / `progress_interval` rows:

execute_cell tool page

Happy to adjust anything further.

antheducation:
Worth reporting: the new docs job failed on its first two runs, and both failures were real bugs in the generation rather than problems with the check.

1. Tool descriptions are Python docstrings, and CPython 3.13 dedents docstrings at compile time while 3.12 and earlier keep the source indentation — so the same server produced a different `mcp.json` depending on the interpreter. `snapshot.mjs` now normalises descriptions, making the snapshot a function of the server only.
2. `pip install ./ext/sandboxes` leaves a setuptools copy at `ext/sandboxes/build/lib/`, so the decorator scan saw the four sandbox tools twice and traversal order decided which path won. `gen_sourcemap.py` now prunes build directories and raises if one name is ever found in two files.

Neither would have been visible without the check. `mcp_reference` is green on `c558b03`.

echarles (approved):
LGTM Thx @antheducation

PR #351, opened 2026-08-06
fix(use_notebook): route document operations to the document server when split

### LINKED ISSUE #344: `use_notebook` cannot use a local notebook with a remote kernel
## Description

I want to run notebooks in my local project directory with a kernel running on a remote GPU server.

When the notebook server and the code sandbox server are different, `use_notebook` tries to access the notebook on the code sandbox server instead of the document server.

Both servers are healthy, and the remote kernel itself works.

## Reproduce

1. Start a local Jupyter server with `jupyter-collaboration` enabled. Use it to serve the project directory containing the notebook.
2. Start or connect to a remote Jupyter server with a GPU kernel.
3. Configure the MCP server with separate document and code sandbox servers:

   ```text
   DOCUMENT_URL=http://localhost:48889
   DOCUMENT_TOKEN=...
   CODE_SANDBOX_URL=https://<gpu-host>:8888
   CODE_SANDBOX_TOKEN=...
   ```

   In version 1.1.4, use `RUNTIME_URL` and `RUNTIME_TOKEN` instead of `CODE_SANDBOX_URL` and `CODE_SANDBOX_TOKEN`.

4. Use `use_notebook` to connect to the local notebook.
5. See an error such as:

   ```text
   404 Not Found ... /api/collaboration/session/<notebook>.ipynb
   ```

The same workflow works when `DOCUMENT_URL` and `CODE_SANDBOX_URL` point to the same server.

The error and the request URL show that the collaboration request is sent to the remote code sandbox server.

## Expected behavior

`use_notebook` should use the local document server for the notebook and the remote code sandbox server only for the kernel.
Notebook changes and outputs should be saved to the local project directory.

## Context

- Datalayer version: `jupyter-mcp-server` 1.1.4
- Operating System and version: macOS 15.7.4
- Browser and version: Not applicable. The issue occurs in an MCP tool call, not in a browser.

The setup uses a local document server and a remote GPU server on RunPod.
The issue occurs in MCP server mode and only when the two URLs differ.

I have implemented a possible fix in my fork and am testing it in a real setup.
I plan to open a PR after this testing is complete.
This issue is not intended to block or interfere with work by other contributors.

Possible fix branch: https://github.com/pokutuna/jupyter-mcp-server/tree/fix-collab-session-document-url

<details><summary>Browser Output</summary>
<pre>
Not applicable. The issue occurs in an MCP tool call, not in a browser.
</pre>
</details>

pokutuna:
Fixes #344

## Summary

Fix `use_notebook` for setups that use separate document and code sandbox servers.

## Changes

- Route notebook existence checks, creation, collaboration registration, and optional UI opening to the document server when the URLs differ.
- Keep kernel discovery, startup, and execution on the code sandbox server.
- Use document-server authentication for document requests and never reuse code sandbox XSRF headers.
- Preserve the caller-supplied client when both URLs point to the same server.

## Verification

- I used this for a day of regular work with a local notebook and a remote RunPod GPU kernel, and it behaved as expected throughout.
- Beyond that, I had an AI agent verify several connection patterns, including a Kaggle remote kernel.
- Added `tests/test_use_notebook_split_servers.py`.
- `make test-mcp-server` and `make test-jupyter-server` pass.

### AI Verification Notes

<details>
<summary>Split-server configurations exercised</summary>

The notebook existed only on the document server in every run.

| Document server | Code sandbox |
| --- | --- |
| local, **password** | local, token |
| local, token | RunPod, **password** |
| local, token | **Kaggle** |

Confirmed in all three:

- The collaboration session is established on the document server — this is the request that returned `404 /api/collaboration/session/` in #344.
- `use_mode="create"` writes the notebook to the document server only; it never appears on the code sandbox.
- Kernels start on the code sandbox and appear in its `/api/kernels`.

Auth-specific:

- Code sandbox password auth: its cookies and `X-XSRFToken` are absent from the document client, including after notebook creation.
- Document password auth: login cookies are injected into the document client, and re-injected after `relogin_document()`.

</details>

<details>
<summary>Regression coverage of the added tests</summary>

Reverting `use_notebook_tool.py` to the pre-fix version fails 4 of the 5 tests. The one that still passes is `test_same_url_keeps_caller_client`, which covers the non-split path that this change intentionally leaves untouched.

`execute()` swallows collaboration failures, so `test_connect_to_notebook_only_on_the_document_server` opens the connection explicitly rather than relying on the returned message.

</details>

### DIFF
--- a/jupyter_mcp_server/server_context.py
+++ b/jupyter_mcp_server/server_context.py
@@ -23,6 +23,7 @@ class ServerContext:
     _kernel_spec_manager = None
     _session_manager = None
     _server_client = None
+    _document_client = None
     _kernel_client = None
     _code_sandbox_password_auth = None
     _document_password_auth = None
@@ -50,6 +51,7 @@ class ServerContext:
             cls._instance._kernel_spec_manager = None
             cls._instance._session_manager = None
             cls._instance._server_client = None
+            cls._instance._document_client = None
 
     def _close_auth(self):
         """Close auth sessions if present; safe to call multiple times.
@@ -237,6 +239,32 @@ class ServerContext:
         return self._server_client
 
     @property
+    def document_client(self):
+        """Return an HTTP client for document server operations.
+
+        Reuse the server client when ``document_url`` is unset or matches the
+        code sandbox URL; otherwise create a client with document auth.
+        """
+        if not self._initialized:
+            self.initialize()
+        if self._document_client is not None:
+            return self._document_client
+        config = get_config()
+        document_url = config.document_url
+        if (not document_url) or (document_url == config.code_sandbox_url):
+            return self._server_client
+        if (
+            self._document_password_auth is not None
+            and self._document_password_auth is not self._code_sandbox_password_auth
+        ):
+            client = JupyterServerClient(base_url=document_url, token=None)
+            self._document_password_auth.inject_into_session(client.http_client.session)
+        else:
+            client = JupyterServerClient(base_url=document_url, token=config.document_token)
+        self._document_client = client
+        return self._document_client
+
+    @property
     def kernel_client(self):
         if not self._initialized:
             self.initialize()
@@ -347,6 +375,10 @@ class ServerContext:
             self.relogin_code_sandbox(timeout=timeout)
             return
         self._document_password_auth.relogin(timeout=timeout)
+        if self._document_client is not None:
+            self._document_password_auth.inject_into_session(
+                self._document_client.http_client.session
+            )
         logger.info("Document session re-authenticated after cookie expiry.")
 
     @property
--- a/docs/docs/code-sandboxes/kaggle/index.mdx
+++ b/docs/docs/code-sandboxes/kaggle/index.mdx
@@ -103,6 +103,48 @@ code sandbox values come from an active browser session:
 You can parse `CODE_SANDBOX_URL` and `CODE_SANDBOX_ID` from that URL automatically by
 setting `CODE_SANDBOX_CHANNELS_URL`.
 
+Alternatively, Kaggle's
+[external editor support](https://www.kaggle.com/product-announcements/567740)
+exposes the URL directly:
+
+1. Open your notebook on [kaggle.com](https://www.kaggle.com) and start an
+  interactive session.
+2. Open **Run** -> **Kaggle Jupyter Server**.
+3. Under **Manually Connect**, copy the **VS Code Compatible URL** — this is
+  your `CODE_SANDBOX_URL`.
+
+> The URL embeds a session JWT: treat it as a secret and don't commit it. It
+> stops working once the interactive session ends (Kaggle stops idle sessions
+> after a while).
+
+## Using it with a coding agent
+
+The interactive URL above is a standard token-authenticated Jupyter server
+endpoint, so it works as a plain `CODE_SANDBOX_URL` — no `SANDBOX_VARIANT`,
+sandbox extension, or `CODE_SANDBOX_TOKEN` required. Point `DOCUMENT_URL` at a
+local JupyterLab to keep the notebook on your machine while cells run on
+Kaggle:
+
+```json
+{
+  "mcpServers": {
+    "jupyter": {
+      "command": "uvx",
+      "args": ["jupyter-mcp-server@latest"],
+      "env": {
+        "DOCUMENT_URL": "http://localhost:8888",
+        "DOCUMENT_TOKEN": "MY_TOKEN",
+        "CODE_SANDBOX_URL": "https://kkb-production.jupyter-proxy.kaggle.net/k/12345678/eyJ.../proxy"
+      }
+    }
+  }
+}
+```
+
+The local JupyterLab needs `jupyter-collaboration` installed; without it,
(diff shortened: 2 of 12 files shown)


echarles:
> Beyond that, I had an AI agent verify several connection patterns, including a Kaggle remote kernel.

Awesome! We have an example with a pydantic-ai agent - would you add a similar example to illustrate that case in https://github.com/datalayer/jupyter-mcp-server/tree/main/examples ?

pokutuna:
@echarles 
Sure!
One question: I verified this from a local coding agent (claude) via the MCP config, not with pydantic-ai or a custom-built agent.
So I'm thinking of an example showing how to work on a local notebook with a coding agent while the code runs on a Kaggle Jupyter server (Kaggle URL steps + MCP config). Does that sound good?
There wouldn't be much to put in the directory, maybe just a README with a masked mcp.json, or a example notebook that ran on Kaggle, if that.

pokutuna:
There is not enough for a separate example, so I am thinking of adding a short section to the existing [docs/providers/kaggle/index.mdx](https://jupyter-mcp-server.datalayer.tech/providers/kaggle/) page instead.

echarles:
> There is not enough for a separate example, so I am thinking of adding a short section to the existing [docs/providers/kaggle/index.mdx](https://jupyter-mcp-server.datalayer.tech/providers/kaggle/) page instead.

That would help also :+1:

jupyter_mcp_server/server_context.py
```diff
@@ -50,6 +51,7 @@ def reset(cls):
             cls._instance._kernel_spec_manager = None
             cls._instance._session_manager = None
             cls._instance._server_client = None
```
echarles, line 53: If there is a `document_client`, should `server_client` be renamed to `code_sandbox_client` as dealing with non document interactions?

ebarkhordar (commented):
`use_notebook` stops working on a setup that sets `CODE_SANDBOX_URL` and leaves `DOCUMENT_URL` alone, because `document_url` has a non-empty default.

`config.document_url` defaults to `http://localhost:8888` (`config.py:73`, and `resolve_url_and_token_variables` in `utils.py:171` defaults the same way), so `config.document_url != config.code_sandbox_url` cannot tell "the operator asked for a split" apart from "the operator never mentioned a document server". In the second case the contents check goes to `localhost:8888`, where there is usually nothing listening.

That second case is the documented shape for the `kaggle`, `colab` and `datalayer` engines: the engine table in `docs/docs/reference/configuration/index.mdx` lists `CODE_SANDBOX_URL` plus ids/tokens for each of them and no `DOCUMENT_URL`, and the Colab page says notebook and cell tools "run against the Colab code sandbox transparently".

Measured at `b2b9523` against the merge base `5d0922a`, clean container, config set to `code_sandbox_url=<kaggle proxy url>` and `code_sandbox_token` and nothing else:

```
BASE  result = "[INFO] Connected to kernel 'kernel-1'.
                [INFO] Successfully activate notebook 'nb'."
      caller client contents.listed = ['work']

HEAD  result = "Failed to check the path 'work/nb.ipynb': Connection error to
                http://localhost:8888/api/contents/work: HTTPConnectionPool(
                host='localhost', port=8888): Max retries exceeded"
      caller client contents.listed = []
```

It fails in `_check_path_http`, before any kernel work, so the whole call is lost rather than just the document half.

What I would suggest is branching on whether `document_url` was explicitly supplied rather than on its inequality with a defaulted value: default `document_url` to `None` and resolve it to `code_sandbox_url` when unset, which is the kind of merging `resolve_url_and_token_variables` already does. `test_same_url_keeps_caller_client` passes `document_url=SANDBOX_URL` explicitly, so the unset case is not covered by the new tests either way.

The routing split itself reads well, and keeping the code sandbox XSRF header off the document session is a good catch.

jupyter_mcp_server/server_context.py
pokutuna: The context is referenced from many files and the rename touches many lines, so I plan to send it as a separate PR.

pokutuna:
@ebarkhordar

Thanks for the detailed report!

Fixed in 315ffd5 as you suggested: `document_url` now defaults to `None`, and unset falls back to the code sandbox server.

The unset state needs to be handled in several places, and the config is updated in place by `set_config` (e.g. `/api/connect`), so I added `config.resolved_document_url()` / `resolved_document_token()` and connections read the fallback from there. Split decisions still compare the raw fields.

The token follows the same rule: with `document_url` unset, the collaboration websocket uses `code_sandbox_token`; an explicitly-set document server never receives the sandbox token. `DocumentCodeSandbox.document_url` is optional now so `jupyter mcp connect` can carry the unset value.

Added `test_unset_document_url_keeps_caller_client` to cover the unset case you pointed out.


@echarles
Added in 18c93f4: how to get the URL from Kaggle's external editor support, and a coding agent example with a local notebook and a Kaggle kernel.

echarles:
Thx @pokutuna for the fast updated. I let @ebarkhordar share further thoughts and will then review.

echarles (approved):
LGTM Thx @pokutuna

echarles:
Merging as needed for https://github.com/datalayer/jupyter-mcp-server/pull/354

PR #352, opened 2026-08-07
fix(execute_code): don't stop a borrowed kernel while its execution task is still in flight

AmirF194:
## Root cause

`_execute_via_notebook_manager`'s `finally` block unconditionally calls `borrowed_sandbox.stop(shutdown_kernel=False)` on a borrowed kernel connection, even when the background execution it just ran is still in flight. On a timeout, `_execute_on_kernel` interrupts the kernel and waits only `TIMEOUT_OUTPUT_SETTLE_SECONDS` (1.0s) via `settle_timed_out_execution` before returning a `TIMEOUT ERROR` result regardless of whether the `asyncio.to_thread` background task has actually finished. The `finally` block then stops the same client object that background thread is still using, closing its transport out from under it.

## Fix

`track_pending_execution` already records the in-flight task on the sandbox client as `_mcp_pending_execution`. The `finally` block now checks it: if the task is still pending, the release is deferred to the task's own `add_done_callback` instead of running synchronously; otherwise it releases immediately as before.

## Verification

- New regression test (`tests/test_execute_code_borrowed_kernel_stop_race.py`) fails on unmodified HEAD (`5d0922a`, borrowed client stopped while its background thread is still asleep) and passes with this fix.
- `tests/test_execute_code_orphaned_timeout.py`, `test_execute_orphaned_timeout.py`, `test_execute_code_timeout.py`, `test_execute_progress_keepalive.py` (18 tests) and the real-kernel `test_execute_code_kernel_id.py` suite (4 tests, against actual Jupyter kernels) all pass.
- `ruff check` and `mypy` on the changed file show the same pre-existing findings before and after this diff; no new issues introduced.
- Did not run the full CI matrix (OS/Python legs); this change only touches the borrowed-kernel path of `execute_code`, not the current-notebook-kernel path.

### DIFF
--- a/jupyter_mcp_server/tools/execute_code_tool.py
+++ b/jupyter_mcp_server/tools/execute_code_tool.py
@@ -152,10 +152,24 @@ class ExecuteCodeTool(BaseTool):
             )
         finally:
             if borrowed_sandbox is not None:
-                try:
-                    borrowed_sandbox.stop(shutdown_kernel=False)
-                except Exception as stop_err:
-                    logger.warning(f"Failed to release kernel {kid}: {stop_err}")
+                pending = getattr(borrowed_sandbox, "_mcp_pending_execution", None)
+                if pending is not None and not pending.done():
+                    # track_pending_execution's background thread is still using this
+                    # client's transport; releasing now would close it out from under
+                    # that thread, so defer until the thread actually finishes.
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @AmirF194

PR #354, opened 2026-08-08
enh: code sandbox

echarles:
Use latest code_sandboxes package that allows to not rely on the kernel_client field anymore and fully on the code sandbox client api.


echarles (commented):
LGTM

PR #358, opened 2026-08-09
fix(execute_code): stop the kernel client's channels when execution raises

### LINKED ISSUE #357: execute_code_local leaves the kernel client's channels open when execution raises
`execute_code_local` (`jupyter_mcp_server/utils.py`) creates its own kernel client and opens its channels:

```python
client = lkm.client()

if not client.channels_running:
    client.start_channels()
```

`stop_channels()` is reached on two paths only: the timeout return, and the normal return at the end of the `try` body. Anything raised in between lands in the generic `except Exception`, which logs and returns `[ERROR: ...]` with the sockets still bound. `lkm.client()` builds a fresh client on every call, so nothing else holds a reference and nothing reclaims it.

A `BEFORE_EXECUTE` hook handler with `propagate_errors=True` is the shortest supported way to get there, since `HookRegistry.fire` re-raises for those handlers and the hook fires immediately after the channels start. A kernel that dies mid-execution reaches the same handler.

### Reproduction

On `74b03fd`, `python:3.11-slim`, against a real kernel:

```python
import asyncio
import os

from jupyter_server.services.kernels.kernelmanager import AsyncMappingKernelManager

from jupyter_mcp_server.hooks import HookEvent, HookRegistry
from jupyter_mcp_server.utils import execute_code_local


class FailingHandler:
    propagate_errors = True

    async def on_event(self, event, **kwargs):
        if event == HookEvent.BEFORE_EXECUTE:
            raise RuntimeError("handler failed")


class ServerApp:
    def __init__(self, km):
        self.kernel_manager = km


async def main():
    km = AsyncMappingKernelManager()
    km.root_dir = "/tmp"
    kernel_id = await km.start_kernel(kernel_name="python3")

    lkm = km.pinned_superclass.get_kernel(km, kernel_id)
    created = []
    build_client = lkm.client

    def capture():
        client = build_client()
        created.append(client)
        return client

    lkm.client = capture
    HookRegistry.get_instance().register(FailingHandler())

    fds_before = len(os.listdir("/proc/self/fd"))
    print("result:", await execute_code_local(ServerApp(km), "nb.ipynb", "1 + 1", kernel_id, timeout=30))
    await asyncio.sleep(0.3)
    print("channels_running:", created[0].channels_running)
    print("fd delta:", len(os.listdir("/proc/self/fd")) - fds_before)

    await km.shutdown_kernel(kernel_id)


asyncio.run(main())
```

```
result: ['[ERROR: handler failed]']
channels_running: True
fd delta: 13
```

Thirteen descriptors per failed call, on a client whose channels are still running. Dropping the handler registration gives `result: ['2']`, `channels_running: False` and `fd delta: 2`, so only the error path is affected.

ebarkhordar:
`execute_code_local` creates its own kernel client and opens its channels:

```python
client = lkm.client()

if not client.channels_running:
    client.start_channels()
```

`stop_channels()` is reached on the timeout return and on the normal return at the end of the `try` body, and nowhere else. An exception raised between those two points lands in the generic `except Exception`, which returns `[ERROR: ...]` with the sockets still bound. Since `lkm.client()` hands back a fresh client on every call, nothing else holds a reference to it and the descriptors are never reclaimed.

The reachable trigger is a `BEFORE_EXECUTE` hook handler with `propagate_errors=True`: `HookRegistry.fire` re-raises for those, and the hook fires immediately after the channels start. A kernel that dies mid-execution reaches the same handler.

A `finally` now stops the channels on every exit. `client` is initialised to `None` before the `try` because that handler also covers failures from before the client exists, an unknown `kernel_id` being the obvious one, and the `channels_running` guard leaves the timeout and success paths at exactly one `stop_channels()` call.

### Verification

- `tests/test_execute_code_local_channel_cleanup.py::test_channels_stopped_when_execution_raises` fails on `main` (`stop_calls == 0`) and passes here.
- Against a real kernel in `python:3.11-slim`, the failing-hook path goes from 13 leaked descriptors with `channels_running == True` to 0 and `False`. An ordinary run returns `['2']` with one `stop_channels()` call and the same descriptor count on both sides. Script and numbers are in #357.
- `pytest tests/` gives the same results on `main` and on this branch: 270 vs 272 passed, the two extra being the new tests, with the same pre-existing `test_span_log_summary` failure and the same 117 live-server errors on both. `ruff check .` output is identical between the two.
- Not checked: the fault path end to end through the `execute_code` tool. The reproduction calls `execute_code_local` directly.

Fixes #357

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -1118,6 +1118,8 @@ async def execute_code_local(
 
         logger = logging.getLogger(__name__)
 
+    client: Any = None
+
     try:
         # Get kernel manager
         kernel_manager = serverapp.kernel_manager
@@ -1308,6 +1310,12 @@ async def execute_code_local(
     except Exception as e:
         logger.error(f"Error executing code locally: {e}")
         return [f"[ERROR: {e!s}]"]
+
+    finally:
+        # lkm.client() hands back a fresh client per call, so channels left
+        # running here are never reclaimed.
+        if client is not None and client.channels_running:
+            client.stop_channels()
 
 
 async def execute_cell_local(
--- a/tests/test_execute_code_local_channel_cleanup.py
+++ b/tests/test_execute_code_local_channel_cleanup.py
@@ -1,0 +1,107 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Regression test for execute_code_local leaking the kernel client's channels.
+
+stop_channels() is reached on the timeout and success paths only, so an
+exception raised after the channels are open falls through to the generic
+handler with the sockets still bound. A BEFORE_EXECUTE hook with
+propagate_errors=True is the shortest supported way to raise there.
+"""
+
+import pytest
+
+from jupyter_mcp_server.hooks import HookEvent, HookRegistry
+from jupyter_mcp_server.utils import execute_code_local
+
+
+class FailingBeforeExecuteHandler:
+    """Raises out of BEFORE_EXECUTE, which fires just after start_channels()."""
+
+    propagate_errors = True
+
+    async def on_event(self, event: HookEvent, **kwargs) -> None:
+        if event == HookEvent.BEFORE_EXECUTE:
+            raise RuntimeError("handler failed")
+
+
+class FakeKernelClient:
+    def __init__(self):
+        self.channels_running = False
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the fast review on this one.

PR #361, opened 2026-08-12
fix(hooks): fire AFTER_EXECUTE on the JUPYTER_SERVER-mode error exits

### LINKED ISSUE #360: BEFORE_EXECUTE fires without a paired AFTER_EXECUTE on the JUPYTER_SERVER-mode error exits
In JUPYTER_SERVER mode, `execute_via_execution_stack` and `execute_code_local` fire `HookEvent.BEFORE_EXECUTE` and then leave through several exits without firing `AFTER_EXECUTE`. A handler that stashes per-execution state in the context it gets back from the BEFORE fire never gets a chance to release it. `OTelHookHandler` ends its span only in the AFTER branch and `create_otel_handler` wires a `SimpleSpanProcessor`, which exports on end, so a timed-out execution is missing from the span file entirely rather than recorded as an error.

The same tool pairs correctly in the other server mode: `execute_code_tool._execute_on_kernel` fires `AFTER_EXECUTE` on all three of its exits (timeout with `error=asyncio.TimeoutError()`, success, and `except Exception`). So `execute_code` emits a complete hook pair in MCP_SERVER mode and an unpaired one in JUPYTER_SERVER mode for identical user-visible behaviour.

Exits that owe an `AFTER_EXECUTE` and do not fire one, at `cf707a1`, enumerated by walking both functions for `Return`/`Raise` nodes past the BEFORE fire:

`jupyter_mcp_server/utils.py` `execute_via_execution_stack` (BEFORE at :954)
- :974 `raise TimeoutError(...)`, the timeout path
- :1032 the `input_request` branch
- :1045 the `json.JSONDecodeError` branch
- :1094 the `except (asyncio.CancelledError, TimeoutError)` re-raise, i.e. an MCP user-cancel
- :1098 the outer `except Exception`

`jupyter_mcp_server/utils.py` `execute_code_local` (BEFORE at :1152)
- :1213 the `remaining_ms <= 0` timeout branch
- :1324 the outer `except Exception`

## Reproduction

No kernel and no server needed. `python:3.13-slim`, `pip install -e .` at `cf707a1`:

```python
import asyncio

from jupyter_mcp_server.hooks import HookEvent, HookRegistry
from jupyter_mcp_server.otel_hook import create_otel_handler
from jupyter_mcp_server.utils import execute_via_execution_stack


class Recorder:
    propagate_errors = False

    def __init__(self):
        self.events = []

    async def on_event(self, event: HookEvent, **kwargs):
        self.events.append(event.name)


class NeverCompletingStack:
    def put(self, kernel_id, code, metadata):
        return "request-1"

    def get(self, kernel_id, request_id):
        return None

    def cancel(self, kernel_id):
        pass


class Extension:
    def __init__(self, stack):
        self._Extension__execution_stack = stack


class ExtensionManager:
    def __init__(self, extension):
        self.extension_apps = {"jupyter_server_nbmodel": {extension}}


class ServerApp:
    def __init__(self, stack):
        self.extension_manager = ExtensionManager(Extension(stack))


async def main():
    recorder = Recorder()
    HookRegistry.get_instance().register(recorder)
    HookRegistry.get_instance().register(create_otel_handler(file_path="/tmp/spans.jsonl"))

    outputs = await execute_via_execution_stack(
        serverapp=ServerApp(NeverCompletingStack()),
        kernel_id="kernel-1",
        code="while True: pass",
        timeout=0,
        poll_interval=0,
    )

    print("returned:", outputs)
    print("hook events:", recorder.events)
    print("otel spans exported:", sum(1 for _ in open("/tmp/spans.jsonl")))


asyncio.run(main())
```

Observed at `cf707a1`:

```
returned: ['[ERROR: Execution timed out after 0 seconds]']
hook events: ['BEFORE_EXECUTE']
otel spans exported: 0
```

Expected: one `AFTER_EXECUTE`, and one exported span carrying `error`.

The `except (asyncio.CancelledError, TimeoutError)` block arrived with #231, which re-indented the polling loop into a new `try` and carried the existing AFTER fires along without adding one inside the new handler, so this looks like an omission rather than a decision. PR with the fix and a test per exit is attached.

ebarkhordar:
`execute_via_execution_stack` and `execute_code_local` fire `BEFORE_EXECUTE` and then return or re-raise on their timeout, cancellation, input-request, unparseable-output and unexpected-error exits without firing `AFTER_EXECUTE`, so a handler that stashes per-execution state in the context never gets it back. `OTelHookHandler` ends its span only in the AFTER branch and `SimpleSpanProcessor` exports on end, so a timed-out execution is absent from the span file rather than recorded as an error. Full exit enumeration and a standalone reproduction are in #360.

## Fix

Fire `AFTER_EXECUTE` on every exit past the BEFORE fire, using the payload shape `execute_code_tool._execute_on_kernel` already uses for the same tool in MCP_SERVER mode: `outputs=` what the caller gets back, `error=` the exception.

Two details worth a second look:

- `hook_ctx` starts as `None`, so the outer `except Exception` can distinguish a failure before the BEFORE fire (a missing `jupyter_server_nbmodel` extension, for instance) from one after it, and stays silent for the first.
- The `except (asyncio.CancelledError, TimeoutError)` handler fires and then re-raises. `TimeoutError` reaches the outer handler and `CancelledError` does not, since it is not an `Exception`, so the fire has to live in the inner handler and `after_execute_fired` keeps the outer one from firing a second time on the timeout path.

## Verification

Clean `python:3.13-slim` container, `pip install -e '.[test]'`, base `cf707a1`, provenance `/w/jupyter_mcp_server/utils.py`.

`tests/test_execute_hook_pairing.py` is new, one test per exit plus two controls:

- at `cf707a1`: **8 failed, 2 passed**. The two that pass are the success path and the pre-BEFORE failure, which owes no AFTER.
- on this branch: **10 passed**.

The standalone reproduction from #360, same container:

```
cf707a1      hook events: ['BEFORE_EXECUTE']                    otel spans exported: 0
this branch  hook events: ['BEFORE_EXECUTE', 'AFTER_EXECUTE']   otel spans exported: 1
```

Regression: `tests/test_otel_integration.py` 7 passed on both transports against a real Jupyter server, and `test_hooks.py` / `test_otel_hooks.py` / `test_execute_via_execution_stack.py` / the `test_execute_*` timeout files 38 passed. The whole suite in that container ends 387 passed / 9 failed / 5 errors, and re-running those five failing files against unmodified `main` in the same container gives the same set (`test_mcpb_version_sync`, `test_tools`, `test_use_notebook_ui_open`, `test_use_notebook_split_servers`, and `test_move_cell`, which varies run to run), so they look environmental rather than caused by this change. CI here is the real check.

`ruff format --check` clean, and `ruff check jupyter_mcp_server/utils.py` reports the same 23 pre-existing errors as `main`.

Fixes #360

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -932,6 +932,12 @@ async def execute_via_execution_stack(
     if logger is None:
         logger = default_logging.getLogger(__name__)
 
+    # hook_ctx is set once BEFORE_EXECUTE has fired; every exit past that point
+    # owes exactly one AFTER_EXECUTE carrying this context back to the handlers.
+    # The cancellation handler re-raises into the outer one, hence the flag.
+    hook_ctx = None
+    after_execute_fired = False
+
     try:
         # Get the ExecutionStack from the jupyter_server_nbmodel extension
         nbmodel_extensions = serverapp.extension_manager.extension_apps.get(
@@ -1029,7 +1035,17 @@ async def execute_via_execution_stack(
                     # Check for pending input (shouldn't happen with allow_stdin=False)
                     if "input_request" in result:
                         logger.warning("Unexpected input request during execution")
-                        return ["[ERROR: Unexpected input request]"]
+                        input_request_output = ["[ERROR: Unexpected input request]"]
+                        await HookRegistry.get_instance().fire(
+                            HookEvent.AFTER_EXECUTE,
+                            code=code,
+                            kernel_id=kernel_id,
+                            metadata=metadata,
+                            outputs=input_request_output,
+                            error=RuntimeError("Unexpected input request during execution"),
+                            context=hook_ctx,
+                        )
+                        return input_request_output
 
                     # Extract outputs
                     outputs = result.get("outputs", [])
@@ -1040,9 +1056,19 @@ async def execute_via_execution_stack(
 
                         try:
                             outputs = json.loads(outputs)
-                        except json.JSONDecodeError:
+                        except json.JSONDecodeError as decode_err:
                             logger.error(f"Failed to parse outputs JSON: {outputs}")
-                            return ["[ERROR: Invalid output format]"]
+                            decode_error_output = ["[ERROR: Invalid output format]"]
+                            await HookRegistry.get_instance().fire(
+                                HookEvent.AFTER_EXECUTE,
+                                code=code,
+                                kernel_id=kernel_id,
+                                metadata=metadata,
+                                outputs=decode_error_output,
+                                error=decode_err,
+                                context=hook_ctx,
+                            )
+                            return decode_error_output
 
                     if outputs:
                         formatted = safe_extract_outputs(outputs)
@@ -1080,7 +1106,7 @@ async def execute_via_execution_stack(
                 # Still pending, wait before next poll
                 await asyncio.sleep(poll_interval)
 
-        except (asyncio.CancelledError, TimeoutError):
+        except (asyncio.CancelledError, TimeoutError) as interrupt_err:
             # Clean up the orphaned execution request to prevent subsequent
             # execute_cell calls from hanging on stale state.
             logger.warning(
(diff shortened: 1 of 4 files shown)


echarles:
Unrelated to these changes, test are failing

```
FAILED tests/test_mcpb_version_sync.py::test_manifest_version_matches_package_version - AssertionError: assert '1.3.0' == '1.3.1'
  
  - 1.3.1
  ?     ^
  + 1.3.0
  ?     ^
FAILED tests/test_mcpb_version_sync.py::test_mcpb_pyproject_version_matches_package_version - AssertionError: assert '1.3.0' == '1.3.1'
  
  - 1.3.1
  ?     ^
  + 1.3.0
  ?     ^
======= 2 failed, 387 passed, 7 skipped, 2 warnings in 400.61s (0:06:40) =======
make: *** [Makefile:33: test-mcp-server] Error 1
```

echarles:
Issue Introduced in https://github.com/datalayer/jupyter-mcp-server/commit/c6ca927154305046b64d47820fbc239f8debbd48 as a direct commit without aligning the other versions.... sorry for that, I let you bake the quick fix in this PR if it works for you.

ebarkhordar:
Baked in as 92516f5. Ran `scripts/sync_mcpb_version.py`, which puts both mcpb files at 1.3.1; the version sync check and its two tests pass locally. No problem on the direct commit.

echarles (approved):
LGTM Thx @ebarkhordar

PR #362, opened 2026-08-13
Document Provider

echarles:
Renames provider to document_provider, so that the name says what the option actually selects.

It only ever chose where the notebook documents live — jupyter for the collaboration API of a Jupyter Server, datalayer for the Datalayer spacer. Where the code runs is a separate choice, made with --sandbox-variant. The old name, and its help text, suggested one option governed both.

Nothing breaks in 1.3.2
The former names are deprecated, not removed. --provider is still accepted as an alias, PROVIDER is still read, and a /connect payload carrying "provider" is still understood. Migrate when convenient.

Who Should Migrate
End users passing --provider or setting PROVIDER, which in practice means anyone pointing the server at Datalayer-hosted documents.
Operators whose manifests, wrapper scripts or MCP client configurations carry PROVIDER.
Anyone calling the /connect endpoint or the connect_to_jupyter tool with a provider field.
Change Summary
Use document_provider names everywhere instead of provider names.
The accepted values are unchanged: jupyter (default) and datalayer.
Execution backends keep their own option, --sandbox-variant / SANDBOX_VARIANT.
Nothing has to change immediately: the old names still work in this version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ working on your data when your laptop does not.
 <table>
 <tr><td>
 
-![Renamed](https://img.shields.io/badge/%F0%9F%94%84-Renamed-E67E22?style=for-the-badge&labelColor=A0522D)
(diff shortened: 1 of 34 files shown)


echarles (commented):
LGTM

PR #365, opened 2026-08-13
fix: start local kernels in notebook directory

### LINKED ISSUE #356: use_notebook starts kernels in the server root, not the notebook directory (path dropped in _start_kernel_local)
### Summary

Kernels started by `use_notebook` (JUPYTER_MCP local/JUPYTER_SERVER mode) start with their working directory set to the Jupyter server root, not the notebook's directory. Every other Jupyter client — JupyterLab's sessions API, `nbconvert --execute`, VS Code — launches the kernel in the notebook's folder, so notebooks written against this MCP server break the universal relative-path idiom (`pd.read_csv("data.csv")` for a file sitting next to the notebook) and produce outputs in the wrong directory.

### Reproduction (MCP tools only)

1. `use_notebook(notebook_name="t", notebook_path="projects/demo/t.ipynb", mode="create")`
2. `execute_code(code="import os; print(os.getcwd())")`

Observed: the server root (e.g. `/home/user`).
Expected: `projects/demo` resolved against the server root — which is what a kernel for this notebook gets from JupyterLab itself.

Reproduced on 1.1.4 and the code is unchanged on current `main`.

### Diagnosis

The path makes it all the way to the kernel-start seam and is dropped in the last step. In `jupyter_mcp_server/tools/use_notebook_tool.py`:

- The call site passes it: `kernel = await self._start_kernel_local(kernel_manager, path=notebook_path)`
- `_start_kernel_local(self, kernel_manager, path=None)` accepts it, then calls `kernel_id = await kernel_manager.start_kernel()` — without the path.

`jupyter_server`'s kernel manager already supports this: `start_kernel(path=...)` resolves the kernel's cwd from the API path (that is exactly how the sessions API gives JupyterLab kernels the notebook's directory). So the fix appears to be one line:

```python
kernel_id = await kernel_manager.start_kernel(path=path)
```

(The subsequent `session_manager.create_session(path=notebook_path, kernel_id=kernel_id, ...)` associates the already-started kernel with the session, so it does not retroactively fix the cwd.)

### Impact

Beyond broken relative reads, agents driving this server learn to compensate with server-root-relative or absolute paths, which poisons the portability of the notebooks they produce — the notebook only works from the MCP server's root, not in JupyterLab, nbconvert, or a colleague's checkout.

Happy to open a PR with the one-liner if that's welcome.

Tethys0:
## Summary
- forward the selected notebook path to the local kernel manager
- add a regression test for creating a nested notebook in JUPYTER_SERVER mode

Fixes #356

## Validation
- `python3 -m pytest tests/test_use_notebook_create_local.py::test_create_starts_the_kernel_in_the_notebook_directory -q`
- `python3 -m pytest tests/test_use_notebook_create_local.py -q`
- `ruff check jupyter_mcp_server/tools/use_notebook_tool.py tests/test_use_notebook_create_local.py` (the file has pre-existing violations; no changed-line violations)
- `make test` was exercised after installing the optional sandbox extension: 473 passed; 3 failures and 4 errors in pre-existing Jupyter extension/UI tests unrelated to this patch.

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -84,10 +84,10 @@ class ExecuteCellTool(BaseTool):
             return True
 
     async def _start_and_bind_kernel(
-        self, kernel_manager, notebook_manager, notebook_path: str
+        self, kernel_manager, notebook_manager, notebook_path: str | None
     ) -> str:
         """Start a kernel and rebind it to the current notebook in local mode."""
-        kernel_id = await kernel_manager.start_kernel()
+        kernel_id = await kernel_manager.start_kernel(path=notebook_path)
         await asyncio.sleep(1.0)
         logger.info(f"Kernel {kernel_id} started and initialized")
 
@@ -257,8 +257,10 @@ class ExecuteCellTool(BaseTool):
             if kernel_manager is None:
                 raise ValueError("kernel_manager is required for JUPYTER_SERVER mode")
 
-            # Get notebook_path and kernel_id first
+            # Preserve the root-relative API path for kernel creation, then resolve
+            # a separate filesystem path for file and YDoc operations below.
             notebook_path, kernel_id = get_current_notebook_context(notebook_manager)
+            notebook_api_path = notebook_path
 
             # Resolve to absolute path
             if notebook_path and serverapp and not Path(notebook_path).is_absolute():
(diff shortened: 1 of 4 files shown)


ebarkhordar (commented):
Confirmed against a real `AsyncMappingKernelManager` (jupyter_server 2.20.0 + ipykernel, kernel cwd read from the kernel process's `/proc/<pid>/cwd` rather than from the stub): for `projects/demo/notebook.ipynb`, `main` starts the kernel in the server root and this branch starts it in `projects/demo`. `RecordingKernelManager` can only pin the argument, so that end-to-end result is worth a line in the PR body.

One call site still has the old shape: `execute_cell_tool.py:90`, inside `_start_and_bind_kernel`, calls `kernel_manager.start_kernel()` with `notebook_path` already a parameter. Its callers are the "no kernel yet" and "kernel disappeared, retry once" paths, so a session that starts in the notebook directory drops back to the server root the first time its kernel is culled and replaced. Measured the same way, it is the server root on both `main` and this branch. That line predates #254, so it was missed by that pass too.

(`execute_code_tool.py:340` is the other bare call, but it has no notebook in scope, so the server root looks right there.)

echarles:
Test are constantly failing on windows (relaunched 3 times).

echarles (approved):
LGTM Thx @Tethys0

PR #372, opened 2026-08-14
enh: use_notebook no longer creates a kernel

echarles:
use_notebook no longer creates a kernel. It sets kernel = None and reports "Notebook opened. A kernel starts on the first execution."

This works for every provider and every sandbox variant, because the lazy path was already correct and nobody was using it. ensure_kernel_alive calls create_kernel in utils.py:589, which asks extensions first — so sandbox_variant=datalayer gets a DatalayerSandbox, and only an unclaimed variant falls through to the Jupyter client. The old code bypassed all of that by calling create_jupyter_sandbox_client directly, which is why your Datalayer worker sat for two minutes polling an /api/status that will never answer.

Opening is now free: reading cells and outputs needs no runtime, so you don't spend compute to answer "what does this notebook do?"

### DIFF
--- a/.github/workflows/enforce-label.yml
+++ b/.github/workflows/enforce-label.yml
@@ -1,0 +1,13 @@
+name: Enforce PR label
+
+on:
+  pull_request:
+    types: [labeled, unlabeled, opened, edited, synchronize]
+jobs:
+  enforce-label:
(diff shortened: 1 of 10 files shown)


echarles (commented):
LGTM

PR #373, opened 2026-08-14
fix(execute_code): don't inherit the kernel's ZMQ identity

Installka:
## The problem

In `JUPYTER_SERVER` mode, one `execute_code` call permanently breaks `execute_cell` for that kernel.

`execute_code` builds its own client with `lkm.client()`. `KernelManager.client()` hands back a *clone* of the manager's `Session`, and `clone()` keeps the session id — which jupyter_client passes straight through as the ZMQ socket identity:

```python
socket = self.connect_shell(identity=self.session.bsession)   # jupyter_client/client.py:359
```

ipykernel sets `ROUTER_HANDOVER` (#300 upstream, working around zeromq/libzmq#2892 so a client can reconnect while the kernel is busy). With handover, a second peer presenting an identity that is already in use **takes it over**, and the first client keeps its sockets but stops reaching the kernel.

`jupyter-server-nbmodel` keeps one long-lived client per kernel to serve `execute_cell`, so `execute_code` orphans it. What the user sees:

- `execute_cell` hangs until its timeout and returns `[ERROR: Execution timed out after N seconds]`, and the file-mode path writes that string into the notebook cell as if it were the cell's output, with an `execution_count` derived by scanning the file.
- `POST /api/kernels/<id>/execute` returns `202 Accepted` / `"queued"` and never completes; `/api/kernels` shows the kernel `idle` with frozen `last_activity` — it never receives the code.
- Restarting the kernel with the same id does **not** help. Only a new kernel id or a server restart does.
- `execute_code` keeps working throughout, which makes this look like an `execute_cell` bug.

Two individually reasonable upstream decisions collide here: jupyter/jupyter_client#252 added `Session.clone()` *specifically* so that "multiple Clients … instantiated against the same kernel" don't share `digest_history`, but it keeps the session id; ipython/ipykernel#300 made same-identity reconnects take over. Multi-client support plus same-identity handover means the newest client silently evicts the others.

## The fix

`create_isolated_kernel_client()` in `utils.py` gives each short-lived client its own session id before its channels are started. The signing key is shared, so the kernel still accepts the messages; clients that expose no `Session` (remote/websocket-backed) pass through unchanged.

Applied at both sites that build a client from a kernel:

- `utils.py` — `execute_code_local()`
- `jupyter_extension/backends/local_backend.py` — `execute_cell()`

## Tests

`tests/test_kernel_client_zmq_identity.py`:

- two unit tests — distinct session ids per client; a client with no `Session` is returned unchanged;
- one live test against a real kernel: a long-lived client executes, a short-lived client connects and executes, and the long-lived client must still be able to execute.

The live test earns its keep — swapping `create_isolated_kernel_client(manager)` back to `manager.client()` fails it:

```
E       AssertionError: assert 'TIMEOUT' == 'ok'
E         - ok
E         + TIMEOUT
```

With the fix: `3 passed in 4.59s`.

Also run, all passing: `test_execute_code_local_channel_cleanup.py`, `test_execute_hook_pairing.py`, `test_local_backend_contents_manager.py`, `test_execute_code_borrowed_kernel_stop_race.py`, `test_execute_code_orphaned_timeout.py` — 26 passed.

`test_execute_code_kernel_id.py` has fixture errors ("kernel … was not ready within 60s"), but they reproduce on an unmodified checkout of the same commit and vary run to run, so they look like pre-existing flakiness in that fixture rather than fallout from this change.

## Notes

Verified on `jupyter-mcp-server` 1.3.8 as a server extension, with `jupyter-server-nbmodel` 0.2.4, `jupyter-client` 8.9.1, `jupyter-collaboration` 5.0.0, Python 3.13, macOS.

### DIFF
--- a/jupyter_mcp_server/jupyter_extension/backends/local_backend.py
+++ b/jupyter_mcp_server/jupyter_extension/backends/local_backend.py
@@ -16,7 +16,7 @@ from jupyter_core.utils import ensure_async
 from mcp.types import ImageContent
 
 from jupyter_mcp_server.jupyter_extension.backends.base import Backend
-from jupyter_mcp_server.utils import safe_extract_outputs
+from jupyter_mcp_server.utils import create_isolated_kernel_client, safe_extract_outputs
 
 if TYPE_CHECKING:
     from jupyter_server.serverapp import ServerApp
@@ -330,9 +330,10 @@ class LocalBackend(Backend):
         cell = cells[cell_index]
         source = "".join(cell["source"]) if isinstance(cell["source"], list) else cell["source"]
 
-        # Get kernel client
+        # Get kernel client. Its ZMQ identity must not collide with any
+        # long-lived client on the same kernel (see create_isolated_kernel_client).
         kernel = self.kernel_manager.get_kernel(kernel_id)
-        client = kernel.client()
+        client = create_isolated_kernel_client(kernel)
 
         # Execute code
         msg_id = client.execute(source)
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import time
+import uuid
 from collections.abc import Callable
 from typing import Any, cast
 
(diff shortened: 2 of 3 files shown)


echarles:
> POST /api/kernels/<id>/execute returns 202 Accepted / "queued" and never completes; /api/kernels shows the kernel idle with frozen last_activity — it never receives the code.

Is that a new issue you have seen with the latest jupyter-server-nbmodel? (which version are you using?)

Installka:
> Is that a new issue you have seen with the latest jupyter-server-nbmodel? (which version are you using?)

Yes. I'm using jupyter-server-nbmodel 0.2.4

echarles (approved):
LGTM Thx @Installka

PR #379, opened 2026-08-16
docs: restore star history chart

FaintFlower:
The star history chart in the README footer was removed in f9caa8c (#372) because it was broken. The GitHub stargazer API restrictions broke the old chart, so a working fix is needed to bring it back.

This PR restores the chart in its original location, pointing to star-history.dera.page which provides the same star history data reliably.


echarles (approved):
LGTM Thx @FaintFlower

PR #381, opened 2026-08-16
fix: do not query jupyter-mcp-tools for an empty allowlist

### LINKED ISSUE #380: An empty allowed_jupyter_mcp_tools publishes every JupyterLab command as an MCP tool
## Problem

Setting `allowed_jupyter_mcp_tools` to an empty string publishes **every** registered JupyterLab command as an MCP tool, which is the opposite of what the setting reads as.

`Config.get_allowed_jupyter_mcp_tools()` returns `[]` for an empty value, and both consumers then build the query as `",".join([])`, which is `""`:

- `jupyter_mcp_server/server.py:1282` (`get_registered_tools`)
- `jupyter_mcp_server/jupyter_extension/handlers.py:183` (`MCPSSEHandler.post`, `tools/list`)

jupyter-mcp-tools filters only when the query is non-empty. Its request handler guards the filter with `if query:`, so an empty query skips filtering entirely and the response carries every registered command. Neither call site narrows the result afterwards, and in `handlers.py` the full set is then cached as `MCPSSEHandler._jupyter_tool_names`, the routing table used for `tools/call`.

The empty value reaches the config through the extension: `jupyter_extension/extension.py:155` assigns `config.allowed_jupyter_mcp_tools` directly, so it bypasses `set_config()`, which would otherwise drop `""` as an alias for `None`.

## Reproduction

Verified in a clean `python:3.12-slim` container against `6bb9e5a`, with a stand-in for `jupyter_mcp_tools.get_tools` that reproduces the `if query:` behaviour read from the installed 0.1.7 wheel, so no JupyterLab frontend is needed.

With `jupyterlab=True` and `config.allowed_jupyter_mcp_tools = ""`, `get_registered_tools()` registers all five commands offered by the stand-in, including `filebrowser_delete` and `terminal_open`. With `allowed_jupyter_mcp_tools = "notebook_run-all-cells"` it registers only that one.

## Expected

An empty allowlist enables no JupyterLab command, so none should be registered and jupyter-mcp-tools should not be queried for a filter it would ignore.

## Environment

- jupyter-mcp-server at `6bb9e5a` (main)
- jupyter-mcp-tools 0.1.7
- Python 3.12

ebarkhordar:
Fixes #380.

## Root cause

`Config.get_allowed_jupyter_mcp_tools()` returns `[]` for an empty `allowed_jupyter_mcp_tools`, and both consumers build their query as `",".join([])`, which is `""`. jupyter-mcp-tools guards its filter with `if query:`, so an empty query means no filter and the response carries every registered JupyterLab command. Neither call site narrows the result afterwards, and `MCPSSEHandler` caches the whole set as `_jupyter_tool_names`, the routing table used for `tools/call`.

The empty value reaches the config through `jupyter_extension/extension.py:155`, which assigns the attribute directly and so bypasses the handling of `""` in `set_config()`.

## Fix

Treat an empty allowlist the way JupyterLab mode being off is already treated, and skip the query. The existing `else` branch at each call site already produces no tools and an empty routing table, so this is one condition in each of `server.py` and `handlers.py`. A non-empty allowlist is unaffected.

## Verification

- `tests/test_allowed_jupyter_mcp_tools.py` is new: the empty and whitespace-only cases fail on `6bb9e5a` (all five commands from the stand-in register, including `filebrowser_delete` and `terminal_open`) and pass here, and the configured-allowlist case passes on both.
- Both suites in a clean `python:3.12-slim` container, this branch against a pristine worktree of main: `TEST_JUPYTER_SERVER=true` gives 436 passed against 434, `TEST_MCP_SERVER=true` gives 436 against 434, the difference being the two tests above. The one remaining failure and the error counts are identical on both sides, from integration fixtures needing a JupyterLab that container could not start.
- `ruff check` reports the same 56 findings on the two changed files before and after, and the new test file is clean.
- Not verified: no live JupyterLab frontend was involved, so the stand-in reproduces the filtering I read in the installed jupyter-mcp-tools 0.1.7 rather than exercising it. The new tests drive the `server.py` path, and the identical guard in `handlers.py` has no unit test.

One thing I left alone: with a non-empty allowlist, jupyter-mcp-tools matches each term as a substring of a tool's id or label, so the published set can still be wider than the allowlist. That is documented behaviour on its side, so narrowing it locally reads as your design call rather than a bug fix. I can add an exact-id filter at both call sites if you want one.

### DIFF
--- a/jupyter_mcp_server/jupyter_extension/handlers.py
+++ b/jupyter_mcp_server/jupyter_extension/handlers.py
@@ -156,23 +156,23 @@ class MCPSSEHandler(JupyterHandler):
 
                         logger.info(f"Querying jupyter_mcp_tools at {base_url}")
 
-                        # Check if JupyterLab mode is enabled before loading jupyter-mcp-tools
+                        # Check if JupyterLab mode is enabled before loading jupyter-mcp-tools.
+                        # An empty allowlist enables no command, and jupyter-mcp-tools applies
+                        # no filter to an empty query, so asking would return every command.
+                        from jupyter_mcp_server.config import get_config
+
                         context = ServerContext.get_instance()
                         jupyterlab_enabled = context.is_jupyterlab_mode()
+                        allowed_jupyter_mcp_tools = get_config().get_allowed_jupyter_mcp_tools()
                         logger.info(f"JupyterLab mode check: enabled={jupyterlab_enabled}")
 
-                        if jupyterlab_enabled:
+                        if jupyterlab_enabled and allowed_jupyter_mcp_tools:
                             # Define specific tools we want to load from jupyter-mcp-tools
                             # (https://github.com/datalayer/jupyter-mcp-tools)
                             # jupyter-mcp-tools exposes JupyterLab commands as MCP tools.
                             # Only tools listed here will be available to MCP clients.
                             # To add new tools, also update the list in server.py and
                             # see docs/docs/reference/tools-jupyterlab/index.mdx for documentation.
-                            from jupyter_mcp_server.config import get_config
-
-                            config = get_config()
-                            allowed_jupyter_mcp_tools = config.get_allowed_jupyter_mcp_tools()
-
                             logger.info(
                                 f"Looking for specific jupyter-mcp-tools: {allowed_jupyter_mcp_tools}"
                             )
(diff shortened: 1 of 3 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

PR #383, opened 2026-08-16
fix: start MCP_SERVER-mode kernels in the notebook's directory

### LINKED ISSUE #382: MCP_SERVER-mode kernels start in the server root, not the notebook's directory
Opening a notebook over MCP and running code in it resolves relative paths against the Jupyter server root rather than the directory the notebook lives in. `open("data.csv")`, a relative `%run` and `Path.cwd()` all point at the wrong place, and nothing errors.

Jupyter Server sets a kernel's working directory from the `path` field of `POST /api/kernels`. In MCP_SERVER mode nothing supplies one: `create_code_sandbox` in `jupyter_mcp_server/utils.py` calls `create_jupyter_sandbox_client` without `path`, so `kernel_path` arrives as `None` and the kernel starts in the root directory.

Both MCP_SERVER kernel paths are affected: the eager one in `use_notebook`, taken whenever `--start-new-code-sandbox` is on (the CLI default), and the deferred one that `ensure_code_sandbox_alive` takes on the first execution and on every culled-kernel replacement.

JUPYTER_SERVER mode is not affected. #365 fixed the same class there on 2026-08-14.

### Reproduction

Measured in a clean container at `94ad96b`, Python 3.12, against a real Jupyter Server started with `--ServerApp.root_dir=/srv/nb`, with a notebook at `projects/demo/notebook.ipynb`.

```python
await UseNotebookTool().execute(
    mode=ServerMode.MCP_SERVER,
    sandbox_server_client=JupyterServerClient(base_url=SERVER, token=TOKEN),
    notebook_manager=notebook_manager,
    notebook_name="demo",
    notebook_path="projects/demo/notebook.ipynb",
    use_mode="connect",
    code_sandbox_url=SERVER,
    code_sandbox_token=TOKEN,
)
sandbox = notebook_manager.get_code_sandbox("demo")
print(sandbox.execute("import os; print(os.getcwd())"))
```

Observed:

```
{'execution_count': 1, 'outputs': [{'output_type': 'stream', 'name': 'stdout', 'text': '/srv/nb\n'}], 'status': 'ok'}
```

Expected `/srv/nb/projects/demo`.

ebarkhordar:
Fixes #382

Jupyter Server sets a kernel's working directory from the `path` field of `POST /api/kernels`. `create_code_sandbox` never sends one, so in MCP_SERVER mode every kernel starts in the server root and relative file access inside a notebook silently resolves against the wrong directory.

`use_notebook` used to send it. f9caa8c (#372) replaced its direct `create_jupyter_sandbox_client(..., path=notebook_path)` call with the shared factory so that `--sandbox-variant` is honoured, and the path went with it. The deferred path never sent one at all.

## Fix

`create_code_sandbox` takes an optional `path` and forwards it, which reaches `CodeSandboxClient.create(kernel_path=...)` and then `start(path=...)`. Two callers supply it:

- `use_notebook` passes `notebook_path`. This is the eager path, taken whenever `--start-new-code-sandbox` is on, which is the CLI default.
- `__ensure_code_sandbox_alive` passes the current notebook's path. This is the deferred path, taken on the first execution and on every culled-kernel replacement.

The extension hand-off (`get_extension_manager().create_code_sandbox(config, logger)`) is left alone: a non-Jupyter sandbox variant need not have a notion of a server-relative path, and widening that contract belongs in its own change. `start_code_sandbox` is unchanged too, since it only runs at startup when there is no `document_id` or when auto-enrollment has already failed.

## Verification

- `tests/test_code_sandbox_notebook_cwd.py` covers the factory and both callers. All three fail on `main` and pass here.
- End to end in a clean container against a real Jupyter Server with `root_dir=/srv/nb` and a notebook at `projects/demo/notebook.ipynb`: `os.getcwd()` in the kernel reads `/srv/nb` on `main` and `/srv/nb/projects/demo` on this branch.
- Full `tests/` run in that container: identical results on `main` and on the branch apart from the three new passes (the container has no live server, so the same suites error on both).
- Not checked: any non-Jupyter sandbox variant, since the extension path is untouched.

#292 also edits `create_code_sandbox`, for kernel spec names rather than the working directory, so the two changes will need a rebase against each other.

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -239,7 +239,9 @@ def __ensure_code_sandbox_alive() -> CodeSandboxClient:
     def __create_code_sandbox() -> CodeSandboxClient:
         """Create a new kernel instance using current configuration."""
         config = get_config()
-        return create_code_sandbox(config, logger)
+        return create_code_sandbox(
+            config, logger, path=notebook_manager.get_current_notebook_path()
+        )
 
     current_notebook = notebook_manager.get_current_notebook() or "default"
     return ensure_code_sandbox_alive(notebook_manager, current_notebook, __create_code_sandbox)
--- a/jupyter_mcp_server/tools/use_notebook_tool.py
+++ b/jupyter_mcp_server/tools/use_notebook_tool.py
@@ -367,7 +367,7 @@ class UseNotebookTool(BaseTool):
                     # extensions first and so honours `--sandbox-variant`.
                     from jupyter_mcp_server.utils import create_code_sandbox
 
-                    kernel = create_code_sandbox(config, logger)
+                    kernel = create_code_sandbox(config, logger, path=notebook_path)
                     info_list.append(f"[INFO] Connected to kernel '{kernel.id}'.")
                 else:
                     kernel = None
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -587,7 +587,7 @@ def format_TSV(headers: list[str], rows: list[list[str]]) -> str:
 ###############################################################################
 
 
-def create_code_sandbox(config, logger) -> CodeSandboxClient:
+def create_code_sandbox(config, logger, path: str | None = None) -> CodeSandboxClient:
     """Create a new code sandbox using current configuration.
 
     Creation is resolved in this order:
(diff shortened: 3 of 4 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick merge. Kernels for notebooks outside the server root now land in the right directory again.

PR #388, opened 2026-08-19
fix: read serverapp from the extension context in get_registered_tools

### LINKED ISSUE #387: get_registered_tools() reads serverapp off the wrong ServerContext, so GET /mcp/tools/list has never exposed JupyterLab commands
`GET /mcp/tools/list` returns the FastMCP tools only. The allowlisted JupyterLab commands are never included, under any configuration, because the jupyter-mcp-tools lookup raises before the fetch is attempted.

`jupyter_mcp_server/server.py:1261` reads `serverapp` off the module-level `server_context`:

```python
if server_context.serverapp is not None:
    base_url = server_context.serverapp.connection_url
    token = server_context.serverapp.token
```

`server_context` is bound at `server.py:216` to `jupyter_mcp_server.server_context.ServerContext`, and that class defines no `serverapp`. The attribute belongs to the other `ServerContext`, at `jupyter_mcp_server/jupyter_extension/context.py:55`. The two classes share a name, which is why the block reads correctly.

The resulting `AttributeError` is caught by the `except Exception` at `server.py:1348`, which logs `Error querying jupyter-mcp-tools extension` and falls through to the FastMCP registry. Apart from that log line the failure is silent.

`handlers.py:145` performs the same lookup correctly, through `get_server_context()`.

## Reproduction

Clean `python:3.12-slim` container, editable install of the checkout at `280d1c8`, driving `get_registered_tools()` in `JUPYTER_SERVER` mode with `jupyterlab=True` and a one-entry allowlist, against a stand-in `jupyter_mcp_tools` that records the queries it receives:

```
PROVENANCE jupyter_mcp_server.__file__ = /work/jupyter_mcp_server/__init__.py
CONTEXT CLASS = jupyter_mcp_server.server_context.ServerContext
hasattr(server_context, 'serverapp') = False

LOG INFO  JupyterLab mode enabled, loading selected jupyter-mcp-tools
LOG ERROR Error querying jupyter-mcp-tools extension: 'ServerContext' object has no attribute 'serverapp'
Traceback (most recent call last):
  File "/work/jupyter_mcp_server/server.py", line 1261, in get_registered_tools
    if server_context.serverapp is not None:
       ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ServerContext' object has no attribute 'serverapp'

RESULT queries sent to jupyter-mcp-tools = []
RESULT JupyterLab tools returned = []
```

Expected: the allowlisted command is queried for and registered. Actual: the fetch is never attempted.

The stand-in only stands in for a running JupyterLab. The `AttributeError` fires before any fetch, so it does not depend on that substitution.

## Scope

`MCPSSEHandler.post` does its own inline fetch with the extension context, so `tools/list` over the MCP protocol is unaffected. The plain HTTP route `GET /mcp/tools/list` (`MCPToolsListHandler`, `handlers.py:596`) is the affected caller.

Nothing assigns `serverapp` onto `jupyter_mcp_server.server_context.ServerContext`: an AST parse for `*.serverapp` / `*._serverapp` assignments across the package returns four writers, all on other types (`LocalBackend.serverapp`, the extension app's `extension.serverapp`, and `_serverapp` on the `jupyter_extension.context.ServerContext` in its own `__init__`/`update()`/`reset()`). The string does not occur in `server_context.py` at all, so the branch has raised since it was written.

Introduced by e61f365 (#109), which added the integration. `git log -G 'server_context\.serverapp' --follow` returns that commit only, so there is no prior fix or revert to reconcile with.

I have a fix ready and will open it as a PR against this issue.

ebarkhordar:
Fixes #387.

`get_registered_tools()` read `serverapp` off the module-level `server_context`, which is bound at `server.py:216` to `jupyter_mcp_server.server_context.ServerContext`. That class defines no `serverapp`; the attribute belongs to the same-named class in `jupyter_extension/context.py`. The `AttributeError` was swallowed by the `except Exception` at `server.py:1348`, so the jupyter-mcp-tools fetch was never attempted and `GET /mcp/tools/list` returned the FastMCP tools only.

The fix reads the connection details from the extension context, which is what `handlers.py:145` already does for the same lookup.

## Verification

Clean `python:3.12-slim` container, editable install, `jupyter_mcp_server.__file__` = `/work/jupyter_mcp_server/__init__.py`, at `280d1c8`.

`tests/test_allowed_jupyter_mcp_tools.py`, which #381 added, injected the missing attribute in its fixture:

```python
monkeypatch.setattr(server_context, "serverapp", None, raising=False)
```

`raising=False` was needed because the attribute is absent, and setting it took the fallback branch, so the suite passed over the bug. That line is removed here and the real extension context reports the serverapp it actually holds.

With the fixture corrected and the source fix reverted, 2 of the 4 tests fail:

```
FAILED test_configured_allowlist_still_registers_its_tools
FAILED test_serverapp_is_read_from_the_extension_context
2 failed, 2 passed
```

With the fix applied, `4 passed`.

Wider suite, same container, `--ignore=tests/test_integration.py --ignore=tests/test_execute_cell_sandbox_routing.py` (the latter fails to collect on `jupyter_mcp_sandboxes`, which is not installed):

| | result |
| --- | --- |
| `280d1c8` unmodified | 2 failed, 369 passed, 1 skipped, 121 errors |
| this branch | 2 failed, 370 passed, 1 skipped, 121 errors |

The failures and errors are identical in both runs and are unrelated to this change; the difference is the one added test.

## Limitation

The added test exercises the branch where no ServerApp is present, so it covers the fallback to the configured sandbox URL and proves the fetch is now reached. It does not exercise `connection_url` / `token` against a live `ServerApp`; that path is unchanged apart from the object it is read from.

Happy to adjust the approach if you would rather the two `ServerContext` classes be reconciled instead.

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -27,6 +27,7 @@ from jupyter_mcp_server.config import get_config, set_config
 from jupyter_mcp_server.enroll import auto_enroll_document
 from jupyter_mcp_server.extensions import get_extension_manager
 from jupyter_mcp_server.hooks import HookEvent, HookRegistry, with_hooks
+from jupyter_mcp_server.jupyter_extension.context import get_server_context
 from jupyter_mcp_server.log import logger
 from jupyter_mcp_server.models import DocumentCodeSandbox
 from jupyter_mcp_server.notebook_manager import NotebookManager
@@ -1256,12 +1257,13 @@ async def get_registered_tools():
 
                 from jupyter_mcp_server.tool_cache import get_tool_cache
 
-                # Get the base_url and token from server context
-                # In JUPYTER_SERVER mode, we should use the actual serverapp URL, not hardcoded localhost
-                if server_context.serverapp is not None:
+                # Get the base_url and token from the extension context, which is the object
+                # that carries the ServerApp (handlers.py reads it the same way).
+                extension_context = get_server_context()
+                if extension_context.serverapp is not None:
                     # Use the actual Jupyter server connection URL
-                    base_url = server_context.serverapp.connection_url
-                    token = server_context.serverapp.token
+                    base_url = extension_context.serverapp.connection_url
+                    token = extension_context.serverapp.token
                     logger.info(f"Using Jupyter ServerApp connection URL: {base_url}")
                 else:
                     # Fallback to configuration (for remote scenarios)
--- a/tests/test_allowed_jupyter_mcp_tools.py
+++ b/tests/test_allowed_jupyter_mcp_tools.py
@@ -66,9 +66,9 @@ def registered_tools(monkeypatch):
     monkeypatch.setattr(tool_cache, "_global_tool_cache", None, raising=False)
     monkeypatch.setattr(server_context.__class__, "_mode", ServerMode.JUPYTER_SERVER)
     monkeypatch.setattr(server_context.__class__, "_initialized", True)
-    # get_registered_tools reads server_context.serverapp, which this context object does
-    # not define; None selects its documented fallback to the configured sandbox URL.
-    monkeypatch.setattr(server_context, "serverapp", None, raising=False)
+    # No `serverapp` is injected onto server_context here. The real extension context is
+    # left to report the serverapp it actually holds (none, outside a running server),
+    # which selects the documented fallback to the configured sandbox URL.
 
     async def run(allowed):
         reset_config()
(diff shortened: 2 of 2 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks Eric. The fixture line I added in #381, `monkeypatch.setattr(server_context, "serverapp", None, raising=False)`, is what hid this one: `raising=False` was needed precisely because the attribute never existed, so the suite went green over a branch that raised on every call.

PR #393, opened 2026-08-21
fix(tool_cache): key in-flight fetches by fetcher and shield the shared task

### LINKED ISSUE #392: ToolCache in-flight coalescing ignores the joining caller's fetch_func, and one caller's cancellation kills the shared fetch
`ToolCache._inflight` (`tool_cache.py:50`) is keyed by `_make_cache_key(base_url, query, enabled_only)`, and `get_tools` joins whatever task already exists for that key regardless of the `fetch_func` the joining caller passed (`tool_cache.py:145-152`). The two call sites pass different fetchers deliberately. `handlers.py` wraps `get_tools` with `wait_timeout=5` and the comment "Shorter timeout - if frontend isn't loaded, don't wait long" (added in #171); `server.py:1298` passes the bare `get_tools`, whose default is 30. In JUPYTER_SERVER mode both compute `base_url` and `token` from the same `serverapp`, both build `query` from `get_allowed_jupyter_mcp_tools()`, and both pass `enabled_only=False`, so the keys are identical and whichever fetch started first decides the timeout for both.

Those are two separate routes on the same server, `POST /mcp/` handling `tools/list` and `GET /mcp/tools/list` going through `get_registered_tools()`, so they interleave under ordinary concurrent use. The direction that matters is the SSE handler joining server.py's fetch: the 5 second bound it asked for becomes 30, on exactly the frontend-not-loaded path that bound exists for.

There is a second problem in the same block. `get_tools` awaits the shared task with a bare `await task` (`tool_cache.py:155`). Cancelling a coroutine that is awaiting a task cancels that task too, so one client disconnecting aborts the fetch for every caller joined on the same key. The recovery below is `except Exception`, and `CancelledError` has been a `BaseException` since 3.8, so those callers never reach the stale-serving fallback the file exists to provide.

Reproduced on `ca89f0167f9eb8bdc40b41054750b8894324d402` in a clean `python:3.12-slim` container, with `tool_cache` loaded from site-packages. Two callers on one key, the second joining the first's in-flight fetch:

```
DEFECT 1: in-flight key has no fetcher component
  fetchers actually invoked : ['server.py fetcher (wait_timeout=30)']
  joiner waited             : 0.30s (its own fetcher takes 0.05s)
  RESULT: the joiner's fetch_func was IGNORED

DEFECT 2: `await task` transfers the joiner's cancellation into the shared task
  A (cancelled)         : CancelledError
  B (never cancelled)   : CancelledError
  fetch invocations         : 1
  cache entries after       : 0
  stale-serving fallback    : not reached (CancelledError is a BaseException, `except Exception` misses it)
```

Both arrived with the coalescing in #371 and neither is a criticism of it: sharing the fetch is the right call, the sharing just needs to be per fetcher and it needs to survive one participant going away. I have a fix and tests for both and will open a PR against this.

ebarkhordar:
Fixes #392.

## Root cause

`_inflight` was keyed by `_make_cache_key(base_url, query, enabled_only)`, the same key as `_cache`, so a caller joining an existing fetch ran the creator's `fetch_func` rather than its own. The two call sites pass different fetchers on purpose: `handlers.py` wraps `get_tools` with `wait_timeout=5` ("if frontend isn't loaded, don't wait long", #171) and `server.py` passes the bare `get_tools` at the library default of 30. In JUPYTER_SERVER mode every other key component is computed identically in both, so `POST /mcp/` and `GET /mcp/tools/list` collide on one key and the first fetch to start decides the timeout for both.

The shared task was also awaited bare. Cancelling a coroutine that awaits a task cancels that task, so one client disconnecting aborted the fetch for every caller joined on it, and because the recovery below is `except Exception` and `CancelledError` is a `BaseException`, those callers did not reach the stale-serving fallback either.

## Fix

The in-flight key becomes `(cache_key, fetch_func)`, so callers coalesce only when they asked for the same thing through the same fetcher. `handlers.py`'s wrapper moves to module level: it was rebuilt per request, and a fresh function object per request would have made that key unique every time and stopped concurrent `tools/list` requests from sharing a call at all. Its `wait_timeout=5` and its reason are unchanged.

The bare `await task` becomes `await asyncio.shield(task)`. A cancelled caller still raises `CancelledError`, it just no longer takes the shared fetch down with it, and the fetch it started still completes and populates the cache.

`_cache` keying is untouched, so a completed fetch is still shared by everyone regardless of fetcher.

## Verification

- `test_a_different_fetcher_is_not_served_by_the_in_flight_one` and `test_one_caller_cancelling_does_not_abort_the_shared_fetch` fail on `ca89f01` and pass here. `test_same_fetcher_still_coalesces` passes on both and guards the coalescing #371 added.
- `tests/test_tool_cache.py`, `tests/test_tool_cache_empty_result.py`, `tests/test_allowed_jupyter_mcp_tools.py`: 17 passed on this branch in a clean `python:3.12-slim` container. Wider run including `test_jupyter_extension.py`, `test_identity*.py` and `test_auth.py`: 103 passed. `test_auth.py`'s password-auth E2E leaves 1 failure and 2 errors, identical on `ca89f01` and on this branch, so it is not from this change.
- `ruff check` on the three files: 18 findings on `ca89f01`, 17 here, all pre-existing and none new.
- Not exercised against a live JupyterLab frontend. The two call sites' key components were read from the source, and the concurrency itself was reproduced against the real `ToolCache` with stand-in fetchers rather than against real `jupyter-mcp-tools` traffic.

One judgement call worth flagging: keying by the function object means a caller that builds its fetcher inline never coalesces. That is why the wrapper moved rather than the key growing a name-based component. If you would rather `ToolCache.get_tools` took `wait_timeout` explicitly and kept a single fetcher, say so and I will send that instead.

### DIFF
--- a/jupyter_mcp_server/jupyter_extension/handlers.py
+++ b/jupyter_mcp_server/jupyter_extension/handlers.py
@@ -30,6 +30,21 @@ from jupyter_mcp_server.utils import clean_mcp_response, clean_mcp_response_cont
 logger = logging.getLogger(__name__)
 
 
+async def _fetch_jupyter_tools(**kwargs):
+    """Fetch jupyter-mcp-tools with the shorter timeout this transport wants.
+
+    If the JupyterLab frontend is not loaded there is nothing to wait for, so
+    tools/list should not sit here for the library default of 30 seconds.
+
+    Defined at module level on purpose: ToolCache keys its in-flight fetches by
+    fetcher as well as by query, so a function rebuilt per request would stop
+    concurrent tools/list requests from sharing one call.
+    """
+    from jupyter_mcp_tools import get_tools
+
+    return await get_tools(wait_timeout=5, **kwargs)
+
+
 class MCPSSEHandler(JupyterHandler):
     """
     Handler for MCP protocol over Streamable HTTP.
@@ -137,8 +152,6 @@ class MCPSSEHandler(JupyterHandler):
                     # Get tools from jupyter_mcp_tools extension first to identify duplicates
                     jupyter_tools_data = []
                     try:
-                        from jupyter_mcp_tools import get_tools
-
                         from jupyter_mcp_server.tool_cache import get_tool_cache
 
                         # Get the server's base URL dynamically from ServerApp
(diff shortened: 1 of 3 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick turnaround. The joining caller now runs its own fetcher, and the shared task survives a cancelled joiner, so the two paths that used to cross are separated at the key rather than by timing.

Issue #21, opened 2025-06-06
🔧 Kernel keeps losing connection between add code cell and execute cell


emilio-gagliardi:
## Description

add_execute_code_cell() keeps failing after adding a cell

The NbModelClient part is robust because it reconnects each time. The KernelClient part is not; it relies on a connection made at startup to persist indefinitely.

Cell Addition Works: Because NbModelClient creates a new connection each time.
Cell Execution Fails (often): Because the global KernelClient's connection may have died since the MCP server started, and there's no mechanism in add_execute_code_cell to refresh or re-establish it. The "Connection to remote host was lost. - goodbye" error you saw in the Docker logs earlier ([06/06/25 20:58:20]) was very likely the global KernelClient's connection dying. After that point, all calls to execute_cell would fail.
In summary: The code doesn't seem to have a mechanism to handle a disconnected global KernelClient. If that initial connection (established when the MCP server starts) drops, all subsequent attempts to execute code via notebook.execute_cell(cell_index, kernel) will fail because kernel is no longer in a usable state. This perfectly explains why cell additions are working (as they use a fresh NbModelClient connection) but executions are failing with a "Connection is already closed" error.

## Reproduce

created a notebook file
updated the mcp_config.json
restarted the IDE
instructed agent to add a new cell with something in it
the cell is inserted but the MCP reports the kernel is disconnected and can't execute the cell

I looked at the source code and reviewed it with Gemini 2.5 Pro

## Expected behavior
add + execute happens immediately

## Context

- Datalayer version: latest
- Operating System and version: windows 11
- Browser and version: Brave 

<details><summary>Browser Output</summary>
<pre>
I'm running the notebook with no browser command
</pre>
</details>

echarles:
Thank you so much for opening the issue @emili@emo-gagliardi 

One way to solve this would be to use a mecanism which runs the model (including the connection to the kernel) on the jupyter server side with https://github.com/datalayer/jupyter-server-nbmodel

I guess we should test, maybe refactor our code, and document... but maybe it can work out-of-the-box for your case?

I would suggest you to try first https://github.com/datalayer/jupyter-server-nbmodel and then see how it can be used with the Jupyter MCP Server. More than happy to support you in this journey, and will welcome any contribution (docs...) in that direction.

ebarkhordar:
Still there on 1.4.4 (`280d1c8`), MCP_SERVER mode over streamable-http, no IDE and no LLM in the loop.

[...]

PR #397, opened 2026-08-24
fix(sandbox): rebuild a code sandbox whose kernel the server no longer has

ebarkhordar:
`CodeSandboxClient.is_alive()` returns `is_started`, a flag set when this process called `start()`. It never asks the server, so it stays `True` after the kernel goes away: an idle cull, a `DELETE /api/kernels/<id>`, or a server restart. `ensure_code_sandbox_alive` is the guard meant to replace a dead sandbox, and it reads exactly that flag, so it finds nothing wrong and hands the caller a sandbox that can no longer execute.

Measured on `9908001` in a clean container, driving the real guard against a real Jupyter server, with `/api/kernels` empty in both rows:

```
kernel deleted server-side   new sandbox: no   execute -> stdout '' after 25.0s, "Timeout waiting for output"
jupyter server restarted     new sandbox: no   execute -> stdout '' at once,     "Connection is already closed."
```

In the first row the outcome still comes back with `success=True` and no output, which is the shape that worries me most: nothing upstream of the sandbox has been told the cell did not run.

## The fix

Ask the server. For a Jupyter-backed sandbox its kernel list is the authority, and it is the check `use_notebook` and `execute_code` already make before they accept a caller-supplied kernel id. `NotebookManager.ensure_code_sandbox_alive` takes an optional liveness callable, defaulting to today's `is_alive()` so no other caller changes, and `utils.ensure_code_sandbox_alive` supplies one that looks the sandbox id up in `/api/kernels`.

Two cases deliberately keep the current sandbox: a non-`jupyter-server` variant, whose kernel does not live on this server at all, and a failed lookup, which is not evidence about the kernel and should not cost a working one.

With the change, both rows above provision a replacement in about 1.1s and the next execution prints normally.

## Verification

- The new test fails on `main`, where the guard returns the culled sandbox, and passes here. The existing `ensure_kernel_alive` tests are untouched and still pass.
- Full suite in both server modes on Python 3.10 and 3.13, in a clean container: identical failures to `main` run in the same container, plus the 6 new tests. Several server-backed tests error there on `main` as well, so that set is not from this change.
- `ruff` and `mypy` over the two changed files report the same as `main`.
- Not checked: a connection that drops while the kernel is still running. The server keeps listing it, so the guard leaves that case exactly as it is today.

Refs #21. This covers the half where the kernel is gone from the server. The report also describes a websocket dropping under a kernel that is still alive, and that is untouched here.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -651,13 +651,44 @@ def start_code_sandbox(notebook_manager, config, logger):
         raise
 
 
+def code_sandbox_is_alive(code_sandbox: Any) -> bool:
+    """Whether `code_sandbox` can still run code.
+
+    `CodeSandboxClient.is_alive()` returns `is_started`, a flag set when this
+    process called `start()`, so it stays True after the kernel goes away on the
+    server side. For a Jupyter-backed sandbox the server's kernel list is the
+    authority, the same check `use_notebook` and `execute_code` already make
+    before they accept a caller-supplied kernel id.
+    """
+    if not hasattr(code_sandbox, "is_alive") or not code_sandbox.is_alive():
+        return False
+    if getattr(code_sandbox, "variant", None) != "jupyter-server":
+        return True
+
+    from jupyter_mcp_server.server_context import ServerContext
+
+    kernel_id = getattr(code_sandbox, "id", None)
+    sandbox_server_client = ServerContext.get_instance().sandbox_server_client
+    if not kernel_id or sandbox_server_client is None:
+        return True
+    try:
+        kernels = sandbox_server_client.kernels.list_kernels()
+    except Exception:
+        # A failed lookup says nothing about the kernel, so keep the sandbox
+        # rather than discarding a working one on a transient error.
+        return True
+    return any(kernel.id == kernel_id for kernel in kernels)
+
+
 def ensure_code_sandbox_alive(
     notebook_manager, current_notebook, create_code_sandbox_fn: Callable[[], CodeSandboxClient]
 ) -> CodeSandboxClient:
     """Ensure the notebook's code sandbox is running, restart if needed."""
     return cast(
         CodeSandboxClient,
-        notebook_manager.ensure_code_sandbox_alive(current_notebook, create_code_sandbox_fn),
+        notebook_manager.ensure_code_sandbox_alive(
+            current_notebook, create_code_sandbox_fn, is_alive_fn=code_sandbox_is_alive
+        ),
     )
 
 
--- a/jupyter_mcp_server/notebook_manager.py
+++ b/jupyter_mcp_server/notebook_manager.py
@@ -44,6 +44,11 @@ logger = logging.getLogger(__name__)
 DISCONNECT_TIMEOUT = 20
 
 
+def _code_sandbox_reports_alive(code_sandbox: Any) -> bool:
+    """Ask the sandbox object itself whether it is alive."""
+    return hasattr(code_sandbox, "is_alive") and bool(code_sandbox.is_alive())
+
+
 class NotebookConnection:
     """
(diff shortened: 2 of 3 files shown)


jupyter_mcp_server/utils.py
```diff
@@ -651,13 +651,44 @@ def start_code_sandbox(notebook_manager, config, logger):
         raise
 
 
+def code_sandbox_is_alive(code_sandbox: Any) -> bool:
```
echarles, line 654: That logic should its place in the code-sandboxes library.

ebarkhordar: Agreed, and thanks for the steer.

Reading `code_sandboxes/client.py:437`, `is_alive()` returns `self.is_started`, and the `Sandbox` ABC exposes only `is_started`, so a variant has no place to answer the question its own way. That is the level this belongs at: a Jupyter-backed sandbox knows its kernel lives in a server's kernel list, and the other variants each have a different authority to ask.

Happy to take it there: a liveness method on `Sandbox` that variants override, `JupyterServerSandbox` looking its kernel id up in `/api/kernels`, and `CodeSandboxClient.is_alive()` delegating to the sandbox rather than returning the local flag. This PR then reduces to a version bump, or to nothing at all.

Would you rather I close this one now and open there, or leave it open until the library side lands?

echarles: > Happy to take it there: a liveness method on Sandbox that variants override, JupyterServerSandbox looking its kernel id up in /api/kernels, and CodeSandboxClient.is_alive() delegating to the sandbox rather than returning the local flag. This PR then reduces to a version bump, or to nothing at all.

Sounds like a plan. Thx!

> Would you rather I close this one now and open there, or leave it open until the library side lands?

Please close and open a new one with the version bump

ebarkhordar:
The `enforce-label` check here is red only because this PR has no triage label yet, and I cannot apply one from a fork. Whenever you get a moment to label it the check goes green on the rerun, the same as it did on #399.

echarles:
I have set the label. Does this need a PR on code-sandboxes?

ebarkhordar:
Thanks for the label, the rerun went green.

On code-sandboxes: yes, that is where the root cause sits. `CodeSandboxClient.is_alive()` is `return self.is_started` (`code_sandboxes/client.py:437-439`), so the stale flag is produced there rather than here.

I scoped this PR so it does not wait on that. The guard takes an optional liveness callable and defaults to today's `is_alive()`, so the behaviour for every other caller is unchanged whether or not code-sandboxes moves.

A fix upstream is a larger piece than it first looks. `ISandboxClient` has no liveness member at all (`id`, `kernel_info`, `start`, `stop`, `execute`, `execute_interactive`, `get_variable`, `set_variable`, `interrupt`), so a general one means adding it to the protocol and implementing it across the 12 sandbox variants, each of which has a different service to ask. The jupyter-server variant is the easy one, since `_find_existing_kernel` already lists kernels through `jupyter-server-client`.

Happy to open that PR if you want it. Tell me whether you would rather have a protocol-wide liveness method or just an override on the jupyter-server variant, and I will follow whichever you prefer.

echarles:
> Happy to open that PR if you want it. Tell me whether you would rather have a protocol-wide liveness method or just an override on the jupyter-server variant, and I will follow whichever you prefer.

A protocol-wide liveness method for all the variant would be great to have in the code-sandboxes repo.

In the meantime I will merge this. Thank you!

echarles (approved):
LGTM

ebarkhordar:
Thanks for merging, and for pointing this at the right layer.

I will open the protocol-wide version on code-sandboxes: a liveness member on the sandbox protocol with an implementation per variant, and the jupyter-server one resolving its kernel id through jupyter-server-client the way `_find_existing_kernel` already does. I will link it here when it is up.

PR #331, opened 2026-07-31
rename runtime to code sandbox

echarles:
Fixes https://github.com/datalayer/jupyter-mcp-server/issues/330

PR #399, opened 2026-08-24
fix(execute): let a missing kernel reach execute_cell's retry instead of becoming an output

### LINKED ISSUE #398: execute_cell's dead-kernel retry never runs: execute_via_execution_stack reports a missing kernel as a returned output, not an exception
## Environment

- jupyter-mcp-server at `99080012febf0c0694c29787b6033c49cc06eaaf` (current `main`), installed with `pip install -e ".[test]"`
- python 3.12-slim container, JUPYTER_SERVER mode

## Summary

`execute_cell` wraps both of its JUPYTER_SERVER-mode calls to `execute_via_execution_stack` in a retry that starts a replacement kernel when the kernel went away underneath it (`execute_cell_tool.py:340` for the RTC branch, `:398` for the file branch):

```python
except Exception as error:
    error_text = str(error).lower()
    if "kernel" in error_text and "not found" in error_text:
        ...
        kernel_id = await self._start_and_bind_kernel(...)
        outputs = await execute_via_execution_stack(...)
```

That handler cannot run. `execute_via_execution_stack` reports a missing kernel as a completed request carrying an `error` key, and the `if "error" in result:` branch (`utils.py:1027`) returns a formatted output at `:1065` rather than raising. Nothing else leaves the function either: its whole body is one `try` whose only handler is `except Exception as e: ... return [f"[ERROR: {e!s}]"]`, so the `RuntimeError` at `utils.py:965` and the `TimeoutError` at `utils.py:998` are caught by it too. Only `asyncio.CancelledError` gets out, and that is not an `Exception`.

So a cell executed against a culled kernel returns the error text to the user, and the self-healing path that exists right above it never gets a chance. The `Raises: RuntimeError / TimeoutError` lines in the function's own docstring are inaccurate for the same reason.

The retry blocks were added in #331. The window they cover is real but narrow: `execute_cell` now pre-checks with `_kernel_exists` before submitting, so what is left is the gap between that check and `ExecutionStack.put`, plus the case where `_kernel_exists` fails open (its `except Exception` returns `True`). I have not driven that race end to end against a live server, so I am not claiming a frequency for it, only that the handler is dead code today.

## Steps to reproduce

Nothing here needs a live server. The `ExecutionStack` stand-in returns the exact terminal result jupyter-server-nbmodel writes for an unknown kernel (see #394 for where that string comes from), and the kernel manager still lists the culled kernel, which is what makes the pre-check pass and the request reach the stack:

```python
import asyncio, json
from types import SimpleNamespace
import nbformat
import jupyter_mcp_server
from jupyter_mcp_server.tools._base import ServerMode
from jupyter_mcp_server.tools.execute_cell_tool import ExecuteCellTool
import jupyter_mcp_server.tools.execute_cell_tool as tool_module
import jupyter_mcp_server.jupyter_extension.context as context_module

GONE = "HTTP 404: Not Found (Kernel does not exist: kernel-1)"

class Stack:
    def __init__(self): self.puts = []
    def put(self, kernel_id, code, metadata):
        self.puts.append(kernel_id); return f"request-{len(self.puts)}"
    def get(self, kernel_id, request_id):
        if request_id == "request-1":
            return {"error": GONE, "pending": False, "request_status": "complete"}
        return {"pending": False, "request_status": "complete", "status": "ok",
                "execution_count": 1,
                "outputs": json.dumps([{"output_type": "stream", "name": "stdout",
                                        "text": "ran on the replacement\n"}])}
    def cancel(self, kernel_id): pass

class Ext:
    def __init__(self, s): self._Extension__execution_stack = s
class EM:
    def __init__(self, e): self.extension_apps = {"jupyter_server_nbmodel": {e}}
class FileIds:
    def get_id(self, p): return "file-id"
    def index(self, p): return "file-id"

class StaleKernels:
    def __init__(self): self.started = []
    def list_kernels(self): return [{"id": "kernel-1"}]      # server has not culled it yet
    async def start_kernel(self, path=None): self.started.append(path); return "kernel-2"

async def main():
    print(jupyter_mcp_server.__file__)
    stack = Stack()
    serverapp = SimpleNamespace(root_dir="/srv/notebooks",
        web_app=SimpleNamespace(settings={"file_id_manager": FileIds()}),
        extension_manager=EM(Ext(stack)))
    context_module.get_server_context = lambda: SimpleNamespace(serverapp=serverapp)
    tool_module.get_current_notebook_context = lambda _m: ("demo.ipynb", "kernel-1")
    async def no_ydoc(a, b): return None
    async def no_sleep(s): return None
    async def no_write(*a, **k): return None
    async def read_nb(p):
        return nbformat.v4.new_notebook(cells=[nbformat.v4.new_code_cell("print('hi')")])
    tool_module.get_jupyter_ydoc = no_ydoc
    tool_module.asyncio.sleep = no_sleep
    ExecuteCellTool._write_outputs_to_cell = no_write
    tool = ExecuteCellTool(); tool._read_notebook_file_with_retry = read_nb
    kernels = StaleKernels()

    outputs = await tool.execute(mode=ServerMode.JUPYTER_SERVER,
        kernel_manager=kernels, notebook_manager=NotebookManager(), cell_index=0)
    print("returned           :", outputs)
    print("replacement kernels:", kernels.started)
    print("requests submitted :", stack.puts)

asyncio.run(main())
```

## Actual

```
/build/app/jupyter_mcp_server/__init__.py
returned           : ['[ERROR: ExecutionError: HTTP 404: Not Found (Kernel does not exist: kernel-1)]']
replacement kernels: []
requests submitted : ['kernel-1']
```

One request, no replacement kernel, and the `Kernel %s disappeared during execute_cell` warning never logged.

## Expected

The retry runs once, so the cell executes on the replacement kernel:

```
returned           : ['ran on the replacement\n']
replacement kernels: ['demo.ipynb']
requests submitted : ['kernel-1', 'kernel-2']
```

## PR, and one question in it

I have opened a PR for the narrow version of this: `execute_via_execution_stack` raises a `MissingKernelError` for that one request-level failure and returns a formatted output for every other, which leaves the existing retry untouched and makes it live.

The question I could not answer from outside the project is whether you want the wider version. `execute_cell` decides to retry by lowercasing the exception text and looking for `kernel` and `not found`, and the PR has to apply the same two-word test in `utils.py` to pick which failure to raise. That is the same predicate written twice, and if either copy drifts the retry goes quietly dead again in exactly the way it is dead now. Having the caller catch `MissingKernelError` instead of sniffing the string would leave one copy, but it changes your code rather than just unblocking it, so I left it out. Happy to add it if you would prefer that shape.

ebarkhordar:
Fixes #398

## The problem

`execute_cell` starts a replacement kernel and retries once when the kernel it was using went away, in both JUPYTER_SERVER branches (`execute_cell_tool.py:340` and `:398`). It looks for that failure in an exception, and `execute_via_execution_stack` never produces one: a missing kernel arrives as a completed request carrying an `error` key, and that branch returns a formatted output at `utils.py:1065`. The rest of the function is covered by a single `except Exception` that turns every other failure into an output too, so nothing but `asyncio.CancelledError` leaves it.

The retry has therefore been dead code since it was added in #331. Details and a standalone repro are in the issue.

## The change

One request-level failure now leaves as an exception, and every other one is unchanged:

- `MissingKernelError`, a `RuntimeError` subclass, plus `is_missing_kernel_message`, which applies the same two-word test `execute_cell` already uses to decide whether to retry.
- The `if "error" in result:` branch raises it instead of returning, when the reason reads as a missing kernel.
- The outer handler re-raises that one type after firing `AFTER_EXECUTE`, so the pairing contract from #360 still holds: exactly one `AFTER_EXECUTE`, carrying the exception, on this exit as on the others.

`execute_cell` is untouched. Its existing predicate matches by construction, because the same predicate is what selected the failure for raising.

## What I did not change, and why it is a question rather than a decision

Having `utils.py` apply the caller's own string test means that predicate is now written in two places, and if either copy drifts the retry goes quietly dead again in exactly the way it is dead now. Letting `execute_cell` catch `MissingKernelError` and drop its string sniff would leave one copy, but that edits working code of yours rather than unblocking it, so I left it out of this PR. Say the word and I will add it here.

## Behaviour change to be aware of

If the retry also fails because the replacement kernel is gone too, the second `MissingKernelError` propagates out of `execute_cell` rather than being returned as an output string. `execute()` already raises `ValueError` and `RuntimeError` on several paths, so this is not a new mode for the tool, but it is a real difference from today, where that case returns text.

## Verification

Clean `python:3.12-slim` container, `pip install -e ".[test]"`, `__file__` = `/build/app/jupyter_mcp_server/__init__.py`, both sides of the differential in the same container.

The issue's repro script, run verbatim, at `99080012` and with this patch:

```
baseline  returned           : ['[ERROR: ExecutionError: HTTP 404: Not Found (Kernel does not exist: kernel-1)]']
          replacement kernels: []
          requests submitted : ['kernel-1']

patched   returned           : ['ran on the replacement\n']
          replacement kernels: ['demo.ipynb']
          requests submitted : ['kernel-1', 'kernel-2']
```

### Tests

`tests/test_execute_cell_dead_kernel_retry.py` is new and covers the two claims above:

- `test_missing_kernel_raises_instead_of_returning_a_formatted_output` for the helper
- `test_execute_cell_starts_a_replacement_kernel_and_retries_once` drives the real `ExecuteCellTool.execute()` over the real `execute_via_execution_stack`, so nothing on the fault path is stubbed apart from the `ExecutionStack` and the kernel manager the server would supply

`tests/test_execute_hook_pairing.py` gains `test_missing_kernel_fires_after_execute_before_it_propagates` for the pairing claim.

`tests/test_execute_via_execution_stack.py::test_string_shaped_error_keeps_its_message` is the one existing expectation this changes. It used the `HTTP 404 ... Kernel does not exist` payload, which now raises, so it moves to the superseded-request payload and keeps covering the string-shaped wrapping from #395. The kernel payload is covered by the new file.

Targeted, on the patched tree:

```
tests/test_execute_cell_dead_kernel_retry.py     2 passed   (new)
tests/test_execute_hook_pairing.py             11 passed
tests/test_execute_via_execution_stack.py       3 passed
                                               16 passed in 3.58s

plus the rest of the execute surface, unchanged:
tests/test_execute_cell_kernel_start_path.py, test_execute_cell_output_fidelity.py,
test_execute_cell_kernel_execution_count.py, test_execute_orphaned_timeout.py,
test_execute_progress_keepalive.py, test_execute_cell_stream.py
                                               43 passed in 37.25s
```

Whole suite, same container both sides, `tests/` with `test_execute_cell_sandbox_routing.py` ignored (it needs the optional `jupyter_mcp_sandboxes` package, which is not installed here):

```
99080012          2 failed, 375 passed, 1 skipped, 1 warning, 121 errors in 80.04s
99080012 + this   2 failed, 378 passed, 1 skipped, 1 warning, 121 errors in 77.97s
```

The 121 errors and 2 failures are identical on both sides, module for module and count for
count. They are the tests that need the live `jupyter_server` fixture, which does not come up
in this bare container: `test_tools.py` 37, `test_move_cell.py` 28, `test_edit_cell_source.py`
24, `test_auth.py` 9, `test_jupyter_extension.py` 7, `test_otel_integration.py` 6,
`test_use_notebook_ui_open.py` 4, `test_execute_code_kernel_id.py` 4,
`test_extension_legacy_hook.py` 3, `test_use_notebook_split_servers.py` 2,
`test_notebook_connection_disconnect_timeout.py` 1. I could not get a clean local baseline for
those, so CI is the real check there. The +3 passed are the three tests this PR adds.

### Not verified

I have not driven a live culling race end to end. `execute_cell` pre-checks with `_kernel_exists` before submitting, so the window this retry covers is the gap between that check and `ExecutionStack.put`, plus the case where `_kernel_exists` fails open. I am not claiming how often that is hit, only that the handler cannot run today and does run after this change.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -897,6 +897,25 @@ async def safe_notebook_operation(operation_func, max_retries=3):
 ###############################################################################
 
 
+class MissingKernelError(RuntimeError):
+    """The kernel an execution request targeted no longer exists on the server.
+
+    Callers that can start a replacement kernel and retry need this failure as an
+    exception; every other request-level failure stays a formatted output.
+    """
+
+
+def is_missing_kernel_message(message: Any) -> bool:
+    """Whether *message* reads as the server reporting an unknown kernel.
+
+    ExecutionStack reports this as free text rather than a code, so this matches
+    the same two words ``execute_cell`` already looks for before it starts a
+    replacement kernel.
+    """
+    text = str(message).lower()
+    return "kernel" in text and "not found" in text
+
+
 async def execute_via_execution_stack(
     serverapp: Any,
     kernel_id: str,
@@ -944,6 +963,7 @@ async def execute_via_execution_stack(
     Raises:
         RuntimeError: If jupyter-server-nbmodel extension is not installed
         TimeoutError: If execution exceeds timeout
+        MissingKernelError: If the request failed because ``kernel_id`` is gone
     """
     import logging as default_logging
 
@@ -1041,6 +1061,13 @@ async def execute_via_execution_stack(
                                 "traceback": [],
                             }
                         logger.error(f"Execution error: {error_info}")
+                        if is_missing_kernel_message(error_info.get("evalue", "")):
+                            # execute_cell starts a replacement kernel and retries
+                            # once when this happens, and it looks for the failure
+                            # in an exception. Leave as one so that path can run;
+                            # the handler at the end of this function fires the
+                            # single AFTER_EXECUTE this execution owes and re-raises.
+                            raise MissingKernelError(error_info.get("evalue", ""))
                         error_output = [
                             f"[ERROR: {error_info.get('ename', 'Unknown')}: {error_info.get('evalue', '')}]"
                         ]
@@ -1175,6 +1202,8 @@ async def execute_via_execution_stack(
                 error=e,
                 context=hook_ctx,
             )
+        if isinstance(e, MissingKernelError):
+            raise
         return [f"[ERROR: {e!s}]"]
 
 
--- a/tests/test_execute_cell_dead_kernel_retry.py
+++ b/tests/test_execute_cell_dead_kernel_retry.py
@@ -1,0 +1,147 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""execute_cell's dead-kernel retry in JUPYTER_SERVER mode.
+
+ExecutionStack reports a kernel it cannot reach as a terminal result rather than
+by raising, so the retry that execute_cell wraps around it needs that one failure
+to leave execute_via_execution_stack as an exception. These tests drive the real
+helper, so nothing on the fault path is stubbed out apart from the ExecutionStack
+and the kernel manager the server would provide.
+"""
+
+import json
+from types import SimpleNamespace
+
+import nbformat
+import pytest
+
+from jupyter_mcp_server.notebook_manager import NotebookManager
+from jupyter_mcp_server.tools._base import ServerMode
+from jupyter_mcp_server.tools.execute_cell_tool import ExecuteCellTool
+from jupyter_mcp_server.utils import MissingKernelError, execute_via_execution_stack
+
+KERNEL_GONE = "HTTP 404: Not Found (Kernel does not exist: kernel-1)"
+
+
+class DeadKernelExecutionStack:
+    """First request reports the kernel gone; a later one runs normally."""
+
+    def __init__(self):
+        self.put_kernels: list[str] = []
+
+    def put(self, kernel_id, code, metadata):
+        self.put_kernels.append(kernel_id)
+        return f"request-{len(self.put_kernels)}"
+
+    def get(self, kernel_id, request_id):
+        if request_id == "request-1":
+            return {"error": KERNEL_GONE, "pending": False, "request_status": "complete"}
+        return {
+            "pending": False,
+            "request_status": "complete",
+            "status": "ok",
+            "execution_count": 1,
+            "outputs": json.dumps(
+                [{"output_type": "stream", "name": "stdout", "text": "ran on the replacement\n"}]
+            ),
+        }
+
+    def cancel(self, kernel_id):
+        pass
+
+
+class _Extension:
+    def __init__(self, execution_stack):
+        self._Extension__execution_stack = execution_stack
+
+
+class _ExtensionManager:
+    def __init__(self, extension):
+        self.extension_apps = {"jupyter_server_nbmodel": {extension}}
+
+
+class _FileIdManager:
+    def get_id(self, _path):
+        return "file-id"
+
+    def index(self, _path):
+        return "file-id"
+
+
+class StaleKernelManager:
+    """Still lists the culled kernel, so execute_cell's pre-check passes and the
+    request reaches an ExecutionStack that knows the kernel is gone."""
+
+    def __init__(self):
+        self.started_paths: list[str] = []
+
+    def list_kernels(self):
+        return [{"id": "kernel-1"}]
+
+    async def start_kernel(self, path=None):
+        self.started_paths.append(path)
+        return "kernel-2"
+
+
+@pytest.mark.asyncio
+async def test_missing_kernel_raises_instead_of_returning_a_formatted_output():
+    stack = DeadKernelExecutionStack()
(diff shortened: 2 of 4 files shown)


echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick review and merge. The predicate is still written in two places, so if you want `execute_cell` to catch `MissingKernelError` and drop its own string test, I am happy to send that as a small follow-up.

echarles:
>  and drop its own string test, I am happy to send that as a small follow-up.

yes, that would be great. Thx you!

PR #349, opened 2026-08-06
fix(deps): update to jupyter-collaboration 5 by removing leftover datalayer_pycrdt install steps

pokutuna:
Fixes #348

## What

- Remove the `pip uninstall -y pycrdt datalayer_pycrdt` / `pip install datalayer_pycrdt==0.12.17` step from `.github/workflows/test.yml`, `Dockerfile`, `Makefile`, `CONTRIBUTING.md`, and `.github/copilot-instructions.md`.
- Allow jupyter-collaboration 5 again by reverting the `<5` bound added in #347.

## Background

- #240 / #241
  - Switched the project to upstream pycrdt.
- #242
  - Removed `datalayer_pycrdt` from the `test` extra, but the install steps above were left behind.
  - They stayed harmless while jupyter-collaboration 4.x kept jupyter-ydoc below 4.0, which still used the old `UndoManager(doc=...)` API.
- #347
  - jupyter-collaboration 5.0.0 (Aug 3) pulls in jupyter-ydoc 4.x, which follows the new pycrdt >=0.14 API (`doc` argument removed). Swapping in `datalayer_pycrdt 0.12.17` (old API) on top of that makes every Y document creation raise `RuntimeError: UndoManager must be created with doc or scopes` — the reported failures, all `[mcp_server]` cases, are downstream of this. Pinned `<5` as a stopgap.

Verified locally: every file from the #348 failure list passes on the jupyter-collaboration 5.0.0 stack, with no code changes to jupyter_mcp_server.

Issue #263, opened 2026-07-08
RTC mode: executing a cell deletes execution_count from all code cells; saved notebook fails nbformat validation


mattnworb:
### Environment

- `jupyter-mcp-server`: 1.0.0 (also checked 1.0.2 — the two tags are code-identical except `__version__.py`)
- Client env (resolved by uvx): `jupyter-nbmodel-client` 0.14.8, `jupyter-kernel-client` 0.9.0, `jupyter-ydoc` 4.1.0, `pycrdt` 0.14.1
- Jupyter server: `jupyterlab` 4.4.1, `jupyter-server` 2.18.2, `jupyter-collaboration` 4.0.2, `jupyter-server-ydoc` 2.3.0, `jupyter-ydoc` 3.4.1
- Mode: `MCP_SERVER` (stdio, `JUPYTER_URL` pointing at a local JupyterLab with RTC enabled)
- OS: macOS (Darwin 25.5.0)

### Problem

Notebooks touched by MCP tools end up as **schema-invalid nbformat JSON**. Two distinct defects:

1. **Executing any cell removes `execution_count` (and `execution_state`) from *all* code cells in the notebook.** The saved file then fails `nbformat.validate` with `'execution_count' is a required property`, and `jupyter nbconvert` refuses to read it.
2. **`display_data` outputs keep the kernel protocol's `transient` field**, which is not part of the nbformat schema — a second validation failure.

### Minimal repro (via MCP tools)

1. `use_notebook` with `mode="create"` → new notebook.
2. `insert_cell` (code) → wait for RTC autosave → **file is valid**: cell has `"execution_count": null`. ✅
3. `execute_cell` on any cell → wait for autosave → **file is invalid**: the `execution_count` key is now absent from *every* code cell (not just the executed one). ❌

Executing a cell that produces `display_data` (e.g. a matplotlib figure) additionally leaves `"transient": {}` in the saved output.

### Evidence that it's a doc-level event, not a cell write

I attached an independent `NbModelClient` watcher to the same room, polling `ycell.to_py()` every 300 ms during an MCP `execute_cell`:

```
  0.00s  c0:ec=Y(7.0)/st=Y | c1:ec=N/st=N
  3.02s  c0:ec=N/st=N      | c1:ec=N/st=N
```

[...]

PR #402, opened 2026-08-25
fix(use_notebook): warn when the document server's collaboration stack predates ours

ebarkhordar:
## Problem

In MCP_SERVER mode the document server is whatever the user points `--document-url` at, and nothing checks what it runs. When that server's collaboration stack predates this project's own (`jupyter-collaboration>=5`, unpinned here since #349), an in-place scalar write to a cell arrives as a deletion of that key. `execute_cell` still reports `[COMPLETED]`, but the notebook that gets saved has no `execution_count` on the cell that ran, so `nbformat.validate` rejects the file.

That is the first defect in #263. The wire behaviour is not something this project can fix. What it can do is notice the condition instead of reporting success.

## What I measured

Two Jupyter servers in one container, the same MCP client driving both, only the remote RTC stack differing:

| document server | `execution_count` after `execute_cell` |
|---|---|
| jupyter-collaboration 4.0.2 | absent, `nbformat.validate` fails |
| jupyter-collaboration 5.0.2 | `1`, validates |

The loss is confined to cells the client writes to: the cell that was never executed still carried its `execution_count` afterwards, and against the older server `nbformat.validate` fails with `'execution_count' is a required property` on the cell that ran.

One measurement decided the shape of this change. The executing client's own replica still holds `execution_count` at +1s, +3s and +5s against the server that dropped it, so a read-back taken inside the connection that performed the write cannot see the loss, and no settle time changes that. The condition is visible earlier and more cheaply: at connect time `/lab/api/extensions` reports `jupyter-collaboration-extension` 4.4.2 on one server and 5.0.2 on the other.

## The change

`use_notebook` reads that endpoint once, in MCP_SERVER mode with a Jupyter document provider, and appends a `[WARNING]` line when the remote major is below ours. Majors only. A probe that errors, a server with no such endpoint, and a payload without the extension all leave the tool silent, because an unanswered probe is not evidence of a healthy server.

You asked on #263 whether the fix belonged in `jupyter-kernel-client`. For this defect I do not think it belongs in either client: the drop happens in the collaboration room protocol between a v5 client and a pre-5 server. The part that is ours is that we drive a remote we never version-check and report success while the file on disk is going invalid.

## Verification

- `tests/test_use_notebook_remote_rtc_version.py`, four cases. The warning case fails on `main` and passes here; the three silence cases guard against false positives and pass on both.
- The full suite in both server modes goes from 469 to 473 passed, the four being these tests, with the failure and error counts unchanged from `main` (one pre-existing `test_otel_integration` failure, plus the JupyterLab-backed integration tests my container cannot start, identical on both trees).
- The extension suites (81) and the `use_notebook` suite on Python 3.10 and 3.13, the matrix extremes. The two-server differential above ran on 3.11 and is field evidence rather than a test in this suite.
- Not checked: whether every pre-5 minor drops scalar writes. I measured 4.0.2 and 5.0.2 only, so the major comparison rests on two points, and a document server without JupyterLab has no `/lab/api/extensions` and will not be reported at all.

Refs #263. This adds the warning only. The dropped write itself is upstream and is not addressed here.

### DIFF
--- a/jupyter_mcp_server/tools/use_notebook_tool.py
+++ b/jupyter_mcp_server/tools/use_notebook_tool.py
@@ -4,6 +4,7 @@
 
 """Use notebook tool implementation."""
 
+import importlib.metadata
 import inspect
 import logging
 from pathlib import Path
@@ -19,6 +20,46 @@ from jupyter_mcp_server.notebook_manager import NotebookManager
 from jupyter_mcp_server.tools._base import BaseTool, ServerMode
 
 logger = logging.getLogger(__name__)
+
+COLLABORATION_EXTENSION = "jupyter-collaboration-extension"
+
+
+def _major(version: str | None) -> int | None:
+    """Leading integer of a version string, or None when it does not start with one."""
+    head = str(version).split(".", 1)[0]
+    return int(head) if head.isdigit() else None
+
+
+def _outdated_collaboration_warning(server_client: JupyterServerClient) -> str | None:
+    """Warn when the document server's collaboration stack predates ours.
+
+    Returns None when the remote is current or its version could not be read: a
+    probe that fails is not evidence of a match, so the caller stays silent
+    rather than reporting a healthy server.
+    """
+    try:
+        local = importlib.metadata.version("jupyter-collaboration")
+        extensions = server_client.http_client.request(
+            "GET", "/lab/api/extensions", timeout=10
+        )
+        remote = next(
+            entry.get("installed_version")
+            for entry in extensions
+            if entry.get("name") == COLLABORATION_EXTENSION
+        )
+    except Exception as error:
+        logger.debug("Could not read the document server's collaboration version: %s", error)
+        return None
+    local_major, remote_major = _major(local), _major(remote)
+    if local_major is None or remote_major is None or remote_major >= local_major:
+        return None
+    return (
+        f"[WARNING] The document server runs {COLLABORATION_EXTENSION} {remote}, older than "
+        f"this MCP server's jupyter-collaboration {local}. Cell metadata written in place is "
+        f"dropped before it reaches the saved notebook, so an executed cell loses its "
+        f"execution_count and the file stops validating against nbformat. "
+        f"See https://github.com/datalayer/jupyter-mcp-server/issues/263."
+    )
 
 
 class UseNotebookTool(BaseTool):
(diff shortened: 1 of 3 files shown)


jupyter_mcp_server/tools/use_notebook_tool.py
```diff
@@ -20,6 +21,46 @@
...
+def _outdated_collaboration_warning(server_client: JupyterServerClient) -> str | None:
+    """Warn when the document server's collaboration stack predates ours.
+
+    Returns None when the remote is current or its version could not be read: a
+    probe that fails is not evidence of a match, so the caller stays silent
+    rather than reporting a healthy server.
+    """
+    try:
+        local = importlib.metadata.version("jupyter-collaboration")
+        extensions = server_client.http_client.request(
```
echarles, line 57: This would generate a request to the server on each use_notebook tool request. Can we make that request only once on file level and cache the reply to avoid unneeded requests.

echarles:
Looks like docs needs to be regenerated.

ebarkhordar:
Both done, thanks.

**Caching.** You were right that it fired on every call. The verdict is now cached per document server URL, so the extension list is fetched once. One asymmetry worth flagging because it is deliberate: only a verdict the probe actually reached is stored. A server does not change its installed extensions while it runs, so a successful read is good for the process, but a probe that failed says nothing, and caching that silence would retire the warning for good over a single timeout. Two tests pin it, one for the single fetch and one for the retry after a failed probe.

**Docs.** Regenerated and committed, though the staleness is not from this branch. `docs/sourcey/mcp.json` records the installed MCP SDK version in its server snapshot, and `mcp` 1.29.1 was published on 2026-08-24 at 18:30 UTC, about an hour after the last docs run on main. Running the pipeline from `docs/sourcey/README.md` here produces exactly the one line CI reported, `1.29.0` to `1.29.1`, and nothing else, so main is stale too and the next pull request after each SDK release will hit this. Happy to open a separate issue if you want the snapshot to stop tracking the SDK version.

Verified in a clean container: the six tests in `tests/test_use_notebook_remote_rtc_version.py` pass. The sibling `use_notebook` tests fail identically on an unmodified `main` checkout here, on a missing `jupyter_mcp_tools` import, so that is my container and not this branch.

echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the review and the merge, and for flagging the docs regeneration. This one covers the first defect on #263 only, which is why the body says Refs rather than Fixes, so that issue stays open.

PR #406, opened 2026-08-27
fix: validate inserted cell types at runtime

mikemikimike:
## Summary
- validate `insert_cell` cell types at runtime before dispatching to any backend
- return a clear error for unsupported types instead of reporting success without inserting a cell
- add unit coverage for supported and unsupported values

## Validation
- `pytest tests/test_insert_cell_validation.py -q` (blocked during collection because the isolated environment could not complete installation of the project's `mcp` dependency)
- `ruff check tests/test_insert_cell_validation.py jupyter_mcp_server/tools/insert_cell_tool.py --ignore E501,RUF013` passed
- `git diff --check` passed

### DIFF
--- a/jupyter_mcp_server/tools/insert_cell_tool.py
+++ b/jupyter_mcp_server/tools/insert_cell_tool.py
@@ -41,6 +41,11 @@ class InsertCellTool(BaseTool):
             IndexError: When cell_index is out of valid range
             ValueError: When cell_type is invalid
         """
+        if cell_type not in ("code", "markdown"):
+            raise ValueError(
+                f"Invalid cell type {cell_type!r}; expected 'code' or 'markdown'."
+            )
+
         if cell_index < -1 or cell_index > total_cells:
             raise IndexError(
                 f"Index {cell_index} is outside valid range [-1, {total_cells}]. "
--- a/jupyter_mcp_server/tools/insert_cell_tool.py
+++ b/jupyter_mcp_server/tools/insert_cell_tool.py
@@ -41,9 +41,9 @@ class InsertCellTool(BaseTool):
             IndexError: When cell_index is out of valid range
             ValueError: When cell_type is invalid
         """
-        if cell_type not in ("code", "markdown"):
+        if cell_type not in ("code", "markdown", "raw"):
             raise ValueError(
-                f"Invalid cell type {cell_type!r}; expected 'code' or 'markdown'."
+                f"Invalid cell type {cell_type!r}; expected 'code', 'markdown', or 'raw'."
             )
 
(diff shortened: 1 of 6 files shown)


jupyter_mcp_server/tools/insert_cell_tool.py
```diff
@@ -41,6 +41,11 @@ def _validate_cell_insertion_params(
             IndexError: When cell_index is out of valid range
             ValueError: When cell_type is invalid
         """
+        if cell_type not in ("code", "markdown"):
```
ebarkhordar: I ran the new test at `6bf0b02` in a clean python:3.12-slim container, since the PR notes say collection was blocked: 3 passed, and on the merge base `9908001` the rejection case fails with `DID NOT RAISE ValueError`, so it is load-bearing.

The check lands on all three insert paths, and `raw` works on two of them today. `_insert_cell_ydoc` (line 91) and `_insert_cell_websocket` (line 181) both hand `cell_type` to `NotebookModel.insert_cell`, and `NbModelClient` inherits the same function object:

```
signature: (self, index: int, source: str, cell_type: Literal['code', 'markdown', 'raw'], **kwargs)
nb.insert_cell(-1, "some raw text", "raw")
RESULT cell types: ['code', 'raw']
```

So this rejects an insert that currently succeeds. Only `_insert_cell_file` was broken, and not in the way the body describes: it has no `else` after 135-138, so `raw` raised `UnboundLocalError` rather than reporting success.

If raw should keep working, widen the tuple to `("code", "markdown", "raw")` and add the missing branch:

```python
elif cell_type == "raw":
    new_cell = nbformat.v4.new_raw_cell(source=cell_source or "")
```

If the restriction is deliberate, the `Literal["code", "markdown"]` annotations already say so, and the raw removal is worth naming in the body.

I did not test the MCP_SERVER path against a live server.

mikemikimike: Thanks for the detailed verification. The follow-up commit 2a465ec preserves raw cells across the validation and file-insertion paths, including nbformat.v4.new_raw_cell, and the regression test now covers raw insertion. I reran pytest with a writable temporary directory: 6 passed. I will mark this thread resolved.

echarles (approved):
LGTM Thx @mikemikimike

jupyter_mcp_server/tools/insert_cell_tool.py
ebarkhordar: Thanks. I read 2a465ec and it covers the two paths I measured, plus the request models and the tool signature, which I had not flagged. I did not rerun the suite against the merged head.

Issue #400, opened 2026-08-24
Support Capabilities to abstract options/features


echarles:
As the code base and behaviors grows, we should consider introducing Capabilities to abstract options an features.

See eg https://github.com/datalayer/jupyter-mcp-server/issues/398#issuecomment-5391737301

echarles:
Closing as fixed with https://github.com/datalayer/jupyter-mcp-server/pull/409

PR #409, opened 2026-08-27
Capabilities, cell ids, one result shape, an audit seam, and the conformance suite

echarles:
Twelve changes toward 2.1.0, most of them about the same thing: a server does
work an agent cannot see, and the failures that follow are quiet ones.

## Making the invisible visible and switchable

**A capability registry** (#400, #398). When a kernel died the server quietly
started another. Every variable, import and definition of the session went
with the old one, the notebook's state was not what the agent believed, and
nothing said so — the next `execute_cell` simply behaved as if the session had
always been empty. That is not a behaviour to delete; a fresh kernel is often
what somebody wants. It was a bug because it was *invisible* and *not
optional*. So `kernel.auto-restart` is now a capability, off by default, and a
dead kernel is reported with both ways forward named. Attaching for the first
time is never a restart and is never refused; `restart_notebook` is a caller
asking for exactly this and is never blocked by a switch about doing it behind
their back.

Read back through `server/discover` and `capabilities://`, which report what
exists *and is off* as well as what is on — a client seeing only what is on
cannot tell a capability this server lacks from one it has and is not using,
and that decides whether asking an operator is worth it. Every value says
where it came from, which is the first question when a capability surprises
somebody. A misspelt `--capability` stops the server rather than being
dropped.

**`cell_id` on every cell tool.** An index is a position, and a position stops
being true the moment anyone inserts a cell above it — constantly, in a
notebook a person and an agent share. The agent reads cell 3, decides to fix
it, and by the time the edit lands cell 3 is a different cell. Nothing errors.
The wrong cell is quietly overwritten. Every result now says which cell id it
acted on, and every cell tool accepts one. An id that is gone is an error
naming it, never a fall back to the index it arrived with.

**Tool annotations.** All eighteen tools answer `idempotentHint` and
`openWorldHint`. The direction that matters is the unsafe one:
`idempotentHint: true` on `insert_cell` or `delete_cell` would invite a client
to retry a timed-out call and insert the cell twice, or delete a second,
different cell.

## Contracts that were being decided by accident

**Extensions register after configuration, not at import.** The CLI imports
the server module in order to start it, so extensions registered while the
command line was still being parsed. An extension asking what the server was
pointed at got the default however it had been invoked — which is why
`jupyter_mcp_spaces` had to read `sys.argv` for itself.

**Extensions register in name order.** `importlib.metadata` returns entry
points in whatever order the installation produced; it varies between a wheel
and an editable install. Registration is not independent — an extension may
replace a tool another registered, and the SDK keeps the *original* when a
name is registered twice — so a replacement that happened to run first
silently did nothing.

**Deterministic tool order.** `list_tools` sorts by name, so a client caching
the list stops seeing changes that are not changes.

## One place for the shape of a result

`results.py`. A `tools/call` may answer with `content` and with
`structuredContent`, and a server cannot know which a client puts in front of
the model — which is why the Core Primitives WG is redesigning that contract,
and why content annotations may be deprecated in the same pass. When either
lands, one file changes and eighteen tools do not. Cache hints (SEP-2549) ride
the same helper, on the three listings only: a cell read or an edit answers
something that has just moved, and a stale hint is worse than none.

## An audit seam

`JUPYTER_MCP_AUDIT_SINK_CLASS`. This server has no opinion about auditing and
should not: a laptop needs none, a hosted deployment needs a durable queryable
record, a bank needs it in a SIEM. A sink that fails never fails the call it
was describing, and is *wrapped* rather than trusted so one declaring
`propagate_errors` cannot opt back in. It logs at `ERROR` every time, because
a silently dropped audit record looks exactly like nothing having happened. A
name that cannot be loaded stops the server: an operator who configured
auditing and got a server running without it has the worst of both.

Also: mcp 2 installs `OpenTelemetryMiddleware` by default, so this server
passes none. Passing one appends a second instance and every message is traced
twice.

## The specification's own test suite

Every other test in this repository was written by us, so every one encodes
what we *believe* the protocol says. `@modelcontextprotocol/conformance` is
written by the people who define it.

Real result at 2025-11-25: **9 passed, 21 failed** — and reading them matters.
Almost all are the suite's own scenario *fixtures* rather than protocol
violations: `tools-call-mixed-content` fails with "Expected multiple content
items; Missing image content" because it looks for a tool that returns those,
and this server's tools return notebook results. So it is adopted with a
baseline naming each one, and a *new* failure blocks the merge. A test refuses
a wildcard in the baseline, and a scenario that starts passing surfaces as a
stale entry to delete.

Two things the first run got wrong, now guarded. Pointed at the authenticated
fixture every scenario failed including `ping` — which measures that
authentication works, true and not what conformance means. And a scenario that
ran *no checks at all* was counted as a pass, while the runner counted it as a
failure.

## Documentation

`operations/audit` and `operations/capabilities`, plus a test suite holding all
three feature pages to the code — down to the `ERROR` line the audit page tells
operators to alert on, checked against the line actually logged. A page naming
a variable the server no longer reads is worse than no page: somebody follows
it, gets no error, and concludes the feature is broken.

## Two things deliberately not in this branch

Both need a push from a token with the `workflow` scope, which this checkout
has not got:

- **The conformance CI job.** The suite is wired up as a test and a
  `make test-conformance` target; only the job that runs it on every pull
  request is missing.
- **Removing `ext/spaces`.** Its tools have moved to Datalayer's hosted
  server, but the workflow does `pip install ./ext/spaces`, so deleting the
  directory without editing the workflow leaves CI installing something that
  is not there. The deletion and the workflow edit are inseparable, so both
  wait.

## Verification

259 tests pass across the unit suites and both extensions. Every new guard was
mutation-tested; the ones that survived a first pass got better tests rather
than being written off — three of them had been asserting on docstrings that
explained a rule rather than on the code implementing it.


echarles (commented):
LGTM

PR #95, opened 2025-10-05
Feat: Run as Jupyter Server Extension

echarles:

PR #411, opened 2026-08-27
fix(cell-source): report no changes only when the sources are equal

### LINKED ISSUE #410: edit_cell_source and overwrite_cell_source report "no changes detected" for an edit they did apply
## Summary

`edit_cell_source` and `overwrite_cell_source` write the new source to the cell and then report `no changes detected`, whenever the edit is confined to line terminators. The cell on disk is rewritten; the message the caller gets back says nothing happened.

## Environment

- HEAD `87ed23211d7541f750c5a526d8867cacd49b4f4d` (`fix: support raw cell insertion (#408)`)
- `python:3.12-slim` container, `pip install .` from a clean checkout
- Import provenance in the run below: `/usr/local/lib/python3.12/site-packages/jupyter_mcp_server/__init__.py`

## Repro

```python
import asyncio, os, tempfile
import nbformat
from jupyter_mcp_server.tools.edit_cell_source_tool import EditCellSourceTool

tool = EditCellSourceTool()

nb = nbformat.v4.new_notebook()
nb.cells = [nbformat.v4.new_code_cell("a = 1\r\nb = 2")]
path = os.path.join(tempfile.mkdtemp(), "repro.ipynb")
with open(path, "w", encoding="utf-8") as f:
    nbformat.write(nb, f)

before = nbformat.read(path, as_version=4).cells[0].source
diff = asyncio.run(tool._edit_cell_file(path, 0, "\r\n", "\n", False))
after = nbformat.read(path, as_version=4).cells[0].source

print("before :", repr(before))
print("after  :", repr(after))
print("changed:", before != after)
print("report :", diff)
```

Output:

```
before : 'a = 1\r\nb = 2'
after  : 'a = 1\nb = 2'
changed: True
report : no changes detected
```

`execute()` turns that report into `Cell 0 edited successfully - no changes detected`, which is what an agent driving the tool reads.

## Expected vs actual

Expected: `no changes detected` only when the new source equals the old one.
Actual: it is also returned for an applied edit, so a caller cannot tell the two apart.

Input shapes that trigger it, all confirmed on the HEAD above:

| old source | new source | reported |
|---|---|---|
| `'a = 1\r\nb = 2'` | `'a = 1\nb = 2'` | `no changes detected` |
| `'a = 1\x0cb = 2'` | `'a = 1\nb = 2'` | `no changes detected` |
| `'a = 1\x0bb = 2'` | `'a = 1\nb = 2'` | `no changes detected` |
| `'a = 1\u2028b = 2'` | `'a = 1\nb = 2'` | `no changes detected` |
| `'a = 1\n'` | `'a = 1'` | `no changes detected` |
| `'a = 1'` | `'a = 1\n'` | `no changes detected` |

An ordinary edit (`'a = 1\nb = 2'` to `'a = 2\nb = 2'`) renders its unified diff as usual, so this is confined to the terminator cases.

## Where it appears to come from

`edit_cell_source_tool.py:57-73` and `overwrite_cell_source_tool.py:27-43` carry the same `_generate_diff`. It decides whether anything changed from the size of the line diff:

```python
old_lines = old_source.splitlines(keepends=False)
new_lines = new_source.splitlines(keepends=False)
diff_lines = list[str](difflib.unified_diff(old_lines, new_lines, lineterm="", n=3))
if len(diff_lines) > 3:
    return "\n".join(diff_lines)
return "no changes detected"
```

`str.splitlines()` drops the final line terminator and treats `\r\n`, `\f`, `\v`, `\x1c` to `\x1e`, `\x85`, `\u2028` and `\u2029` as line breaks, so for the pairs above both sides split identically and `unified_diff` returns an empty list.

The `> 3` threshold looks vestigial rather than intended as a change test. It arrived in `31ddbcf` (#95) next to `return '\n'.join(diff_lines[3:])`, where 3 was the number of unified-diff header lines being stripped. `1ea92f2` (#139) dropped the slice and kept the threshold.

## Fix

A PR follows this issue: decide the change from `old_source == new_source`, and when the sources differ but the line diff is empty, show the escaped forms instead, since that is the only rendering the change is visible in. Both tools are fixed together because they carry the same function.

ebarkhordar:
Closes #410.

## The problem

`edit_cell_source` and `overwrite_cell_source` carry the same `_generate_diff`, which decides whether anything changed from the size of the line diff:

```python
old_lines = old_source.splitlines(keepends=False)
new_lines = new_source.splitlines(keepends=False)
diff_lines = list[str](difflib.unified_diff(old_lines, new_lines, lineterm="", n=3))
if len(diff_lines) > 3:
    return "\n".join(diff_lines)
return "no changes detected"
```

`str.splitlines()` drops the final line terminator and treats `\r\n`, `\f`, `\v`, `\x1c` to `\x1e`, `\x85`, `\u2028` and `\u2029` as line breaks. For an edit confined to those characters both sides split identically, `unified_diff` returns an empty list, and the tool reports `no changes detected` for a cell it has just rewritten.

The `> 3` threshold looks vestigial rather than intended as a change test: it arrived in `31ddbcf` (#95) beside `return '\n'.join(diff_lines[3:])`, where 3 was the number of unified-diff header lines being stripped, and `1ea92f2` (#139) dropped the slice while keeping the threshold.

## The change

Decide the change from `old_source == new_source`, which is the property the report is making a claim about. When the sources differ but the line diff is empty, show the escaped forms rather than the empty diff, since that is the only rendering the change is visible in:

```
@@ line terminators changed @@
-'a = 1\r\nb = 2'
+'a = 1\nb = 2'
```

Ordinary edits keep their existing unified-diff rendering, and equal sources still report `no changes detected`. Both tools are changed together because they carry the same function.

## Verification

Run in a clean `python:3.12-slim` container against this branch and against its base, `pip install .` from the checkout each time.

Base `87ed23211d7541f750c5a526d8867cacd49b4f4d`, new test file only:

```
24 failed, 8 passed in 3.24s
```

This branch, same file:

```
32 passed in 2.55s
```

The new tests are pure unit tests over `_generate_diff`, parametrized across both tool classes and across the six input shapes, so each tool is held to the same contract.

No regression in the suites that exercise these tools. `tests/test_edit_cell_source.py` and `tests/test_cell_tools_explicit_notebook_target.py` report `33 passed, 24 errors` on the base commit and `33 passed, 24 errors` on this branch. The 24 errors are the integration tests that need a live Jupyter and MCP server, which that container does not start; they are identical before and after and are not touched by this change.

I have not run the integration legs against a live server, so the claim above is about the unit-level contract and the absence of a regression in what the container can run.

### DIFF
--- a/jupyter_mcp_server/tools/edit_cell_source_tool.py
+++ b/jupyter_mcp_server/tools/edit_cell_source_tool.py
@@ -55,7 +55,19 @@ class EditCellSourceTool(BaseTool):
         return source.replace(old_string, new_string, 1)
 
     def _generate_diff(self, old_source: str, new_source: str) -> str:
-        """Generate unified diff between old and new source."""
+        """Generate unified diff between old and new source.
+
+        Whether anything changed is decided by comparing the two sources, not by
+        the size of the line diff. str.splitlines() drops the final line
+        terminator and treats "\\r\\n", "\\f", "\\v", "\\x1c"-"\\x1e", "\\x85",
+        "\\u2028" and "\\u2029" as line breaks, so an edit confined to those
+        characters leaves both sides splitting identically and unified_diff
+        returns nothing. Reporting that as "no changes detected" tells the caller
+        the cell was untouched when it has in fact been rewritten.
+        """
+        if old_source == new_source:
+            return "no changes detected"
+
         old_lines = old_source.splitlines(keepends=False)
         new_lines = new_source.splitlines(keepends=False)
 
@@ -68,9 +80,12 @@ class EditCellSourceTool(BaseTool):
             )
         )
 
-        if len(diff_lines) > 3:
-            return "\n".join(diff_lines)
-        return "no changes detected"
+        if not diff_lines:
+            # The sources differ only in characters splitlines() consumes, so
+            # there is nothing to show line by line. Show the escaped forms
+            # instead, which is the only rendering the change is visible in.
+            return f"@@ line terminators changed @@\n-{old_source!r}\n+{new_source!r}"
+        return "\n".join(diff_lines)
 
     def _edit_source(
         self, old_source: str, old_string: str, new_string: str, replace_all: bool
--- a/jupyter_mcp_server/tools/overwrite_cell_source_tool.py
+++ b/jupyter_mcp_server/tools/overwrite_cell_source_tool.py
@@ -25,7 +25,19 @@ class OverwriteCellSourceTool(BaseTool):
     """Tool to overwrite the source of an existing cell."""
 
     def _generate_diff(self, old_source: str, new_source: str) -> str:
-        """Generate unified diff between old and new source."""
+        """Generate unified diff between old and new source.
+
+        Whether anything changed is decided by comparing the two sources, not by
+        the size of the line diff. str.splitlines() drops the final line
+        terminator and treats "\\r\\n", "\\f", "\\v", "\\x1c"-"\\x1e", "\\x85",
+        "\\u2028" and "\\u2029" as line breaks, so an edit confined to those
+        characters leaves both sides splitting identically and unified_diff
+        returns nothing. Reporting that as "no changes detected" tells the caller
+        the cell was untouched when it has in fact been rewritten.
+        """
+        if old_source == new_source:
+            return "no changes detected"
+
         old_lines = old_source.splitlines(keepends=False)
         new_lines = new_source.splitlines(keepends=False)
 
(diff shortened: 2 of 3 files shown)


echarles:
Closing/Reopening to relaunch CI

echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks @echarles, and for bouncing CI to get the suite running. Good to have edits that only touch line terminators reporting as real changes now instead of coming back as no changes detected.

PR #413, opened 2026-08-29
ext/ becomes extensions/, and jupyter-mcp-sandboxes is 0.2.0

echarles:
The folder is called `extensions` now, which is what everything else already called it: the entry-point group is
`jupyter_mcp_server.extensions`, the module that loads them is `extensions.py`, the docs page is Extensions, and the make target is `test-extensions`. Only the directory was abbreviated.

Every reference moved with it — 21 files: the four workflows, `lint.sh`, the Makefile, `pyproject.toml`'s per-file lint rules, the version-sync script and its test, the sourcemap generator and the generated sourcemap, the architecture and configuration pages, four generated tool pages, and both READMEs. Two of them built the path as `"ext"` segments rather than a literal, which a search for `ext/` does not find.

One reference deliberately stays: the release note about `pip install ./ext/spaces` going away. That describes a release in which the folder *was* `ext`, and rewriting it would make the note claim a path that never existed.

`jupyter-mcp-sandboxes` goes to 0.2.0 — a minor rather than a patch, because anybody with an editable install has to reinstall from the new path. The gateway already floors at `>=0.2.0`; `datalayer` still says `>=0.1.4`, which resolves fine.

One thing found on the way: `tests/test_mcpb_version_sync.py` has never run. It imports `scripts.sync_mcpb_version` and `scripts/` had no `__init__.py`, so it failed at collection — the suite was green and the version-drift it guards was unguarded. Three tests now pass, and the check they run is what catches the mcpb manifest drifting from `__version__.py`.

264 tests pass, including all 60 of the sandboxes suite from the new path, and the entry point still resolves.


echarles (commented):
LGTM

PR #417, opened 2026-08-31
fix(hooks): fire AFTER_EXECUTE when a local execution is cancelled

serhiizghama:
`execute_code_local` fires BEFORE_EXECUTE and then lets a `CancelledError` walk straight past its `except Exception` handler, so an execution the client cancelled never gets its AFTER_EXECUTE. A handler that stashed per-execution state in the context never gets it back: `OTelHookHandler` ends its span only in the AFTER branch and `SimpleSpanProcessor` exports on end, so a cancelled `execute_code` is missing from the span file altogether, and an audit sink pairing on the context leaks the entry.

#361 went after exactly this and its message names cancellation and both functions, but only `execute_via_execution_stack` got the branch — `execute_code_local` got the "every exit past that point owes exactly one AFTER_EXECUTE" comment and no handling to match. Added the same branch there. It's guarded on `hook_ctx` because a cancel can also land on the channel-startup sleep, before BEFORE_EXECUTE has fired, and that exit owes nothing. No `after_execute_fired` flag needed here, unlike the sibling, since `CancelledError` can't fall through into `except Exception`.

The test mirrors `test_user_cancellation_fires_after_execute` from the stack path and reuses the zmq fakes already in that file; on main it fails with `expected 1 AFTER_EXECUTE, got 0`.

One thing I left alone on purpose: this path never interrupts the kernel the way the stack path calls `execution_stack.cancel()`, so the code keeps running after the client has given up. Felt like a separate change rather than something to bundle in here.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -1602,6 +1602,22 @@ async def execute_code_local(
         )
         return result
 
+    except asyncio.CancelledError as interrupt_err:
+        # CancelledError is not an Exception, so the handler below never sees
+        # it — the AFTER this exit owes has to be fired here.
+        logger.warning(f"Code execution on kernel {kernel_id} was cancelled")
+        if hook_ctx is not None:
+            await HookRegistry.get_instance().fire(
+                HookEvent.AFTER_EXECUTE,
+                code=code,
+                kernel_id=kernel_id,
+                metadata={},
+                outputs=[],
+                error=interrupt_err,
+                context=hook_ctx,
+            )
+        raise
+
     except Exception as e:
         logger.error(f"Error executing code locally: {e}")
         if hook_ctx is not None:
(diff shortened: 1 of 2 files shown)


serhiizghama:
The red `test (ubuntu-latest, 3.13)` here isn't this change — it's the `TestPasswordAuthE2E` YDoc lag that `_read_cell_until` already documents: the poll times out under CI load and the assertion sees `execution count: N/A`. Same failure shape hit `test_password_auth_create_notebook` on `feat/mcp-2` before this branch existed (run 33106191253) and `main` itself on 21 Aug (run 32459586958). This diff only adds an `except asyncio.CancelledError` arm to `execute_code_local`, which the auth suite never goes through — it runs in MCP_SERVER mode over the ExecutionStack.

`enforce-label` also needs a label I can't set from a fork. Happy to rebase or rerun if that's easier.

echarles (approved):
LGTM Thx @serhiizghama

PR #419, opened 2026-09-01
fix(execute_cell): rebind the replacement kernel under the current notebook name

serhiizghama:
When `execute_cell` replaces a culled kernel, `_start_and_bind_kernel` registers it in the notebook manager keyed by the notebook *path*, but `get_current_notebook_context()` looks the kernel up by the current notebook *name*. In extension mode the auto-enrolled entry is named `default` with path `notebook.ipynb`, so the two never meet: the replacement lands in a second, unreferenced entry and the stale binding stays where the lookup reads it. Every following `execute_cell` then decides the kernel is gone and starts yet another one, so cell state doesn't survive between calls (`x = 1`, then `print(x)` raises `NameError`), kernels pile up one per call, and `list_notebooks` grows a phantom notebook.

Keyed it by the current notebook name instead and kept that entry's recorded path — the same thing `restart_notebook_tool._reprovision_kernel` already does. With an empty manager it still falls back to the path, so the start-on-demand path is unchanged.

The existing coverage in `test_execute_cell_dead_kernel_retry.py` can't see this: it passes an empty `NotebookManager` and stubs out `get_current_notebook_context`. The new tests use both for real and fail on main — one on a second kernel being started, one on the phantom entry.

### DIFF
--- a/jupyter_mcp_server/tools/execute_cell_tool.py
+++ b/jupyter_mcp_server/tools/execute_cell_tool.py
@@ -93,14 +93,17 @@ class ExecuteCellTool(BaseTool):
 
         if notebook_manager is not None:
             kernel_info = {"id": kernel_id}
+            # The context lookup resolves the kernel by notebook name, so a
+            # path-keyed entry would leave the replaced binding in place.
+            name = notebook_manager.get_current_notebook() or notebook_path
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @serhiizghama

PR #421, opened 2026-09-01
fix(capabilities): apply the environment before publishing the registry

### LINKED ISSUE #420: An unknown name in JUPYTER_MCP_CAPABILITIES is refused once, then the half-applied registry is cached and the setting is lost
`JUPYTER_MCP_CAPABILITIES` is read lazily, and `get_capabilities()` (capabilities.py:231) assigns the module global before it applies the environment:

```python
if _registry is None:
    _registry = CapabilityRegistry()
    _registry.apply(from_environment(), source="config")
return _registry
```

If any name in the variable is misspelt, `apply` raises `UnknownCapability` after `_registry` is already assigned. The first caller sees the error. Every caller after it gets the half-applied registry back with no error at all, so the setting is silently dropped, which is the outcome `UnknownCapability` exists to prevent.

`--capability` does not have this problem: `serve.py` applies the flags before the server starts and an unknown name stops it. The environment path is only ever touched from the request path (`capabilities_resource`, and `capabilities_enabled` inside `ensure_code_sandbox_alive`), so the operator does not learn about it at startup either.

Reproduced on 5e2e775 in a clean `python:3.11-slim` container with `pip install -e .`:

```python
import os
os.environ["JUPYTER_MCP_CAPABILITIES"] = "kernel.autorestart,kernel.auto-restart"
from jupyter_mcp_server.capabilities import KERNEL_AUTO_RESTART, enabled, get_capabilities
try:
    print("call 1:", enabled(KERNEL_AUTO_RESTART))
except Exception as e:
    print("call 1 raised:", type(e).__name__, e)
print("call 2:", enabled(KERNEL_AUTO_RESTART))
print("advertised:", get_capabilities().advertise()["capabilities"])
print("source:", get_capabilities().get(KERNEL_AUTO_RESTART).source)
```

Output:

```
call 1 raised: UnknownCapability 'kernel.autorestart' is not a capability of this server. Known: kernel.auto-restart
call 2: False
advertised: []
source: default
```

Expected: `kernel.auto-restart` was named in the variable, so either it is on, or every call keeps refusing. Actual: it is off, `source` reads `default`, `capabilities://` advertises nothing, and nothing after the first call says why.

Which settings survive depends on where the bad name sits. Reversing the order to `kernel.auto-restart,kernel.autorestart` makes call 2 return `True`: entries before the bad one are applied and everything after it is lost.

The consequence is the behaviour the capability governs. With `kernel.auto-restart` believed on and actually off, `ensure_code_sandbox_alive` raises `KernelGoneError` instead of starting the replacement, which is the opposite of what the operator configured.

Version: current main, 5e2e775 (2.1.3 plus #417).

ebarkhordar:
`get_capabilities()` assigned `_registry` before it applied `JUPYTER_MCP_CAPABILITIES`, so when `apply` raised `UnknownCapability` the half-applied registry was already cached. The first caller saw the error and every caller after it read that registry back with no error at all, which turns a refused name into a dropped one. Which settings survived depended on where the bad name sat in the list.

The fix builds the registry in a local and assigns the global only after `apply` returns, so a name nobody declared is refused on every call. That is what `--capability` already does through `serve.py`, and what `UnknownCapability` was written for.

Verified in a clean `python:3.11-slim` container on 5e2e775:

- `tests/test_capabilities.py::TestSettings::test_a_misspelt_name_in_the_environment_is_refused_on_every_call` and `::test_the_environment_is_never_applied_in_part` fail before the change and pass after. The second covers the order dependence, since a bad name placed after a good one used to leave the good one applied.
- `pytest tests/test_capabilities.py` goes from 32 passed to 34 passed.
- Whole suite, `-m "not integration"`: 693 passed before, 695 after, with the same 2 failures and 129 errors on both sides. Those are the live JupyterLab fixture failing to start in the container, not this change.
- `ruff check` on both touched files reports only the pre-existing `N818` on the `UnknownCapability` class name.

Not checked: whether raising on every call rather than once is the behaviour you want on the request path. The alternative is to apply the names that are valid and log the bad one, which reads more forgiving but goes against what the class docstring and the `serve.py` comment both say. Happy to switch it if you prefer that.

Fixes #420

### DIFF
--- a/jupyter_mcp_server/capabilities.py
+++ b/jupyter_mcp_server/capabilities.py
@@ -229,11 +229,17 @@ _registry: CapabilityRegistry | None = None
 
 
 def get_capabilities() -> CapabilityRegistry:
-    """The registry of this process, built on first use."""
+    """The registry of this process, built on first use.
+
+    Published only once the environment has been applied. Assigning first
+    leaves a registry the bad entry stopped halfway through, and every call
+    after the one that raised reads it back as though it were configured.
+    """
     global _registry
     if _registry is None:
-        _registry = CapabilityRegistry()
-        _registry.apply(from_environment(), source="config")
+        registry = CapabilityRegistry()
+        registry.apply(from_environment(), source="config")
+        _registry = registry
     return _registry
 
 
--- a/tests/test_capabilities.py
+++ b/tests/test_capabilities.py
@@ -32,6 +32,7 @@ from jupyter_mcp_server.capabilities import (
     Capability,
     CapabilityRegistry,
     UnknownCapability,
+    enabled,
     get_capabilities,
     parse_setting,
     reset_capabilities,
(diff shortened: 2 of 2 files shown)


echarles:
Haven't look if we have duplicates, if such, can you make it unique?

https://github.com/datalayer/jupyter-mcp-server/issues/420#issuecomment-5494444958

ebarkhordar:
There is only one. `kernel.autorestart` in that line is not a second capability, it is a deliberate misspelling: the repro sets `JUPYTER_MCP_CAPABILITIES` to one good name and one bad one so `apply` raises partway through the list, which is what leaves the half-applied registry in the cache.

`BUILT_IN` declares a single entry, `kernel.auto-restart` (capabilities.py:62). The unhyphenated spelling occurs three times in the tree and each one is an example of a name nobody declared: the `UnknownCapability` docstring, the capabilities doc page, and `tests/test_capabilities.py:104`, which asserts the raise.

Sorry it read as a duplicate, that was a poor string to illustrate with. Happy to reword the issue to use something obviously fake instead.

echarles:
no worries, sounds goods

echarles (approved):
LGTM Thx @ebarkhordar

ebarkhordar:
Thanks for the quick merge. I will pick an obviously fake name for the next repro so it does not read as a second capability.

PR #422, opened 2026-09-01
enh: tasks cancellation + resources (notebook, cells, outputs)

echarles:
This PR enhances the tasks subsystem so that (1) cancelling a task running a cell interrupts the kernel (stopping the cell, not just the wait loop) and (2) a cancelled (or failed) task can preserve what it produced before termination by recording partial outputs during execution.

**Changes:**
- Add `TaskRecord.partial` plus a `record_output()` helper, and promote partial output into `result` on cancellation (and retain it on failure).
- Update the cell execution loop to register the kernel interrupt handler and record outputs as they appear.
- Add/extend tests and documentation to lock the behavior and keep docs/code in sync.
</details>

It also add a bump script to ease the version bump at the needed places.

It adds resources: the notebook, its cells and their outputs, for an agent to read.

Lighten README.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -48,7 +48,7 @@ jobs:
         run: |
           git add -AN docs/sourcey
           git diff --exit-code -- docs/sourcey \
-            || (echo "::error::docs/sourcey is stale; re-run the generation (docs/sourcey/README.md) and commit the result" && exit 1)
+            || (echo "::error::docs/sourcey is stale; run 'make sync-sourcey' and commit the result (see docs/sourcey/README.md)" && exit 1)
 
       - name: Build the docs
         working-directory: docs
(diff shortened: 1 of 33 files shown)


ebarkhordar (commented):
`execute_cell(stream=True)` runs a second monitor loop at `execute_cell_tool.py:492` that calls neither `register_interrupt` nor `record_output`, so cancelling there leaves the kernel running and `result` empty. That is the documented mode for long-running cells.

Both paths at `b6c494a0`, clean container, same fake kernel, a cell printing twice:

```
non-streaming: interrupt True,  partial 2
streaming:     interrupt False, partial 0
```

That loop already tracks `last_output_count` and interrupts on timeout, so the two lines fit. `_loop_source()` reads only `utils`, so the new tests cannot see it.

echarles:
Thx for the review @ebarkhordar d12ad39717a97da4c692949e7834c0eac3761b3a fixes your report

ebarkhordar:
Re-ran the same differential at 17e9085: the streaming loop now registers the interrupt and records both partial outputs, matching the non-streaming one. Thanks for picking it up.

echarles (commented):
LGTM

PR #424, opened 2026-09-02
feat: sandboxes snapshot

echarles:
The current snapshot_name handling is truthiness-based, which can silently ignore empty/whitespace snapshot requests and undermine the “refuse rather than start empty” behavior guarantee.

Once you've addressed the issues Copilot identified, you can request another Copilot review.

Pull request overview
This PR enhances the sandboxes extension to support starting a sandbox launch from a named snapshot (for the datalayer variant only) and improves extension composition by adding public accessors for looking up registered extensions and sharing the sandboxes registry, while adjusting tests to avoid making environment-dependent assertions about third-party extensions.

Changes:

Thread snapshot_name from the launch_sandbox tool through to CodeSandboxManager.launch, and explicitly refuse snapshot launches for non-datalayer variants.
Add ExtensionManager.get(name) and SandboxesExtension.sandboxes to support composing extensions without reaching into private attributes.
Make the notebook resource scheme test assert only the core notebook/cell/output templates, not the full registry (so extensions can add their own schemes).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,6 +57,12 @@ To get started with development, you'll need to set up your environment.
    make test
    ```
 
+1. **Sync sourcey:**
+
+   ```bash
+   make sync-sourcey
+   ```
+
 1. **Build Python Package/Docker Image:**
 
    ```bash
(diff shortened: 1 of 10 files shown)


echarles (commented):
LGTM

PR #427, opened 2026-09-05
fix(outputs): strip every CSI escape, not only colour

### LINKED ISSUE #426: strip_ansi_codes only removes colour codes, so cursor escapes reach the client
`strip_ansi_codes` in `jupyter_mcp_server/utils.py` is documented as removing ANSI escape sequences, but the pattern is `\x1b\[[0-9;]*m`, which is only the colour ones. Anything else the kernel wrote comes back in the text `extract_output` returns.

The case I ran into is nested tqdm bars. tqdm moves the cursor up with `ESC[A` on each redraw and does that whether or not the stream is a terminal, so an epoch bar plus a batch bar gives stream output with `ESC[A` scattered through it:

```python
import io
from tqdm import tqdm
from jupyter_mcp_server.utils import extract_output

buf = io.StringIO()
for i in tqdm(range(2), file=buf, position=0, ncols=30, mininterval=0):
    for j in tqdm(range(2), file=buf, position=1, ncols=30, mininterval=0, leave=False):
        pass

print(repr(extract_output({"output_type": "stream", "name": "stderr", "text": buf.getvalue()})))
```

On 2.1.5 that prints the bar text with `\x1b[A` still in it. Erase-line (`ESC[2K`) and the cursor-hide pair (`ESC[?25l` and `ESC[?25h`) survive the same way, so anything driving a live display leaks control bytes into the tool result.

ebarkhordar:
Fixes #426.

`strip_ansi_codes` matched `\x1b\[[0-9;]*m`, so it only took out colour. A nested tqdm bar moves the cursor with `ESC[A` on every redraw whether or not the stream is a terminal, and those bytes came back in the stream output; erase-line and the cursor-hide pair did the same.

Widened the pattern to the CSI form, so `ESC[`, parameter bytes, intermediate bytes, one final byte. That is a superset of what it matched before, and text that merely contains something like `[0-9]*` is untouched because the `ESC` is required.

I left OSC alone, so the `ESC]8;;` hyperlink form that rich emits still comes through. I have no case of one reaching a cell output, and guessing at where the string terminator lands seemed worse than leaving it.

Tests cover colour, cursor movement, erase-line, the cursor-hide pair, and a captured nested-tqdm redraw through `extract_output`. The non-colour ones fail on main. The suite's live-server tests do not start for me (jupyter lab will not run as root in my container), so those are unrun on both sides.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -540,7 +540,7 @@ def extract_output(output: dict | Any) -> str | ImageContent:
 
 def strip_ansi_codes(text: str) -> str:
     """Remove ANSI escape sequences from text."""
-    ansi_escape = re.compile(r"\x1b\[[0-9;]*m")
+    ansi_escape = re.compile(r"\x1b\[[0-?]*[ -/]*[@-~]")
     return ansi_escape.sub("", text)
 
 
--- a/tests/test_ansi_escape_stripping.py
+++ b/tests/test_ansi_escape_stripping.py
@@ -1,0 +1,41 @@
+# Copyright (c) 2024- Datalayer, Inc.
+#
+# BSD 3-Clause License
+
+"""Unit tests for stripping ANSI escape sequences out of cell output.
+
+strip_ansi_codes matched only the SGR colour sequences, so every other escape
+a kernel writes survived into the text handed back to the client. A nested
+tqdm bar moves the cursor with ESC[A on each redraw whether or not the stream
+is a terminal, and those bytes came through in the stream output.
+"""
+
+from jupyter_mcp_server.utils import extract_output, strip_ansi_codes
+
+# Two redraws of a nested tqdm bar, captured from tqdm writing to a
+# non-terminal stream with position=1.
+NESTED_TQDM_TEXT = (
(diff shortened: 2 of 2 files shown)


ebarkhordar:
`enforce-label` wants a triage label and I cannot add one from outside the org, so that leg needs you.

The `test` legs are red on main too, at `da406454`: `extensions/sandboxes/tests/test_attach_by_name.py` has three `async def` tests that nothing runs them with, so they fail with "async def functions are not natively supported". That run is from before this branch. Not from this change, but I would rather say so than have it read as mine.

echarles:
Label is set and failing test prior this PR are fixed with https://github.com/datalayer/jupyter-mcp-server/pull/428

Please rebase to confirm green CI :+1:

ebarkhordar:
Rebased onto 89a4a02, so this now sits on top of #428. Thanks for picking up the label and the test fix.

For the record on the old base da40645, the six red `test` legs there were failing identically on main, so none of them were coming from this diff. Let us see what the fresh run says.

echarles (approved):
LGTM Thx @ebarkhordar

PR #159, opened 2025-11-03
Prompt/add cite component

ChengJiale150:
## Summary

A init implement of #153 and bump it to 0.19.0 . It introduce a powerful and flexible prompt tool **jupyter-cite**, which enable user cite specific cells from specified notebook and also release the stable verison of `prompt/general`.

### `prompt/general`
- enhance description of `AGENT.md`
- update the README

### `jupyter-cite`
- add `jupyte-cite` and `JupyterCitePrompt` class
- update docs: add a new section `prompt`

PR #147, opened 2025-10-25
Init prompt repository

ChengJiale150:
see in #145 

- init `prompt/` with `README.md`
- add `prompt/general/` as general prompt for  a wide variety of use cases
- update `README.md`

PR #431, opened 2026-09-06
feat(extension): serve prompts and resources in JUPYTER_SERVER mode

### LINKED ISSUE #113: `JUPYTER_SERVER` mode should support `Prompts` MCP API

emecii:
A client connected to the JupyterLab-extension deployment is told in the handshake that this server has prompts and resources, and then cannot reach any of them. `jupyter_cite` is invisible, the three `notebook://` templates are invisible, and three of the five methods do not exist at all:

```
POST /mcp  initialize
  -> {"capabilities": {"tools": {}, "prompts": {}, "resources": {}}, "serverInfo": {"version": "2.1.7"}}
POST /mcp  prompts/list             -> {"result": {"prompts": []}}
POST /mcp  resources/list           -> {"result": {"resources": []}}
POST /mcp  resources/templates/list -> {"error": {"code": -32601, "message": "Method not found: resources/templates/list"}}
POST /mcp  prompts/get              -> {"error": {"code": -32601, "message": "Method not found: prompts/get"}}
POST /mcp  resources/read           -> {"error": {"code": -32601, "message": "Method not found: resources/read"}}
```

The standalone server, from the same checkout, serves `jupyter_cite`, `capabilities://` and all three `notebook://` templates.

`jupyter_mcp_server/jupyter_extension/handlers.py:419-426`:

```python
elif method == "prompts/list":
    # List available prompts - return empty list if no prompts defined
    logger.info("Listing prompts...")
    response = {"jsonrpc": "2.0", "id": request_id, "result": {"prompts": []}}
elif method == "resources/list":
    # List available resources - return empty list if no resources defined
    logger.info("Listing resources...")
    response = {"jsonrpc": "2.0", "id": request_id, "result": {"resources": []}}
```

Those two branches and the `capabilities` dict above them arrived together in `3a4e06d` ("fix ressources and prompts list", #117), which closed #113. At that date the comments were true — the server had no prompts and no resources, so `[]` was the honest answer and advertising the capability was reasonable. `jupyter_cite` landed three weeks later (#147, #159, #161) and the notebook resources ten months later (#422). Both registered on the shared `MCPServer` object, neither reached this transport, and the placeholder became an advertisement the extension does not honour.

## What this does

`prompts/list`, `prompts/get`, `resources/list`, `resources/templates/list` and `resources/read` now go to the same `mcp` object the `tools/list` and `tools/call` branches already delegate to. There is no second registry, and no new prompt or resource — the extension serves what the server has.

Each response is built from the SDK's own result model (`ListPromptsResult`, `GetPromptResult`, …) and dumped with the same arguments the `tools/call` branch uses for its `CallToolResult`, so the wire is the one the standalone server produces for the same request. `resources/read` turns `ReadResourceContents` into `TextResourceContents` / `BlobResourceContents` the way the SDK's own `resources/read` handler does, so binary content arrives base64 in a `blob` rather than coerced into `text`, and it maps `ResourceNotFoundError` to `-32602` and every other `ResourceError` to `-32603`, again matching the SDK.

`resources.read_notebook` needed a mode branch. It called `resolve_notebook_connection`, and `NotebookConnection` refuses a local notebook by design — *"This is only used in MCP_SERVER mode […] In JUPYTER_SERVER mode (local), notebook content is accessed directly via contents_manager"*. It now reads the way `ReadNotebookTool` does in that mode: the live YDoc first, `contents_manager.get()` as the fallback. `JupyterCitePrompt` already had its own `ServerMode.JUPYTER_SERVER` branch and needed no change.

**Subscriptions are deliberately out of scope.** `resources/subscribe` and `resources/unsubscribe` are registered on `mcp._lowlevel_server` and driven over the SDK's session bus, which a Tornado request does not have. They stay MCP_SERVER-only; there is a line saying so on the architecture page.

The same five requests after the change:

```
POST /mcp  prompts/list
  -> {"result": {"prompts": [{"name": "jupyter_cite", "arguments": [...]}]}}
POST /mcp  resources/list
  -> {"result": {"resources": [{"uri": "capabilities://", "mimeType": "text/plain", ...}]}}
POST /mcp  resources/templates/list
  -> {"result": {"resourceTemplates": [
       {"name": "notebook", "uriTemplate": "notebook://{name}", ...},
       {"name": "notebook-cell", "uriTemplate": "notebook://{name}/cells/{cell_id}", ...},
       {"name": "notebook-cell-output", "uriTemplate": "notebook://{name}/cells/{cell_id}/outputs/{index}", ...}]}}
POST /mcp  prompts/get  {"name": "jupyter_cite", "arguments": {"cell_indices": "0", ...}}
  -> {"result": {"messages": [{"role": "user", "content": {"type": "text",
       "text": "USER Cite cells [0] from notebook default, ..."}}]}}
POST /mcp  resources/read  {"uri": "capabilities://"}
  -> {"result": {"contents": [{"uri": "capabilities://", "mimeType": "text/plain", "text": "{...}"}]}}
```

## Tests

`tests/test_prompts.py` loses its `pytestmark` skipif — the only mode-gated skip in the suite, whose reason read *"Prompt feature is only available in MCP_SERVER mode now."* — and moves from `mcp_client` to `mcp_client_parametrized`, so the existing `test_jupyter_cite` now runs against both deployments unchanged. A `prompts/list` test is added beside it.

`tests/test_resources_both_modes.py` is new: both transports, the same assertions. It covers the three templates, `resources/list`, reading a notebook back as nbformat JSON, a cell listing its outputs by URI rather than inlining them, reading one of those outputs, and an unopened notebook coming back as a protocol error rather than an empty document. It creates its own notebook for the cell case, because the ones in `dev/content/` predate nbformat 4.5 and have no cell ids to address. `tests/test_notebook_resources.py` stays as the unit test of `resources.py`.

Against the unfixed tree, `TEST_JUPYTER_SERVER=true`:

```
FAILED tests/test_prompts.py::test_the_prompt_is_listed[jupyter_extension]
  AssertionError: assert 'jupyter_cite' in set()
FAILED tests/test_prompts.py::test_jupyter_cite[jupyter_extension]
  MCPError: Method not found: prompts/get
FAILED tests/test_resources_both_modes.py::test_the_three_notebook_templates_are_offered[jupyter_extension]
  MCPError: Method not found: resources/templates/list
FAILED tests/test_resources_both_modes.py::test_a_concrete_resource_is_listed[jupyter_extension]
  AssertionError: assert 'capabilities://' in set()
FAILED tests/test_resources_both_modes.py::test_a_notebook_reads_back_as_nbformat_json[jupyter_extension]
  MCPError: Method not found: resources/read
FAILED tests/test_resources_both_modes.py::test_a_cell_lists_its_outputs_by_uri_rather_than_inlining_them[jupyter_extension]
  MCPError: Method not found: resources/read
FAILED tests/test_resources_both_modes.py::test_a_notebook_nobody_opened_is_refused_rather_than_answered[jupyter_extension]
  MCPError: Method not found: resources/read
7 failed in 4.01s
```

With the fix, both modes:

```
14 passed in 26.74s
```

Full suite:

- `make test-extensions` — 67 passed
- `make test-mcp-server` — 1 failed, 1089 passed, 8 skipped, 4 errors
- `make test-jupyter-server` — 1 failed, 1075 passed, 22 skipped, 4 errors

The one failure and the four errors are `tests/test_use_notebook_ui_open.py` and `tests/test_use_notebook_split_servers.py::test_split_opens_ui_on_document_server`; they fail the same way on `main` at `3aa8cfd` with none of this applied, and touch nothing here.

`ruff check` reports nothing new on the touched files.

### DIFF
--- a/docs/docs/architecture/index.mdx
+++ b/docs/docs/architecture/index.mdx
@@ -356,6 +356,13 @@ never propagates errors into a tool call. See
 1. The `CallToolResult` is serialized in wire form and returned as the JSON-RPC
    response.
 
+`prompts/list`, `prompts/get`, `resources/list`, `resources/templates/list` and
+`resources/read` take the same route: the handler asks the same `MCPServer`
+instance, so the extension serves the prompts and resources the standalone
+server serves. `resources/subscribe` is the exception — it is driven over the
+SDK's own session bus, which a Tornado request does not have, so a client that
+wants notifications about a resource uses the standalone server.
+
 ### `execute_code` with a sandbox variant
 
 1. The sandboxes extension's `intercept_execute_code()` is asked first; when a
--- a/jupyter_mcp_server/jupyter_extension/handlers.py
+++ b/jupyter_mcp_server/jupyter_extension/handlers.py
@@ -9,11 +9,23 @@ This module provides handlers that bridge between Tornado (Jupyter Server) and
 MCPServer, managing the MCP protocol lifecycle and request proxying.
 """
 
+import base64
 import json
 import logging
 from typing import Any
 
 from jupyter_server.base.handlers import JupyterHandler
+from mcp.server.mcpserver.exceptions import ResourceError, ResourceNotFoundError
+from mcp.types import (
+    INTERNAL_ERROR,
+    INVALID_PARAMS,
+    BlobResourceContents,
+    ListPromptsResult,
+    ListResourcesResult,
+    ListResourceTemplatesResult,
+    ReadResourceResult,
+    TextResourceContents,
+)
 from tornado.web import HTTPError
 
 from jupyter_mcp_server.__version__ import __version__
@@ -44,6 +56,30 @@ async def _fetch_jupyter_tools(**kwargs):
     from jupyter_mcp_tools import get_tools
 
     return await get_tools(wait_timeout=5, **kwargs)
+
+
+def _sdk_result(request_id: Any, result: Any) -> dict:
+    """A JSON-RPC response carrying an SDK result object.
+
+    Serialized the way the ``tools/call`` branch already serializes its
+    ``CallToolResult``: camelCase keys, nothing null, and no ``resultType``,
+    which belongs to a newer protocol revision than the one this handler
+    announces. Building the SDK's own result model rather than a hand-rolled
+    dict is what keeps this transport's wire identical to the one the
+    standalone server puts out for the same request.
+    """
+    return {
+        "jsonrpc": "2.0",
+        "id": request_id,
+        "result": result.model_dump(
+            by_alias=True, mode="json", exclude_none=True, exclude={"result_type"}
+        ),
+    }
+
+
+def _error_response(request_id: Any, code: int, message: str) -> dict:
+    """A JSON-RPC error response."""
+    return {"jsonrpc": "2.0", "id": request_id, "error": {"code": code, "message": message}}
 
 
 class MCPSSEHandler(JupyterHandler):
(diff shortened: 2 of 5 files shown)


echarles (approved):
LGTM Thx @emecii

PR #432, opened 2026-09-06
fix: reject overlapping ambiguous cell edits

emecii:
`edit_cell_source` promises a unique `old_string` unless `replace_all=True`. `str.count` misses overlapping matches: `aba` occurs twice in `ababa`, but the tool previously accepted it and silently selected the first occurrence. Repeated notebook lines exhibit the same ambiguity.

Find the first occurrence, then search again from the next character. Reject ambiguous edits before writing the file or live notebook. Explicit `replace_all=True` retains Python's existing non-overlapping replacement behavior; empty, absent and unique matches retain their behavior.

Validation:
- Eight new cases cover ASCII, multiline and Unicode overlaps, unchanged file bytes, unique matches and explicit replacement, including both live server modes. Six fail before the fix; all eight pass after it.
- 102 related tests pass across both modes; 67 extension tests pass.
- The full baseline suite has an unrelated cell-ID/insert readback failure (1,150 passed, one failed, three skipped). A patched full-suite pass is not claimed.
- Ruff/mypy findings match baseline; the new test file passes formatting/lint. Package build and sourcey synchronization pass.

Implemented and checked with OpenAI Codex (GPT-6).

### DIFF
--- a/jupyter_mcp_server/tools/edit_cell_source_tool.py
+++ b/jupyter_mcp_server/tools/edit_cell_source_tool.py
@@ -35,12 +35,15 @@ class EditCellSourceTool(BaseTool):
         if not old_string:
             raise ValueError("old_string must not be empty")
 
-        count = source.count(old_string)
-        if count == 0:
+        first_match = source.find(old_string)
+        if first_match == -1:
             raise ValueError("old_string not found in cell source")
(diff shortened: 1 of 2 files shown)


echarles (approved):
LGTM Thx @emecii

PR #434, opened 2026-09-07
fix(extension): route direct tool calls through MCP server

emecii:
## Summary

- route the documented `/mcp/tools/call` extension endpoint through the shared MCP server
- preserve the authenticated Jupyter identity for direct tool calls
- add an end-to-end regression test for a non-placeholder tool result

## Validation

- `TEST_MCP_SERVER=false TEST_JUPYTER_SERVER=true pytest tests/test_jupyter_extension.py -q`
- `ruff format --check tests/test_jupyter_extension.py`
- `git diff --check`

The repository's full Ruff run reports pre-existing findings in `handlers.py`; this change introduces none of those findings.

### DIFF
--- a/jupyter_mcp_server/jupyter_extension/handlers.py
+++ b/jupyter_mcp_server/jupyter_extension/handlers.py
@@ -741,6 +741,23 @@ class MCPToolsCallHandler(MCPHandler):
     Body: {"tool_name": "...", "arguments": {...}}
     """
 
+    _identity_token = None
+
+    async def prepare(self):
+        """Authenticate the direct route and give its tool call a Jupyter identity."""
+        await super().prepare()
+        if not self.current_user:
+            raise HTTPError(403, "Authentication required")
+        self._identity_token = set_current_identity(
+            identity_from_jupyter_user(self.current_user)
+        )
+
+    def on_finish(self):
+        """Clear the request identity after the direct tool call finishes."""
+        if self._identity_token is not None:
+            reset_current_identity(self._identity_token)
+            self._identity_token = None
+
     async def post(self):
         """Handle tool execution request."""
         try:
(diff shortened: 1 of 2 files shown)


echarles:
@emecii Can you open a linked issue and explain the reason for the changes you propose. Thx :+1:

emecii:
Opened #435 with the endpoint behavior, reason for the change, and focused validation. The PR is linked there.

jupyter_mcp_server/jupyter_extension/handlers.py
```diff
@@ -741,6 +741,23 @@ class MCPToolsCallHandler(MCPHandler):
     Body: {"tool_name": "...", "arguments": {...}}
     """
 
+    _identity_token = None
+
+    async def prepare(self):
```
ebarkhordar: `MCPToolsCallHandler` already inherits both of these from `MCPHandler` (`handlers.py:634` and `:647`), so the identity gets set twice and reset once. `super().prepare()` on line 748 sets it and stores the token in `self._identity_token`; line 751 sets it again and overwrites that attribute, so `on_finish` resets only the second token. The first is never reset and the identity is still set once the request is over, which is the thing `on_finish` is there to prevent.

Driving the real `prepare`/`on_finish` at f576231 in one task, with Jupyter's own auth step stubbed and everything below it untouched:

```
MCPToolsListHandler  (inherits)  during: Identity(username='ehsan')  after on_finish: None
MCPToolsCallHandler  (this PR)   during: Identity(username='ehsan')  after on_finish: Identity(username='ehsan')
```

I think dropping both methods here is enough, the base gives you the same 403 and the balanced set/reset. I did not check whether Tornado reuses the context across requests on a keep-alive connection, so I can't tell you whether it reaches the next request or just stays set for the rest of this one.

emecii: Fixed in 3468870 by removing the duplicate lifecycle methods; the handler now inherits the balanced set/reset. Extension tests: 13 passed.

ebarkhordar: Confirmed at 3468870: both overrides are gone and the handler picks up the base set and reset. I read the diff rather than running it.

echarles (approved):
LGTM Thx  @emecii

PR #437, opened 2026-09-08
fix(cell-source): split a cell's source on newlines only

### LINKED ISSUE #436: Cell source is split with str.splitlines, so a form feed reads back as a line break
`normalize_cell_source` in `jupyter_mcp_server/utils.py` splits a cell's source with `str.splitlines(keepends=True)`. That breaks on `\v`, `\f`, `\x1c`, `\x1d`, `\x1e`, `\x85`, `\u2028` and `\u2029` as well as on `\n`, and none of those ends a line in nbformat.

`read_cell` and `read_notebook` join the pieces back together with `\n`, so a cell holding one of those characters reads back with a line break the notebook does not contain. `get_overview` counts the same pieces, so a one line cell shows up in the brief listing as having a line hidden.

On 2.1.10 (bcfbc6c):

```python
from jupyter_mcp_server.models import Cell

cell = Cell(index=0, cell_type="code", source="text = 'a\fb'", id="c1")
print(repr(cell.get_source("readable")))
print(repr(cell.get_overview()))
```

```
"text = 'a\x0c\nb'"
"text = 'a\x0c...(1 lines hidden)"
```

The cell holds one line and reads back as two. Form feeds turn up in text pasted out of a PDF and as section markers in Python source, and `\x85` and `\u2028` come along with text copied from a word processor. An agent that reads such a cell, edits it and writes it back leaves an extra newline behind in the notebook.

ebarkhordar:
Fixes #436.

`normalize_cell_source` used `str.splitlines`, which also breaks on `\v`, `\f`, `\x1c`, `\x1d`, `\x1e`, `\x85`, `\u2028` and `\u2029`. Those don't end a line in a notebook, so `read_cell` and `read_notebook` joined the pieces back with `\n` and handed back source carrying a line break the cell doesn't have. `get_overview` counted the same pieces, so a one line cell reported a hidden line.

Split on `\n` and put the endings back. Everything else stays as it was, including the trim of the last line's trailing newline.

One thing changed past the split: the non-str fallback at the bottom now goes through the same path, so it trims that last newline the way the str branch always did. That read like an oversight rather than a difference anyone wanted, but say the word and I'll leave it alone.

Added `tests/test_cell_source_line_splitting.py`. The first two cases fail on main, the third pins the ordinary newline behaviour so the trim doesn't drift.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -621,17 +621,17 @@ def normalize_cell_source(source: Any) -> list[str]:
     if isinstance(source, list):
         return [str(line) for line in source]
 
-    # If it's a string, split by newlines
-    if isinstance(source, str):
-        # Split by newlines but preserve the newline characters except for the last line
-        lines = source.splitlines(keepends=True)
-        # Remove trailing newline from the last line if present
-        if lines and lines[-1].endswith("\n"):
-            lines[-1] = lines[-1][:-1]
-        return lines
+    if not isinstance(source, str):
+        source = str(source)
 
-    # Fallback: convert to string and split
-    return str(source).splitlines(keepends=True)
+    # Split on "\n" and nothing else. str.splitlines also breaks on \v, \f,
+    # \x1c-\x1e, \x85, \u2028 and \u2029, none of which ends a line in a
+    # notebook, so a cell holding one came back with a line break in it.
+    lines = source.split("\n")
+    # A trailing "\n" ends the last line, it does not start a new one.
+    if lines[-1] == "":
+        lines.pop()
+    return [line + "\n" for line in lines[:-1]] + lines[-1:]
 
 
 def format_TSV(headers: list[str], rows: list[list[str]]) -> str:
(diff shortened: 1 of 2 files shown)


ebarkhordar:
The red `test (ubuntu-latest, 3.10)` isn't from this branch. It's `TestPasswordAuthE2E::test_password_auth_create_notebook`, and the cell comes back with `execution count: N/A` and the source echoed as its own output, so the kernel never ran it.

The sibling test in the same class fails the same way on main: runs 33948846045 (09-05) and 33975386483 (09-05), both `expected product 335121091 not found in cell` on `test_password_auth_notebook_operations`, same `execution count: N/A` shape.

This PR only touches `normalize_cell_source`, and on that test's input (`31337 * 71993`, one line, no newline) the old `splitlines(keepends=True)` and the new `split('\n')` return the same single-element list, so it can't be what broke it. I haven't worked out what does.

echarles (commented):
LGTM Thx @ebarkhordar

ebarkhordar:
One thing I noticed after: `local_backend.py` does the same normalization at lines 170, 212 and 275 (`source.splitlines(keepends=True)` on append, insert and overwrite), so in JUPYTER_SERVER mode a form feed still turns into a line break on the way into the notebook. That's from reading it, I haven't run that path yet.

I can do those three in a follow-up if you want it.

echarles:
> I can do those three in a follow-up if you want it.

That would be great, thx!

PR #438, opened 2026-09-08
fix(use_notebook): bind the kernel_id passed in MCP_SERVER mode

### LINKED ISSUE #425: use_notebook ignores kernel_id in MCP_SERVER mode and executes on a freshly created kernel (2.1.4)
## Summary

In 2.1.4, MCP_SERVER mode, `use_notebook(kernel_id=<existing id>)` verifies that the kernel exists and then discards the id. Execution goes to a new kernel built from the server configuration. The tool's own parameter description says *"Specific kernel ID to use (will create new if skipped)"*, so a caller reasonably expects the named kernel to be used. It is not.

This breaks the main pair-programming use case: attaching the MCP to the kernel a human already has open in JupyterLab / PyCharm so both sides see the same variables.

JUPYTER_SERVER (extension) mode binds the id correctly; only the MCP_SERVER branch drops it.

## Introduced in

#372 ("enh: use_notebook no longer creates a kernel"), released in **1.3.7**. Bisected against the PyPI wheels:

- **1.3.6**: `create_jupyter_sandbox_client(..., kernel_id=kernel_id, ...)` — the id is passed through.
- **1.3.7**: the call is replaced by `create_kernel(config, logger)` with no id, and the "only checked for existence" comment appears.
- **1.4.0 → 2.1.4**: unchanged apart from the rename to `create_code_sandbox`.

The PR's goal was lazy creation so a non-Jupyter `--sandbox-variant` stops stalling on `/api/status`. Dropping the per-call `kernel_id` was a side effect: it is noted in the new code comment but not in the PR description, and the `kernel_id` parameter and its description were left in place.

## Reproduction

Jupyter Server 2.x on `http://localhost:8889`, password auth, `jupyter-mcp-server==2.1.4` launched with `JUPYTER_URL` + `JUPYTER_PASSWORD`.

1. `list_kernels` — shows a live kernel `afd51e4a-…` with `Connections=1` (the user's notebook kernel).
2. `use_notebook(notebook_name="t", notebook_path="x.ipynb", kernel_id="afd51e4a-…")`
3. Response: `[INFO] Connected to kernel '6f0a6eeb-…'` — a different, newly created kernel.
4. `list_kernels` now shows both. `execute_code("import os; os.getpid()")` runs on `6f0a6eeb`'s process, not `afd51e4a`'s.

With `START_NEW_CODE_SANDBOX=false` the message changes to *"A kernel starts on the first execution"*, and that first execution still creates a new kernel. The requested id is never used in either path.

## Where it happens

`jupyter_mcp_server/tools/use_notebook_tool.py`, MCP_SERVER branch (around lines 412 to 447):

- `kernel_id` is only used to check existence against `sandbox_server_client.kernels.list_kernels()`.
- The code comment states the behaviour explicitly: *"A `kernel_id` given here is only checked for existence: it is not carried to that first execution, which builds a kernel from the configuration."*
- When `start_new_code_sandbox` is set, `create_code_sandbox(config, logger, path=…)` is called with no `kernel_id`.

`jupyter_mcp_server/utils.py:703` — `create_code_sandbox` passes `kernel_id=config.code_sandbox_id`, i.e. the process-wide `--code-sandbox-id` / `CODE_SANDBOX_ID` setting, never the per-call id.

`jupyter_mcp_server/server.py:543` — `__ensure_code_sandbox_alive` (used by `execute_cell`, `insert_execute_code_cell`, …) builds the sandbox through the same config-only path, so the lazy start has no way to receive the id either.

## Why `CODE_SANDBOX_ID` is not a workaround

It is a launch-time setting for the whole server. A user's kernel id changes on every JupyterLab / PyCharm restart, so reusing it means editing the MCP client config and restarting the client each time. `execute_code(kernel_id=…)` can target the kernel but does not write to the notebook, and `execute_cell` has no `kernel_id` parameter.

## Expected

`use_notebook(kernel_id=X)` binds the notebook to kernel X. Later `execute_cell` / `execute_code` on that notebook run on X. If X does not exist, return the existing "not found" error. When `kernel_id` is omitted, keep the current behaviour (lazy or eager creation from config).

## Suggested fix

Store the requested id on the notebook entry in `NotebookManager` at `use_notebook` time, and have `__ensure_code_sandbox_alive` / `create_code_sandbox` prefer that stored id over `config.code_sandbox_id` when creating the sandbox client. This keeps the lazy-creation goal of #372 intact: nothing is started at open time, and the first execution attaches to the named kernel instead of building a fresh one. The JUPYTER_SERVER branch of the same tool already does the equivalent with the local kernel manager and can serve as the reference. Happy to open a PR.

EnyMan:
Fixes #425

## Problem

In MCP_SERVER mode, `use_notebook(kernel_id=X)` checked that X exists, then ignored it. Execution ran on a new kernel built from the configuration. JUPYTER_SERVER mode already binds X. Regressed in #372 (1.3.7).

## Fix

- `use_notebook` with a `kernel_id` attaches to that backend right away. The backend already exists, so nothing is started. #372's goals hold: no runtime is spent on open, and the sandbox variant is honoured.
- `create_code_sandbox` gets a per-call `code_sandbox_id`. It overrides `config.code_sandbox_id`, the field the sandboxes extension already reads, so every sandbox variant sees the id. The extension hook is unchanged.
- The Jupyter kernel-list existence check runs only for the `jupyter-server` variant.
- Without `kernel_id`, behaviour is unchanged (eager or lazy from config).
- Tool description now says "kernel (or sandbox)".

## Tests

`tests/test_use_notebook_kernel_id.py`, 5 tests:

- requested kernel is bound, eager and lazy config
- no `kernel_id` keeps the lazy path
- unknown `kernel_id` is rejected before any sandbox is built
- for another variant, the id reaches the extension hook and no Jupyter client is built

Related suites (`test_code_sandbox_notebook_cwd`, `test_use_notebook_already_connected`, `test_use_notebook_reconnect_interval`, `test_ensure_code_sandbox_alive_server_liveness`, `test_cli_typer_config`) pass. Ruff clean on the touched code.

## Not in this PR

Pre-existing, also reachable today with `CODE_SANDBOX_ID`:

- `restart_notebook` on an attached kernel reconnects instead of restarting, but reports success. Fix: REST `POST /api/kernels/{id}/restart`.
- With `kernel.auto-restart` on, a dead attached kernel is replaced from the configuration without telling the caller.
- The sandboxes extension forwards `code_sandbox_id` only for `jupyter-server`, `google-colab` and `kaggle`. The generic variants (`datalayer`, `docker`, `modal`, ...) do not apply it yet.

### DIFF
--- a/docs/sourcey/mcp.json
+++ b/docs/sourcey/mcp.json
@@ -1679,7 +1679,7 @@
           },
           "kernel_id": {
             "default": null,
-            "description": "Specific kernel ID to use (will create new if skipped)",
+            "description": "Id of an existing kernel (or sandbox, for a non-Jupyter sandbox variant) to attach the notebook to. A new one is created if skipped.",
             "title": "Kernel Id",
             "type": "string"
           }
--- a/docs/sourcey/tools/use_notebook.md
+++ b/docs/sourcey/tools/use_notebook.md
@@ -19,7 +19,7 @@ Reactivate previously activated notebook using same notebook_name and notebook_p
 | `notebook_name` | string | yes | — | Unique identifier for the notebook |
 | `notebook_path` | string | yes | — | Path to the notebook file, relative to the Jupyter server root (e.g. 'notebook.ipynb') |
 | `mode` | `connect` · `create` | no | `"connect"` | Notebook operation mode: 'connect' to connect to existing and activate it, 'create' to create new and activate it |
-| `kernel_id` | string | no | `null` | Specific kernel ID to use (will create new if skipped) |
+| `kernel_id` | string | no | `null` | Id of an existing kernel (or sandbox, for a non-Jupyter sandbox variant) to attach the notebook to. A new one is created if skipped. |
 
 ## Output
 
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -796,7 +796,11 @@ async def use_notebook(
         ),
     ] = "connect",
     kernel_id: Annotated[
-        str, Field(description="Specific kernel ID to use (will create new if skipped)")
+        str,
+        Field(
+            description="Id of an existing kernel (or sandbox, for a non-Jupyter sandbox variant)"
+            " to attach the notebook to. A new one is created if skipped."
+        ),
     ] = None,
 ) -> ToolAnswer:
     """Use a notebook and activate it for following cell operations.
--- a/jupyter_mcp_server/tools/use_notebook_tool.py
+++ b/jupyter_mcp_server/tools/use_notebook_tool.py
@@ -410,6 +410,14 @@ class UseNotebookTool(BaseTool):
             # # Create/connect to kernel based on mode
             if mode == ServerMode.MCP_SERVER and sandbox_server_client is not None:
                 if kernel_id is not None:
+                    if config.uses_sandbox_variant():
+                        # The extension builds the sandbox and never sees the
+                        # id; attaching to a Jupyter kernel is a jupyter-server
+                        # variant thing.
+                        return (
+                            f"kernel_id only applies to the 'jupyter-server' sandbox variant;"
+                            f" the configured variant is '{config.sandbox_variant}'."
+                        )
                     kernels = sandbox_server_client.kernels.list_kernels()
                     kernel_exists = any(kernel.id == kernel_id for kernel in kernels)
                     if not kernel_exists:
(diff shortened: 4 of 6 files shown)


echarles:
It looks tailored for a jupyter kernel only, can this be made agnostic, so support all sandbox variants?

echarles:
Docs need an update, you can run ` make sync-sourcey `

EnyMan:
Done, should be agnostic now. The id rides `config.code_sandbox_id` into the extension hook.
Also the extension applies `code_sandbox_id` only for `jupyter-server`, `google-colab` and `kaggle` only. What about the `datalayer`, `docker`, `modal`?

echarles:
> What about the datalayer, docker, modal?

For now, you can use "code_sandbox_id" also, if not supported, that will need to be fixed in code_sandboxes.

EnyMan:
regenerated docs

echarles (approved):
LGTM Thx @EnyMan

PR #440, opened 2026-09-08
fix(revalidation): keep the declared envelope in a not-modified reply

serhiizghama:
The ETag round trip documented in `docs/architecture/caching` raises on the official SDK client instead of saving anything. Send back a matching etag for `read_notebook` or `read_cell` and `ClientSession.call_tool` blows up with `RuntimeError: Tool read_notebook has an output schema but did not return structured content` — the revalidation extension answers with `structured_content=None`, and both tools declare an output schema, so `validate_tool_result` (`mcp/client/session.py:1145`) rejects the reply before the caller ever sees the `notModified` flag.

The history is easy to follow: `661a4e8` gave all the tools output schemas so a client would not have to call one to learn what comes back, and `0ba594d` added revalidation two days later, which quietly reintroduced the missing-schema case on that one branch.

The fix is one field — `not_modified()` now carries `{"kind": ...}`, taken from the result the tool actually produced. The payload is still omitted, which is where the saving is, and `kind` is the only property either schema requires (checked against the live schemas of both tools). Nothing upstream is going to cover this: python-sdk #3345/#3346 deal with the neighbouring `structuredContent: null` on the wire and say explicitly that servers on this SDK never emit that shape, since results serialise with `exclude_none=True`. "Schema declared, field omitted" is deliberate client behaviour.

I had to change one existing assertion — `test_revalidation.py` asserted `structured_content is None`, which reads more like a description of what the code happened to do than a contract; the "not modified rather than nothing" guarantee lives in the `notModified` key and is still asserted separately. The reason none of the existing tests caught this is worth noting: they call `intercept_tool_call` with a stub `call_next` and inspect the result themselves, and the schema is only enforced one layer further out, on the client. So the new test runs the round trip through a real `ClientSession` over in-memory streams — it fails on current `main` with exactly the `RuntimeError` above.

Full suite is green (1179 passed, 3 skipped), and ruff/mypy on the touched files show no new findings.

### DIFF
--- a/docs/docs/architecture/caching/index.mdx
+++ b/docs/docs/architecture/caching/index.mdx
@@ -53,6 +53,12 @@ It says `notModified` rather than simply being empty, because a client that
 read an empty result as an empty *notebook* would show somebody no cells at
 all.
 
+The reply still carries `"structuredContent": {"kind": "notebook.read"}`. Both
+tools advertise an output schema, and a client validates what comes back
+against it, so an answer with no structured content at all fails there rather
+than being read as *not modified*. Only the payload is left out — which is
+where the saving is.
+
 :::caution This saves bandwidth, not work
 
 The tool still runs. The version is computed from what it answered, so there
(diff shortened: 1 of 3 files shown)


echarles (commented):
LGTM Thx @serhiizghama

echarles (approved):

PR #441, opened 2026-09-09
fix(notebook_manager): don't freeze the caller's token onto a notebook binding

serhiizghama:
`add_notebook` resolves the document token when it records the binding, and `resolved_document_token()` hands back the caller's credential whenever a request is being served. So the entry keeps whichever token first opened the notebook, and since `NotebookConnection.__aenter__` prefers the stored value over resolving one, every later connection on that alias goes out with that older token rather than the current caller's.

Where this bites is the shape the identity docs call supported — a worker serving one user still sees many of their tokens over time. After a rotation the notebook connection keeps presenting the one from the request that opened it. (With an alias shared between callers it's someone else's token, but that case is already covered by "one process per user".)

I kept a stored fallback rather than dropping it, so anything relying on the entry having a token still works — it's the configured token now, split out of `resolved_document_token()`. Same value in a single-user setup, minus the caller's credential. `notebook_info["token"]` is read in exactly one place, that fallback in `__aenter__`, so nothing else moves.

Tests went into `test_identity_per_request.py`, which already owns this property; the first one fails on `main` with `alice-token != bob-token`. Full suite is unchanged against a clean-`main` control run — the `[jupyter_extension]` cell-mutation tests flap either way, on different names each run.

### DIFF
--- a/jupyter_mcp_server/config.py
+++ b/jupyter_mcp_server/config.py
@@ -177,9 +177,14 @@ class JupyterMCPConfig(BaseModel):
         everyone. Reading it here rather than at each call site means a tool
         cannot forget to ask.
         """
-        caller = _caller_token()
-        if caller:
-            return caller
+        return _caller_token() or self.configured_document_token()
+
+    def configured_document_token(self) -> str | None:
+        """Document server token as configured, without the caller's.
+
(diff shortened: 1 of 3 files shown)


echarles (approved):
LGTM Thx @serhiizghama

PR #442, opened 2026-09-09
feat(execute): run cells through the runtime's /execute route so outputs survive the worker

echarles:
## What

In MCP_SERVER mode the worker drives the kernel over a WebSocket and writes cell outputs into the document itself, so a run's outputs stop when the worker stops (gateway rollout, idle reap, crash). This routes cell execution through the runtime's own `POST /api/kernels/{id}/execute` instead: the runtime runs the cell and writes outputs into the collaborative document **server-side** (`jupyter-server-nbmodel`'s `ExecutionStack`), so a run's outputs — and the rest of the run — survive the loss of the worker that started it.

- `execute_via_execution_stack_http` (utils.py): POSTs and polls the request URL, branching on the HTTP status (202 pending, 200 complete, 300 input, 500 error). The route serializes the same result dict the in-process `ExecutionStack.get` returns, so the terminal handling (execution_count, string-vs-mapping error, `MissingKernelError` for a gone kernel, outputs parse, the single `AFTER_EXECUTE`) is shared with the in-process driver via a new `_finish_execution_stack_result`.
- `document_id_from_ws_url` (utils.py): parses the RTC room id off the `NbModelClient` ws_url (`json:notebook:<fileId>` as-is, a bare Datalayer file id wrapped).
- `config.execute_via_http` (default off; reads `JUPYTER_MCP_EXECUTE_VIA_HTTP`): wired into `execute_cell`'s MCP_SERVER branch, gated on the default `jupyter-server` variant **and** a resolvable document+cell id, WebSocket path the fallback — routing a run whose outputs would land nowhere is the one thing it must not do. `execute_code` (no cell) keeps the WebSocket path.

## Tests

- `tests/test_execute_via_execution_stack_http.py` — the HTTP driver against a faked transport (complete, string/mapping error, missing kernel, input-required, timeout) + `document_id_from_ws_url`.
- `tests/test_execute_cell_via_http_routing.py` — the routing decision (HTTP taken with document+cell; fallback when off, on a sandbox variant, or with no room).
- Existing in-process and dead-kernel-retry tests still green after the shared-helper refactor.

Off by default, so behaviour is unchanged until `JUPYTER_MCP_EXECUTE_VIA_HTTP` is set.

### DIFF
--- a/docs/docs/architecture/index.mdx
+++ b/docs/docs/architecture/index.mdx
@@ -202,6 +202,22 @@ document with `NbModelClient` (jupyter-nbmodel-client over Y.js), so cell edits
 land in the same document the user's JupyterLab is showing. Execution goes
 through a `CodeSandboxClient` built by `sandbox_client.py`.
 
+By default that client drives the kernel from this process and writes each
+output into the document as it arrives, so a cell's outputs stop if this
+process stops — a gateway rollout, an idle reap, a crash. Set
+`execute_via_http` (env `JUPYTER_MCP_EXECUTE_VIA_HTTP`) and, on the default
+`jupyter-server` variant, `execute_cell` instead **POSTs to the runtime's own
+`/api/kernels/{id}/execute` route** (jupyter-server-nbmodel) and polls the
+request it returns. The runtime runs the cell and writes the outputs into the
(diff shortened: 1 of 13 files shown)


echarles (commented):
LGTM

PR #443, opened 2026-09-09
feat(execute): no WebSocket fallback for execute_via_http (2.1.12)

echarles:
Follow-up to #442. When `execute_via_http` is on for a variant that has the runtime `/execute` route, that route is now **the** execution path for `execute_cell` — **no WebSocket fallback**. A run whose document or cell id cannot be resolved is raised as a clear error (so it is fixed) rather than quietly downgraded to the worker-driven WebSocket path, which would lose the server-side output durability and hide the problem.

- Variants without the route (Modal, E2B, …) still execute through their sandbox client.
- `execute_code` (no cell) keeps the WebSocket path.
- Routing test updated to assert it **raises** rather than falls back; architecture doc updated.

Bumps to 2.1.12.


echarles (commented):
LGTM

PR #445, opened 2026-09-10
fix(read-notebook): handle empty notebooks

### LINKED ISSUE #444: read_notebook rejects an empty notebook as an out-of-range page
## Problem

A valid nbformat notebook may contain zero cells. In JUPYTER_SERVER mode, after `use_notebook` connects to such a file, the default `read_notebook(notebook_name=..., start_index=0)` call returns:

```text
Start index 0 is out of range. Notebook has 0 cells.
```

The tool already has a valid empty representation (`Notebook is empty`) in `Notebook.format_output`, but the pagination guard returns before it can be used. This prevents MCP clients from distinguishing a valid empty notebook from an invalid page request.

## Reproduction

1. Write `nbformat.v4.new_notebook()` to an `.ipynb` file.
2. Connect to it with `use_notebook` through the Jupyter server extension.
3. Call `read_notebook` with its defaults.

The behavior also reproduces directly through the shared `ReadNotebookTool` with an empty notebook model. Both `brief` and `detailed` formats return the range error.

## Expected

`start_index=0` should return the normal zero-cell summary and `Notebook is empty`. A nonzero start index should remain out of range.

AI assistance disclosure: OpenAI Codex was used to investigate and prepare this report.

emecii:
## Summary

Allow `read_notebook(start_index=0)` to return the existing empty-notebook representation when a valid notebook has zero cells. Nonzero offsets remain out of range.

Fixes #444.

## Validation

- 84 focused read/resource/result tests passed, including a real Jupyter-extension MCP call against an empty `.ipynb`
- `make sync-sourcey`
- Ruff check/format on the new regression and changed files
- pyproject validation, Python compilation, package sdist/wheel build, and `git diff --check`

## AI assistance

OpenAI Codex was used to investigate, implement, and test this change.

### DIFF
--- a/jupyter_mcp_server/tools/read_notebook_tool.py
+++ b/jupyter_mcp_server/tools/read_notebook_tool.py
@@ -80,7 +80,10 @@ class ReadNotebookTool(BaseTool):
         else:
             raise ValueError(f"Invalid mode or missing required clients: mode={mode}")
 
-        if start_index >= len(notebook):
+        # Index zero is the canonical first page of an empty collection, not a
+        # position past it. Let Notebook.format_output render "Notebook is
+        # empty" while keeping genuinely advanced offsets out of range.
+        if start_index >= len(notebook) and not (start_index == 0 and len(notebook) == 0):
             return f"Start index {start_index} is out of range. Notebook has {len(notebook)} cells."
 
         info_list = [f"Notebook {notebook_name} has {len(notebook)} cells.\n"]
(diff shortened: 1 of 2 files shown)


emecii:
The two Ubuntu failures appear unrelated to this PR and need a maintainer rerun:

- Python 3.10: all three new direct regression tests passed; the suite later errored while starting an unrelated MCP fixture on random port 45597 (`test_resources_both_modes.py`).
- Python 3.13: all three new direct regression tests passed; the pre-existing `test_execute_code_runs_in_the_requested_kernel` timed out after 30s. The same test passed on Ubuntu 3.10 and both macOS jobs.

I attempted to rerun the failed jobs, but GitHub rejected it because the fork account lacks repository admin permission. No code change is warranted by these logs.

echarles (approved):
LGTM Thx @emecii

PR #448, opened 2026-09-10
fix(identity): stop forwarding the MCP endpoint secret as the Jupyter credential

sachhg:
## What happens

When the MCP server logs in to Jupyter with a password rather than a token (`--document-password` / `--code-sandbox-password`), every notebook operation sends the MCP client secret to the Jupyter server:

- as the token on `PUT /api/collaboration/session/<path>`
- as a `?token=` query parameter on the collaboration room WebSocket URL

Nothing fails, because the login cookie is what actually authenticates those requests. The secret is sent anyway, to a server that has no use for it.

With a configured Jupyter token (`--document-token` / `--code-sandbox-token`) this does not happen: the token stored for each managed notebook wins over the caller's. Kernel requests use the configured credential in every setup. An earlier version of this description said the token setup was broken, which was wrong; see the thread for the correction.

## Reproduction on unmodified `main`

A Jupyter identity provider that records the token each request presents, then authenticates as usual:

```python
# tokenlog.py
import os
from jupyter_server.auth.identity import PasswordIdentityProvider

class LoggingIdentityProvider(PasswordIdentityProvider):
    def get_user(self, handler):
        with open(os.environ["TOKENLOG"], "a") as f:
            f.write(f"{handler.request.method} {handler.request.path} token={self.get_token(handler)!r}\n")
        return super().get_user(handler)
```

Jupyter with a password and no token, started in a directory holding a `notebook.ipynb`, and the MCP server logging in with that password:

```
TOKENLOG=tokens.log PYTHONPATH=. jupyter lab --port 8899 --no-browser \
  --IdentityProvider.token= \
  --PasswordIdentityProvider.hashed_password="$(python -c 'from jupyter_server.auth import passwd; print(passwd("jupyter-pw"))')" \
  --ServerApp.identity_provider_class=tokenlog.LoggingIdentityProvider

python -m jupyter_mcp_server --transport streamable-http --port 4040 \
  --document-url http://localhost:8899 --document-id notebook.ipynb --document-password jupyter-pw \
  --code-sandbox-url http://localhost:8899 --code-sandbox-password jupyter-pw \
  --mcp-token MCP_CLIENT_SECRET
```

A client sending `Authorization: Bearer MCP_CLIENT_SECRET` calls `read_cell` and then `insert_cell`. Both succeed on `main` and on this branch. What Jupyter was shown:

```
main
  PUT /api/collaboration/session/notebook.ipynb   token='MCP_CLIENT_SECRET'  (3 of 4)
  GET /api/collaboration/room/<id>                token='MCP_CLIENT_SECRET'  (3 of 3)

this branch
  PUT /api/collaboration/session/notebook.ipynb   token=''                   (4 of 4)
  GET /api/collaboration/room/<id>                token=''                   (3 of 3)
```

The same run with `--document-token` / `--code-sandbox-token` set to the Jupyter token shows only that token on both branches.

## The chain

1. `CodeSandboxTokenVerifier` (`server.py:108`) returns an `AccessToken` whose `token` is the MCP shared secret.
2. `IdentityMiddleware` publishes the caller for each request, and `identity.py:163` copies that token into `Identity.token`.
3. `resolved_document_token()` (`config.py:197`) prefers the caller's token over the configured one.
4. `NotebookConnection` (`notebook_manager.py:76`) uses the token stored for the notebook and falls back to the resolved one. With a configured token the stored one wins. With password auth nothing is stored, so the MCP secret is what goes out, into both the session request and the room WebSocket URL.

## The fix

`identity_from_access_token` already documents the sanctioned escape hatch at `identity.py:151`:

> A verifier that would rather not do that can return an `AccessToken` with no token.

The built in shared secret verifier never took it. This adds `EndpointOnlyVerifier`, which wraps the verifier selected for `--mcp-token` and empties `token` on the verified access token. Verification itself is untouched: the same secrets are accepted and the same secrets are refused.

Forwarding is preserved for a deployment's own verifier configured through the token verifier class setting, where the bearer token genuinely is the caller's upstream credential.

It matches intent already in the codebase: `notebook_manager.py:220` stores the configured token, "never the resolved one", for the same reason. This handles the case where nothing is configured.

## Why emptying the token is safe

In `mcp`'s `server/auth/middleware/bearer_auth.py`, the authorization decision reads only `expires_at`, `resource`, `scopes`, and `client_id`. `AccessToken.token` is not read anywhere in the SDK, so emptying it is inert to authentication and authorization.

## Impact

Low severity. Jupyter scrubs `token` from its own request logs, but a reverse proxy or ingress in front of it usually records full URLs, including the room WebSocket's query string. Anyone who can read those logs and reach the MCP endpoint can authenticate as an MCP client, without knowing the Jupyter password. There is no authentication bypass: the MCP endpoint verifies exactly what it did before.

It went unnoticed because nothing fails. The cookie authenticates the requests, and the secret never appears anywhere Jupyter itself logs.

## Tests

`TestEndpointOnlyVerifier` in `tests/test_identity.py` covers:

- the shared secret is accepted but does not become `Identity.token`
- with a configured Jupyter token, both resolvers return it rather than the shared secret
- a wrong secret is still refused, so emptying the credential does not soften the check that earns it
- the wrapper satisfies the `TokenVerifier` protocol

`test_falls_back_to_the_shared_secret` is updated to assert the wrapper and the wrapped `CodeSandboxTokenVerifier`.

`tests/test_identity.py` passes, 29 tests. `tests/test_auth.py` produces identical results with and without this change. A few `TestPasswordAuthE2E` cases fail in my environment both before and after, unrelated to this change.

## Out of scope, happy to follow up

- `server.py:108` could set `token=""` directly on `CodeSandboxTokenVerifier` instead of wrapping. I kept the change in `identity.py` so the built in and custom verifier paths stay handled in one place, but I am glad to move it.

### DIFF
--- a/jupyter_mcp_server/identity.py
+++ b/jupyter_mcp_server/identity.py
@@ -149,7 +149,9 @@ def identity_from_access_token(access_token: AccessToken) -> Identity:
     server acting for many users presents each user's own token to the
     document and sandbox servers rather than one configured for all of them.
     A verifier that would rather not do that can return an ``AccessToken``
-    with no ``token``, and the configured credential is used as before.
+    with no ``token``, and the configured credential is used as before —
+    which is what :class:`EndpointOnlyVerifier` makes the built-in
+    ``--mcp-token`` verifier do.
     """
     scopes = tuple(getattr(access_token, "scopes", ()) or ())
     return Identity(
@@ -189,6 +191,39 @@ def identity_from_jupyter_user(user: Any) -> Identity:
     )
 
 
+class EndpointOnlyVerifier:
+    """A verifier whose bearer token authenticates this endpoint and nothing else.
+
+    ``--mcp-token`` / ``MCP_TOKEN`` is the shared secret that says who may call
+    the MCP endpoint. It is documented as independent of the Jupyter
+    credential — ``--code-sandbox-token`` / ``--document-token`` — and a
+    deployment is expected to give it a different value.
+
+    :func:`identity_from_access_token` otherwise carries a verified bearer
+    token through as :attr:`Identity.token`, and
+    :meth:`~jupyter_mcp_server.config.JupyterMCPConfig.resolved_document_token`
+    prefers that over the configured one. For a platform's own verifier that is
+    the point: the bearer token *is* the caller's credential upstream. For the
+    shared secret it is wrong twice over — every document and kernel request
+    would present the MCP endpoint's secret to the Jupyter server, which fails
+    wherever the two values differ, and hands that secret to a server with no
+    business holding it.
+
+    So this passes the verification through unchanged and empties the token,
+    the documented way for a verifier to say "use whatever is configured".
+    """
+
+    def __init__(self, verifier: TokenVerifier) -> None:
+        #: The verifier doing the actual checking.
+        self.verifier = verifier
+
+    async def verify_token(self, token: str) -> AccessToken | None:
+        access_token = await self.verifier.verify_token(token)
+        if access_token is None:
+            return None
+        return access_token.model_copy(update={"token": ""})
+
+
 def load_token_verifier_class(path: str) -> type:
     """Import a verifier class named as ``module:Class`` or ``module.Class``."""
     if ":" in path:
@@ -212,7 +247,8 @@ def resolve_token_verifier(default_token: str | None = None) -> TokenVerifier | 
     1. ``JUPYTER_MCP_TOKEN_VERIFIER_CLASS`` — a class supplied by the
        deployment, which is how a platform plugs in its own OAuth;
     2. ``default_token`` — the shared secret of ``--mcp-token``, the simple
-       case that needs no code;
+       case that needs no code. It authenticates the endpoint only, so it is
+       wrapped in :class:`EndpointOnlyVerifier` and never travels upstream;
     3. nothing, leaving the endpoint unauthenticated, which the caller is
        expected to refuse unless it was asked for explicitly.
     """
@@ -232,7 +268,7 @@ def resolve_token_verifier(default_token: str | None = None) -> TokenVerifier | 
         from jupyter_mcp_server.server import CodeSandboxTokenVerifier
 
         logger.info("MCP endpoint token authentication enabled (using MCP_TOKEN)")
-        return CodeSandboxTokenVerifier(default_token)
+        return EndpointOnlyVerifier(CodeSandboxTokenVerifier(default_token))
 
     return None
 
--- a/tests/test_identity.py
+++ b/tests/test_identity.py
@@ -19,6 +19,7 @@ import pytest
 
 from jupyter_mcp_server.identity import (
     TOKEN_VERIFIER_CLASS_ENV,
+    EndpointOnlyVerifier,
     Identity,
     TokenVerifier,
     current_identity,
@@ -200,7 +201,8 @@ class TestResolveTokenVerifier:
 
         verifier = resolve_token_verifier("secret-token")
 
-        assert isinstance(verifier, CodeSandboxTokenVerifier)
+        assert isinstance(verifier, EndpointOnlyVerifier)
+        assert isinstance(verifier.verifier, CodeSandboxTokenVerifier)
 
     def test_a_named_class_wins_over_the_shared_secret(self, monkeypatch):
         """The point of the hook: a platform's own OAuth takes precedence."""
@@ -230,6 +232,65 @@ class TestResolveTokenVerifier:
         access_token = await verifier.verify_token("anything")
 
         assert identity_from_access_token(access_token).client_id == "test-client"
+
+
+class TestEndpointOnlyVerifier:
+    """``MCP_TOKEN`` says who may call this endpoint, and nothing more.
+
+    It is documented as independent of the Jupyter credential, so a deployment
+    gives it a different value. Carrying it through as `Identity.token` would
+    make `resolved_document_token()` prefer it over the configured token, and
+    every document and kernel request would present the MCP endpoint's secret
+    to the Jupyter server.
+    """
+
+    @pytest.mark.asyncio
+    async def test_the_shared_secret_is_not_a_credential_upstream(self, monkeypatch):
+        monkeypatch.delenv(TOKEN_VERIFIER_CLASS_ENV, raising=False)
+
+        verifier = resolve_token_verifier("MCP_CLIENT_SECRET")
+        access_token = await verifier.verify_token("MCP_CLIENT_SECRET")
+
+        assert access_token is not None, "the shared secret must still be accepted"
+        assert not identity_from_access_token(access_token).token
+
+    @pytest.mark.asyncio
+    async def test_the_configured_jupyter_token_is_what_travels(self, monkeypatch):
+        """The end of the chain: what a tool actually presents to Jupyter."""
+        monkeypatch.delenv(TOKEN_VERIFIER_CLASS_ENV, raising=False)
+        from jupyter_mcp_server.config import get_config, reset_config, set_config
+
+        reset_config()
+        try:
+            set_config(
+                code_sandbox_url="http://jupyter:8888",
+                code_sandbox_token="JUPYTER_SECRET",
+                document_url="http://jupyter:8888",
+                document_token="JUPYTER_SECRET",
+            )
+            verifier = resolve_token_verifier("MCP_CLIENT_SECRET")
+            access_token = await verifier.verify_token("MCP_CLIENT_SECRET")
+
+            token = set_current_identity(identity_from_access_token(access_token))
+            try:
+                assert get_config().resolved_document_token() == "JUPYTER_SECRET"
+                assert get_config().resolved_code_sandbox_token() == "JUPYTER_SECRET"
+            finally:
+                reset_current_identity(token)
+        finally:
+            reset_config()
+
+    @pytest.mark.asyncio
+    async def test_a_wrong_token_is_still_refused(self, monkeypatch):
+        """Emptying the credential must not soften the check that earns it."""
+        monkeypatch.delenv(TOKEN_VERIFIER_CLASS_ENV, raising=False)
+
+        verifier = resolve_token_verifier("MCP_CLIENT_SECRET")
+
+        assert await verifier.verify_token("JUPYTER_SECRET") is None
+
+    def test_it_satisfies_the_verifier_protocol(self):
+        assert isinstance(EndpointOnlyVerifier(AcceptingVerifier()), TokenVerifier)
 
 
 class TestTokenVerifierProtocol:


sachhg:
Note on CI: the only failing check is `enforce-label`, which needs a maintainer to apply one of the triage labels. `bug` is the fitting one here. I do not have permission to label my own PR, so I cannot clear it myself.

The `test` matrix is also still queued and has not started yet. Everything that has run is green: builds on all three platforms for 3.10 and 3.13, `test_lint`, `header-license-fix`, `mcp_reference`, and `test_sdist`.

Happy to rebase or adjust if anything is needed.

echarles:
Thx for this @sachhg - As this touches sensible parts, I'd love more eyes to review. @abbbe Any opinions on this?

abbbe:
@echarles I can try, but this is something related to sandbox, seems to be a new feature you guys have implemented while I was not actively following up on the project...

@sachhg could you please explain worst possible impact and maybe spell out a realistic attack / abuse scenario for this issue? this will help triage the problem

sachhg:
Thanks for taking a look, @abbbe. Working through your question I tested this end to end against a running server, and I need to correct the description first.

With a configured Jupyter token (`--document-token` / `--code-sandbox-token`), which is the setup the description leads with, the MCP secret is not sent. Every notebook the server manages stores the configured token, and that wins over the caller's. My reproduction called the resolvers directly and overstated it. Kernel requests use the configured token in every setup and are not affected either.

It does happen when no Jupyter token is configured, which in practice means password auth (`--document-password` / `--code-sandbox-password`). Then every notebook operation sends the MCP client secret to the Jupyter server: as the token on `PUT /api/collaboration/session/...`, and as a `?token=` query parameter on the collaboration room WebSocket URL. Notebook operations keep working, because the login cookie is what actually authenticates them, so nothing surfaces. I confirmed it with a Jupyter identity provider that logs the token each request presents: on main the session and room requests carry the MCP secret, and with this change they carry nothing.

The worst realistic impact is disclosure of the MCP endpoint secret. Jupyter scrubs `token` from its own logs, but a reverse proxy or ingress in front of Jupyter usually records full request URLs. Anyone who can read those logs and reach the MCP endpoint can then authenticate as an MCP client, which means reading and editing notebooks and running code in the kernel, without ever knowing the Jupyter password. There is no authentication bypass: the MCP endpoint still verifies exactly what it did before.

So I would triage it as low severity: a credential sent where it doesn't belong, in password-auth deployments. I'll update the description to match.

abbbe:
Afaik the security issue is real (even if not critical) and the proposed patch does the job. Merging the PR as is will slightly improves the security posture and does no immediate harm.

But while reviewing the code I could not shake the feeling something is off and pressed Claude a bit. I am not the author of OAuth path, so let me just share the last couple turns with you and let you guys decide what to do about it. My humble advice -- keep your code path related to authentication / authorisation as clean and as readable (to you) as possible.

===

❯ i still can't wrap my head around why a class called CodeSandboxTokenVerifier leaks a token by default, and we need yet another wrapper class around it to stop that.

CodeSandboxTokenVerifier isn't buggy — the field it's populating is being reinterpreted by a different layer for a purpose it was never designed for.

AccessToken.token is not this project's field — it's the mcp SDK's generic model (mcp.server.auth.provider.AccessToken). At that layer, "put the verified bearer token in .token" is just the ordinary, boring way to implement verify_token(): you checked this string, so you echo it back as part of "here's what I verified." Every generic TokenVerifier implementation looks like that — there's nothing sandbox-specific or careless about it. If you squint, token=token reads as almost a no-op: "yes, this is the token, obviously."

The leak isn't created there. It's created one layer up, in this project's own bridge code:
```
# identity.py — identity_from_access_token
token=str(getattr(access_token, "token", "") or ""),
```
and then:
```
# config.py
def resolved_document_token(self) -> str | None:
    return _caller_token() or self.configured_document_token()
```
This second layer silently reinterprets "the bearer token that authenticated the MCP call" as "the credential to hand to Jupyter." That reinterpretation is correct for exactly one case — a custom OAuth verifier in a multi-tenant deployment, where the bearer token genuinely is meant to double as the user's upstream Jupyter credential (that's the whole feature Identity.token/resolved_document_token() exists for). It's wrong for the other case — the built-in shared secret — where the bearer token was only ever meant to answer "may you call this endpoint at all," and has no relationship to Jupyter whatsoever.

So the actual defect is: one field (AccessToken.token) is overloaded to mean two different things depending on which verifier produced it, and nothing on the AccessToken object itself says which meaning applies. identity_from_access_token can't tell "a real upstream credential" from "an unrelated shared secret" just by looking at a string — they're indistinguishable at that point. Something has to supply the missing bit of information: does this particular token mean "forward me"?

There are exactly two places that missing bit could live:
1. Inside CodeSandboxTokenVerifier itself — but that bakes deployment-specific policy into a small, generic, already-independently-tested "compare a secret" primitive (the thing we discussed last message — it'd cost you that class's standalone contract for no real gain).
2. At resolve_token_verifier(), the one place that already knows, structurally, which kind of verifier it's handing back — built-in shared secret vs. a deployment's own class.

The wrapper isn't a patch bolted onto a broken class — it's attaching the missing signal at the only point in the code that actually has the information needed to supply it. CodeSandboxTokenVerifier keeps doing the ordinary, correct thing a TokenVerifier does; EndpointOnlyVerifier is the place that says "for this deployment path, that echoed-back token must never be treated as a forwardable credential." The name CodeSandboxTokenVerifier doesn't help your intuition here, admittedly — it reads as sandbox-specific when it's really just a generic HMAC-secret checker (recall it was apparently even called RuntimeTokenVerifier at some point) — but its behavior was never the bug; resolved_document_token()'s blind trust in Identity.token was.

===

❯ okay, so what would cleaner (even if more complicated and intrusive) fix could be?

The root cause I described — one field silently means two things — points to a specific class of fix: make the "should this travel upstream?" bit an explicit, separate signal instead of deriving it implicitly from which field is non-empty. A few options, roughly in order of how intrusive they are:

1. Piggyback on AccessToken.claims as an explicit opt-in marker.
AccessToken already has claims: dict[str, Any] | None — a free-form extension point (the SDK's own comment: "additional claims (e.g. iss, act)"). Instead of identity_from_access_token unconditionally copying access_token.token, it would only do so when the verifier explicitly says so:
```
FORWARD_CLAIM = "jupyter_mcp_forward_token"

def identity_from_access_token(access_token):
    forward = bool((access_token.claims or {}).get(FORWARD_CLAIM))
    return Identity(..., token=str(access_token.token) if forward else "")
```
A custom OAuth verifier that wants its bearer token forwarded to Jupyter sets claims={FORWARD_CLAIM: True} when it builds the AccessToken. CodeSandboxTokenVerifier needs no wrapper at all — it just never sets that claim, so it's safe by construction, not by a bolted-on decorator. This directly fixes the actual defect (an unmarked field being trusted) rather than papering over one caller of it.

Cost: it's a breaking change for any existing JUPYTER_MCP_TOKEN_VERIFIER_CLASS deployment relying on today's implicit forwarding — they'd need to add the claim. That's the real "intrusive" part: you're flipping the default from opt-out (forward unless stripped) to opt-in (don't forward unless marked), which is the safer posture but isn't free for whoever's already depending on the old behavior. Given the bug class is credential disclosure, I'd argue that trade is worth it — but it's a compatibility decision the maintainers would need to sign off on, not something you slip in as a "fix."

2. Split "authenticate the caller" from "resolve the upstream credential" into two separate hooks entirely.
Philosophically the cleanest: stop asking one verify_token() call to answer two unrelated questions ("who is this" and "what do I tell Jupyter"). Add a second, optional protocol:
class UpstreamCredentialResolver(Protocol):
    def upstream_document_token(self, identity: Identity) -> str | None: ...
resolved_document_token() calls this instead of reading Identity.token, only if the configured verifier also implements it. Plain shared-secret and no-op verifiers implement nothing extra and forwarding is structurally impossible for them — there's no field to misuse in the first place. This is the most work (a second Protocol, wiring it through resolve_token_verifier/config, docs, tests for a second extension point) but it's the version where "leaking the wrong thing" isn't a possible bug anymore, because there's no longer a single overloaded channel to leak through.

3. Thread the verifier identity through to identity_from_access_token, not just its output.
Rather than inferring intent from the AccessToken's contents, tag the verifier instance (e.g. isinstance(verifier, ForwardsUpstreamCredential)) and have the ASGI middleware pass that alongside the AccessToken when building Identity. This needs the most plumbing changes — _identity_from_scope currently only has scope["user"].access_token, not a reference to which verifier produced it — so you'd be threading a new piece of state through the SDK's own middleware stack. More invasive than #1 for the same conceptual result; I'd only reach for this if claims turned out to be unavailable or already used for something else.

My pick, if I were pushing for "do it right, not just fix the report": option 1. It removes the ambiguity at its actual source (an unmarked field on AccessToken), needs no wrapper class at all for the built-in case, costs one extra field check, and reuses infrastructure (claims) the SDK already designed for exactly this kind of "verifier hands the caller extra facts" use case. It's a real, disclosed breaking change for custom-verifier deployments, but that's an honest cost, not a hidden one — which is more than you can say for the status quo, where the implicit default was already silently wrong for one of its two callers.

That said — for landing this specific low-severity fix on a live PR, I wouldn't ask sachhg to redo it as option 1. The wrapper is correct, minimal, doesn't break anyone, and matches the size of the bug. Option 1 is the right thing to propose as a follow-up issue (possibly the same one you flagged for the CLI-default mismatch), not a blocker on #448.

===

sachhg:
Thanks @abbbe, that is a fair reading, and I agree with where it lands. The underlying problem is that `AccessToken.token` means two different things depending on which verifier produced it, and `identity_from_access_token` cannot tell them apart. The wrapper supplies that missing bit at the one place that knows which kind of verifier it built, which is why it stays small, but it leaves the ambiguity in place for the next verifier someone writes.

Making forwarding an explicit opt in through `AccessToken.claims` would remove it at the source, and the built in verifier would be safe by construction. The cost is real: a deployment whose custom verifier class relies on today's implicit forwarding would need to set the claim, so it is a compatibility call for you and @echarles.

If it helps, I am happy to open a follow-up issue laying out that option with the migration impact, and to take it on as a separate PR if you decide you want it. I would keep #448 as the minimal fix either way.

echarles:
If we need to break in a followup PR with a documented migration path, I would do it as part of this PR, unless you think there are good reasons to have separated ones, like eg better understanding that touchy security area. WDYT?

sachhg:
Updated in 2cfad54! Per your feedback @echarles, token forwarding is now an explicit opt-in via `FORWARD_CLAIM = "jupyter_mcp_forward_token"` in `AccessToken.claims`. `identity_from_access_token` only forwards the bearer token upstream when this claim is set, making default verifiers safe by construction without needing a wrapper class.

echarles:
Thx, Can you add a note for that in the docs (https://jupyter-mcp-server.datalayer.tech/security/identity/) and revert the notebooks in dev folder.

echarles:
@sachhg Gentle ping, this is nearly ready for merge.

sachhg:
Done in b0784ce! Added the documentation note explaining `FORWARD_CLAIM` opt-in token forwarding under `docs/docs/security/identity/index.mdx` and reverted the notebooks in `dev/content/` to match `main`. Thanks @echarles!

PR #339, opened 2026-08-03
fix(use_notebook): build created notebooks via nbformat so cells get a valid id

imcovangent:
## What
`use_notebook` (create mode) builds the seed notebook with nbformat's v4 constructors instead of a hand-written dict, so the format version and the per-cell `id` stay consistent.

## Why
Fixes #338. `use_notebook`'s create branch hand-built the notebook JSON:

- In the **1.1.4** release it declared `nbformat_minor: 5` but the seed cell had **no `id`** — nbformat 4.5 requires a unique `id` on every cell, so the written notebook is non-conformant with the version it claims (`'id' is a required property`).
- On **main** it was lowered to `nbformat_minor: 4`, which is valid but produces notebooks with **no cell ids at all** — worse for anything that keys on stable cell ids (cell-level diff/review, collaborative rooms) and inconsistent with `insert_cell`, which already creates cells via `nbformat.v4.new_code_cell()`.

## Change
```python
content = nbformat.v4.new_notebook(
    cells=[nbformat.v4.new_markdown_cell("New Notebook Created by Jupyter MCP Server")]
)
```
`new_notebook()` stamps the current `nbformat_minor` and `new_markdown_cell()` assigns the cell `id`, giving a valid nbformat 4.5 notebook with ids.

## Verification
```
nbformat: 4  minor: 5
cell has id: True -> a01a4577
nbformat.validate: OK
```
Both create paths (`contents_manager.new(...)` and `server_client.contents.create_notebook(...)`) receive the resulting `NotebookNode` unchanged (it is a `dict` subclass and serializes as before).

Fixes #338

Issue #338, opened 2026-08-03
use_notebook (create) writes a notebook with nbformat_minor: 5 but a cell missing the required id


imcovangent:
### Summary
When `use_notebook` creates a new notebook, it hand-builds the notebook JSON declaring `nbformat_minor: 5`, but the seed cell has no `id`. The nbformat 4.5 schema requires a unique `id` on every cell, so the written file is non-conformant with the `nbformat_minor` it claims.

### Where
`jupyter_mcp_server/tools/use_notebook_tool.py`, the create branch (v1.1.4, ~L211–224):

```python
content = {
    "cells": [
        {"cell_type": "markdown", "metadata": {}, "source": ["New Notebook Created by Jupyter MCP Server"]}
    ],
    "metadata": {},
    "nbformat": 4,
    "nbformat_minor": 5,
}
```

The cell dict has no `id`.

### Impact
- nbformat validation of the created notebook fails/warns (`'id' is a required property`).
- Tooling that keys on stable cell ids (e.g. anything correlating a cell across edits, cell-level diff/review, collaborative rooms) can't reference that first cell — it has no id until something later normalizes the notebook and assigns one.
- Inconsistent with `insert_cell`, which uses `nbformat.v4.new_code_cell()` and gets a valid id.

### Suggested fix
Build the seed notebook with nbformat's constructors instead of a raw dict, so ids and the format version stay consistent:

```python
import nbformat
nb = nbformat.v4.new_notebook(
    cells=[nbformat.v4.new_markdown_cell("New Notebook Created by Jupyter MCP Server")]
)
content = nb  # nbformat sets nbformat_minor and a per-cell id
```

(or, minimally, add `"id": <uuid>` to the cell dict).

### Environment
- `jupyter-mcp-server` 1.1.4

echarles:
makes sens, could open a PR with that change?

imcovangent:
Here it is: #339

PR #449, opened 2026-09-11
fix(insert_cell): don't write a 4.5 cell id into an older notebook

sachhg:
## What happens

`insert_cell` builds cells with `nbformat.v4.new_code_cell` and its siblings, which always attach a cell id. Cell ids arrived in nbformat 4.5, and the notebook is read with `as_version=4`, which converts the major version only — a file that declares 4.2 or 4.4 is still 4.2 or 4.4 when it is written back, now carrying a field its own schema forbids.

```
input nbformat 4.2 -> written 4.2  ids=[None, '4ba7a514']  INVALID: Additional properties are not allowed ('id' was unexpected)
input nbformat 4.4 -> written 4.4  ids=[None, 'aee3bd20']  INVALID: Additional properties are not allowed ('id' was unexpected)
input nbformat 4.5 -> written 4.5  ids=['8f374d00', '2a74495f']  VALID
```

`nbformat.write` logs `Notebook JSON is invalid` and writes the file anyway, so the tool answers `Cell inserted successfully` over a notebook that `nbformat.validate` rejects.

This is the mirror image of #338. That issue was the create path writing a 4.5 notebook whose cell had no id, and #339 fixed it by building cells through nbformat's constructors. `_insert_cell_file` already builds cells that way, which is exactly why it now writes an id into notebooks that cannot hold one.

The comment above the read says the version is pinned "to ensure consistency and support for cell IDs", which is the assumption that does not hold: `as_version` never touches `nbformat_minor`.

## Why it matters

`FileContentsManager.get()` and `.save()` both report `Notebook validation failed: Additional properties are not allowed ('id' was unexpected)` for the resulting file, and saving does not repair it.

The id is also not durable. `jupyter_ydoc` is minor aware: given a 4.2 notebook it reads the cell ids back as `None`, so the id this path writes disappears the moment the notebook is opened over RTC or through the WebSocket backend. An id that survives only until someone opens the notebook is the situation `cell_ids.py` exists to avoid.

`dev/content/notebook.ipynb` is checked in at `nbformat_minor: 2` with 161 cells, exactly one of which carries `"id": "a92dadf9"` — eight hex characters, the shape `random_cell_id` produces. It fails `nbformat.validate` on main today.

## The change

The id is dropped when the notebook predates 4.5. Upgrading the notebook to 4.5 instead would let the id stay, but that silently rewrites the version of a file the user did not ask to convert, so the file keeps the version it declared. A 4.5 notebook still gets an id exactly as before.

## Testing

`tests/test_insert_cell_nbformat_version.py` inserts each cell type into 4.2, 4.4 and 4.5 notebooks, then runs `nbformat.validate` on the result and checks the notebook's own version did not change.

On main the six pre-4.5 cases fail and the three 4.5 cases pass, which is the shape of the bug. With this change all nine pass, and the surrounding suite has the same set of failures before and after.

### DIFF
--- a/jupyter_mcp_server/tools/insert_cell_tool.py
+++ b/jupyter_mcp_server/tools/insert_cell_tool.py
@@ -119,7 +119,8 @@ class InsertCellTool(BaseTool):
         """
         # Read notebook file
         with open(notebook_path, encoding="utf-8") as f:
-            # Read as version 4 (latest) to ensure consistency and support for cell IDs
+            # as_version converts the major version only, so nbformat_minor stays
+            # whatever the file declared.
             notebook = nbformat.read(f, as_version=4)
 
         # Clean any transient fields from existing outputs (kernel protocol field not in nbformat schema)
@@ -138,6 +139,13 @@ class InsertCellTool(BaseTool):
             new_cell = nbformat.v4.new_markdown_cell(source=cell_source or "")
         elif cell_type == "raw":
             new_cell = nbformat.v4.new_raw_cell(source=cell_source or "")
+
+        # Cell ids arrived in nbformat 4.5 and the constructors always attach one.
+        # Writing it into an older notebook produces a file that its own schema
+        # rejects, so drop it rather than silently upgrading the user's notebook.
+        if notebook.nbformat_minor < 5:
+            new_cell.pop("id", None)
+
         notebook.cells.insert(actual_index, new_cell)
 
         # Write back to file
--- a/tests/test_cell_id_addressing.py
+++ b/tests/test_cell_id_addressing.py
@@ -83,6 +83,11 @@ async def test_an_edit_by_id_survives_an_insert_above_it(
     the newcomer; addressed by id, it lands on the cell the agent read.
     """
     async with mcp_client_parametrized as client:
+        # Skip before touching the notebook. This file is shared with the tests
+        # that run after this one, and skipping once TARGET is already in it
+        # would leave TARGET behind as their first cell.
+        await _id_or_skip(client, 0)
+
         await client.insert_cell(0, "markdown", "TARGET")
         assert "TARGET" in _text_of(await _read(client, 0))
         target_id = await _id_or_skip(client, 0)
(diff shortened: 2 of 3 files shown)


echarles:
Tests are failing with 

```
SKIPPED [1] tests/test_tools.py:348: TEST_MCP_SERVER disabled
FAILED tests/test_prompts.py::test_jupyter_cite[jupyter_extension] - AssertionError: Cell 0 should contain Matplotlib Examples
assert '# Matplotlib Examples' in "USER Cite cells [0] from notebook notebook, here are the cells:\n=====Cell 0 | type: markdown | execution count: N/A=====\nTARGET\n=====End of Cited Cells=====\nUSER's Instruction are follow: test prompt"
FAILED tests/test_resources_both_modes.py::test_a_notebook_reads_back_as_nbformat_json[jupyter_extension] - assert '# Matplotlib Examples' in '"TARGET"'
 +  where '"TARGET"' = <function dumps at 0x105c6e950>('TARGET')
 +    where <function dumps at 0x105c6e950> = json.dumps
===== 2 failed, 1162 passed, 23 skipped, 12 warnings in 451.65s (0:07:31) ======
make: *** [test-jupyter-server] Error 1
Error: Process completed with exit code 2.
``

sachhg:
Thanks for running it. Those two failures were caused by this PR, through a cell leaking between tests.

`test_an_edit_by_id_survives_an_insert_above_it` inserts `TARGET` into `dev/content/notebook.ipynb` and only then calls `_id_or_skip`. That fixture is nbformat 4.2, so with this change the inserted cell correctly has no id, the test skips, and it skips with `TARGET` already written into the shared notebook. The cleanup at the end never runs, so `test_prompts` and `test_resources_both_modes` read `TARGET` as cell 0 instead of `# Matplotlib Examples`.

Before this change `insert_cell` wrote an id into that 4.2 notebook, so the skip never actually fired and the leak stayed hidden.

I pushed a fix that checks for ids before inserting, so a notebook without them skips while it is still untouched.

One consequence worth flagging: the id addressing tests now skip against this fixture, because it predates 4.5. If you would rather they run, I can upgrade `dev/content/notebook.ipynb` to 4.5 in this PR. It is invalid as it stands anyway, being 4.2 with a single stray id on cell 160, which is the output this PR stops producing. Happy to go either way.

echarles (approved):
LGTM Thx @sachhg

PR #450, opened 2026-09-11
fix(outputs): strip the escapes that are not CSI

sachhg:
## What is left behind

#427 widened `strip_ansi_codes` from the SGR colour sequences to every CSI sequence, which is what a progress bar writes. CSI is one of several escape forms, and the rest still reach the client.

The one a kernel is most likely to write is an OSC string, because that is how a hyperlink is expressed. `rich` emits one whenever it believes it is writing to a terminal, as it does inside Jupyter:

```python
>>> from rich.console import Console
>>> import io
>>> buf = io.StringIO()
>>> Console(file=buf, force_terminal=True).print("see [link=https://example.com]the docs[/link] for details")
>>> buf.getvalue()
'see \x1b]8;id=12903882;https://example.com\x1b\\the docs\x1b]8;;\x1b\\ for details\n'
>>> strip_ansi_codes(buf.getvalue())
'see \x1b]8;id=12903882;https://example.com\x1b\\the docs\x1b]8;;\x1b\\ for details\n'
```

Nothing is removed. The same holds for a window title (`ESC ] 0 ; ... BEL`, which any shell command run from a cell can write), for the two character forms such as save and restore cursor or reverse index, for a charset selection like `ESC ( B`, and for the device control strings.

This matters for the same reason #426 did. The text goes to a model, which now reads a URL and an id it was never meant to see, and to a terminal, where the bytes are interpreted rather than displayed.

## The change

The pattern covers the other forms alongside CSI: an OSC string up to its BEL or ST terminator, the DCS, PM, APC and SOS strings up to ST, and the two character and charset sequences. The OSC and DCS bodies are bounded so a sequence that was truncated cannot swallow the rest of the output. The regex is compiled once at import rather than on every call.

Every CSI sequence is matched exactly as before, so nothing that #427 fixed changes.

## Testing

The existing tests are unchanged and still pass. New ones cover an OSC hyperlink captured from `rich`, an OSC string terminated by BEL, the two character escapes, a charset selection, and a device control string, plus a reading of the `rich` output through `extract_output` that asserts no escape byte survives, matching the shape of the nested tqdm test.

On main the six new tests fail and the five existing ones pass. With this change all eleven pass, and the surrounding suite has the same set of failures before and after.

### DIFF
--- a/jupyter_mcp_server/utils.py
+++ b/jupyter_mcp_server/utils.py
@@ -539,10 +539,22 @@ def extract_output(output: dict | Any) -> str | ImageContent:
         return f"[Unknown output type: {output_type}]"
 
 
+# Every escape a kernel can write, not only the CSI sequences: an OSC string
+# carries a hyperlink or a window title, DCS and friends carry device control,
+# and the two character forms move the cursor or select a charset.
+_ANSI_ESCAPE = re.compile(
+    r"\x1b(?:"
+    r"\[[0-?]*[ -/]*[@-~]"              # CSI, e.g. colour or cursor movement
+    r"|\][^\x07\x1b]*(?:\x07|\x1b\\)"  # OSC, terminated by BEL or ST
+    r"|[P^_X][^\x1b]*\x1b\\"           # DCS, PM, APC and SOS strings
+    r"|[ -/]*[0-~]"                    # two character and charset sequences
+    r")"
+)
+
+
 def strip_ansi_codes(text: str) -> str:
     """Remove ANSI escape sequences from text."""
-    ansi_escape = re.compile(r"\x1b\[[0-?]*[ -/]*[@-~]")
-    return ansi_escape.sub("", text)
+    return _ANSI_ESCAPE.sub("", text)
 
 
 def clean_notebook_outputs(notebook):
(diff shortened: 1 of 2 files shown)


echarles:
Does this address a specific mcp client? How can we validate this fix is good for the mcp client ecosystem?

sachhg:
It is not specific to one client. The stripping happens on the server before the output is sent, so every client receives the same cleaned text.

Two kinds of consumer are affected by the escapes getting through. The model reading the output sees the raw OSC framing, so a link printed by rich arrives as `\x1b]8;id=...;https://example.com\x1b\\the docs\x1b]8;;\x1b\\` rather than `the docs`: a URL and an id it was never meant to read, spent from its context. And any client that shows the text in a terminal interprets those bytes rather than displaying them, which can emit a live hyperlink or retitle the window.

To validate it against a real client, run a cell that prints a rich link, for example `from rich import print; print("[link=https://example.com]the docs[/link]")`, then read that cell's output through any MCP client. On main the output contains `\x1b]8;`; with this change it reads `the docs`. That is the same check #427 relied on for the tqdm cursor movement, extended to the escapes that are not CSI.

The rich sample in the tests is captured from rich itself rather than written by hand, so the test tracks what a kernel actually produces.

echarles (approved):
LGTM Thx @sachhg

Issue #446, opened 2026-09-10
Make agents aware of MCP server quirks (multi-agent race conditions + maybe others)


abbbe:
### Problem

Datalayer Jupyter MCP Server has some quirks which can bite users or agents. The ones which bite me personally are related to race conditions / concurrency.

(1) File extensions/prompt-templates/general/AGENT.md says:

> **ALWAYS MCP**: All operations on the Notebook, such as creating, editing, and code execution, MUST be performed via tools provided by Jupyter MCP. **NEVER Directly create or modify the Notebook Source File Content**.

This critical insight is not communicated to the agent. MCP server level instructions are implemented (see jupyter_mcp_server/server.py), but they only say:
```
INSTRUCTIONS = (
    "Read, edit and run Jupyter notebooks. Cells are addressed by index within "
    "a notebook you have opened with use_notebook, and execution happens on the "
    "server, so a long computation keeps running after this session ends."
)
```

(2) **WITHDRAWN - SEE #447**. There use_notebook tool which sets current notebook and other tools use that notebook. If agent A selects notebook A, then agent B selects notebook B, tools invocation of agent A will be applied to notebook B. Described in #256 and marked closed by PR #312 -- technically correct, the agents now have a way to ensure their actions applied to notebooks they mean to work on. But this behavior is optional and agents do not really know they must do it.

### Proposed Solution

First we need to enumerate all relevant bugs/features/quirks, maybe there are more than two mentioned above. Then consider options:

1. Extend MCP server level instructions -- probably good to address the first issue
2. **WITHDRAWN - SEE #447**. For 2nd issue - the same or extend tool-specific instructions. Or perhaps more robust solution is to make `notebook_name` parameter mandatory?

The first draft of this issue looked completely different, I was going to propose to create a custom skill to work teach agents how to  use Datalayer MCP server correctly. While spelling out the issues I have figure out external skills is probably wrong tool for these two issues, but maybe there are other uses.

The original idea was inspired by https://herdr.dev/docs/agent-skill/ -- which has two parts: (1) teaches agents how to use herds and (2) teaches agent to teach user how to use herdr ;). I looked around and found no skills specific to jupyter mcp servers (there is https://agenthivex.com/skills/datalayer-jupyter-mcp-server but it seems to be some auto-generated junk).

echarles:
Thx for starting the discussion. 

> First we need to enumerate all relevant bugs/features/quirks, m

[...]

PR #451, opened 2026-09-11
feat(instructions): state the ALWAYS MCP rule in the server instructions

abbbe:
Option 1 from #446.

`extensions/prompt-templates/general/AGENT.md:41` already states the rule, but nothing
loads it — the README asks a human to paste the template into a system prompt, so agents
nobody set up by hand never see it. `INSTRUCTIONS`, returned in the `initialize` reply, is
what every agent does read, and it said nothing about the file. It is also the one thing
about this server an agent cannot infer from `tools/list`.

The rule is pasted verbatim, plus one sentence saying why.

Notes for the reviewer:

- The same server blurb is restated in `docs/sourcey/snapshot.mjs:36` (hardcoded because
  mcp-parser drops `instructions` from the snapshot) and in the generated
  `docs/sourcey/mcp.json`. Left untouched here — the script could read `INSTRUCTIONS` from
  the installed package instead of restating it, if you want that fixed properly.

### DIFF
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -340,7 +340,14 @@ class MCPServerWithCORS(MCPServer):
 INSTRUCTIONS = (
     "Read, edit and run Jupyter notebooks. Cells are addressed by index within "
     "a notebook you have opened with use_notebook, and execution happens on the "
-    "server, so a long computation keeps running after this session ends."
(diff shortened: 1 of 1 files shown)


abbbe:
@echarles the I seems to fail due to some labelling -- I guess it is something on your side

echarles (approved):
LGTM Thx @abbbe

PR #453, opened 2026-09-12
Bind streamable-HTTP to loopback by default and decouple CORS wildcard from credentials

kta1kri:
Two hardening changes to the `streamable-http` transport, prepared for the maintainers who asked for fixes on the two draft advisories (GHSA-ww23-6c43-m7qh and GHSA-26gq-g6jf-8hvv).

**1. `--host`, defaulting to loopback**

`do_start` passed a hardcoded `host="0.0.0.0"` to uvicorn with no way to override it, so `jupyter-mcp-server start --transport streamable-http` was reachable from every interface of the machine. This adds a `--host` option (env `HOST`) defaulting to `127.0.0.1` and threads it through `_resolve_and_start` → `do_start` → `uvicorn.run`. Operators who want remote access pass `--host 0.0.0.0` deliberately, which is also what `jupyter server` itself asks for.

**2. CORS: no wildcard together with credentials**

`allow_origins=["*"]` with `allow_credentials=True` is not the permissive-but-harmless combination it looks like - Starlette then echoes the request's `Origin` back and adds `Access-Control-Allow-Credentials: true`, so a browser on any site can read authenticated responses. The change keeps the wildcard for the common token-in-header case but only turns credentials on for origins the operator names in `JUPYTER_MCP_CORS_ORIGINS` (comma-separated).

Neither change alters behaviour for the default stdio transport, and no option is removed - the previous behaviour is available as `--host 0.0.0.0` and an explicit origin list.

I am the reporter of both advisories. Happy to adjust naming (e.g. a CLI flag instead of an env var for the origins) if you prefer a different shape.

### DIFF
--- a/jupyter_mcp_server/cli/commands/serve.py
+++ b/jupyter_mcp_server/cli/commands/serve.py
@@ -43,6 +43,7 @@ def _resolve_and_start(
     jupyter_token: str | None,
     jupyter_password: str | None,
     port: int,
+    host: str,
     document_provider: str,
     jupyterlab: str,
     open_notebook_in_ui: str,
@@ -98,6 +99,7 @@ def _resolve_and_start(
         document_token=resolved_document_token,
         document_password=resolved_document_password,
         port=port,
+        host=host,
         document_provider=document_provider,
         jupyterlab=parse_bool_option(jupyterlab, "--jupyterlab"),
         open_notebook_in_ui=parse_bool_option(open_notebook_in_ui, "--open-notebook-in-ui"),
@@ -142,6 +144,17 @@ def server_callback(
             help="The port to use for the Streamable HTTP transport. Ignored for stdio transport.",
         ),
     ] = 4040,
+    host: Annotated[
+        str,
+        typer.Option(
+            "--host",
+            envvar="HOST",
+            help=(
+                "The address to bind for the Streamable HTTP transport. Defaults to loopback; "
+                "pass 0.0.0.0 to accept connections from other hosts. Ignored for stdio transport."
+            ),
+        ),
+    ] = "127.0.0.1",
     capability: Annotated[
         list[str] | None,
         typer.Option(
@@ -398,6 +411,7 @@ def server_callback(
         jupyter_token=jupyter_token,
         jupyter_password=jupyter_password,
         port=port,
+        host=host,
         document_provider=document_provider.value,
         jupyterlab=jupyterlab,
         open_notebook_in_ui=open_notebook_in_ui,
@@ -440,6 +454,17 @@ def start_command(
             help="The port to use for the Streamable HTTP transport. Ignored for stdio transport.",
         ),
     ] = 4040,
+    host: Annotated[
+        str,
+        typer.Option(
+            "--host",
+            envvar="HOST",
+            help=(
+                "The address to bind for the Streamable HTTP transport. Defaults to loopback; "
+                "pass 0.0.0.0 to accept connections from other hosts. Ignored for stdio transport."
+            ),
+        ),
+    ] = "127.0.0.1",
     capability: Annotated[
         list[str] | None,
         typer.Option(
@@ -693,6 +718,7 @@ def start_command(
         jupyter_token=jupyter_token,
         jupyter_password=jupyter_password,
         port=port,
+        host=host,
         document_provider=document_provider.value,
         jupyterlab=jupyterlab,
         open_notebook_in_ui=open_notebook_in_ui,
--- a/jupyter_mcp_server/server.py
+++ b/jupyter_mcp_server/server.py
@@ -6,6 +6,7 @@
 Jupyter MCP Server Layer
 """
 
+import os
 import hmac
 import re
 from typing import Annotated, Literal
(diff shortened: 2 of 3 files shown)


jupyter_mcp_server/cli/commands/serve.py
```diff
@@ -142,6 +144,17 @@ def server_callback(
...
+        str,
+        typer.Option(
+            "--host",
+            envvar="HOST",
+            help=(
+                "The address to bind for the Streamable HTTP transport. Defaults to loopback; "
+                "pass 0.0.0.0 to accept connections from other hosts. Ignored for stdio transport."
+            ),
+        ),
+    ] = "127.0.0.1",
```
ebarkhordar, line 157: I ran the documented Docker invocation against 7c508ae5 in a clean container on Linux. The new loopback default makes the published-port Docker recipe unreachable.

`docs/docs/mcp-transports/streamable-http/index.mdx:143` (MacOS/Windows) publishes with `-p 4040:4040` on a bridge network. Docker forwards a published port to the container's eth0 address, not to its loopback, so a 127.0.0.1 bind inside the container accepts nothing from outside it. The Linux recipe just below uses `--network=host`, where the bind still works, so only the bridge one breaks. Both entry points carry the default, this one at :157 and `start_command` at :467, and the image's `ENTRYPOINT ["python", "-m", "jupyter_mcp_server"]` reaches this one through `app.callback(invoke_without_command=True)`.

Built the image from this branch and ran the docs command unchanged against a JupyterLab container on a user-defined network:

```
$ docker run -d --rm -p 14040:4040 -e JUPYTER_URL=http://jlab:8888 \
    -e JUPYTER_TOKEN=MY_TOKEN -e MCP_TOKEN=MY_MCP_TOKEN \
    jmcp-pr453:test --transport streamable-http
INFO:     Uvicorn running on http://127.0.0.1:4040 (Press CTRL+C to quit)

$ curl -o /dev/null -w '%{http_code}\n' http://127.0.0.1:14040/mcp
curl: (56) Recv failure: Connection reset by peer
000
```

Inside that container 127.0.0.1:4040 connects and its own 172.20.0.2:4040 is refused. Adding `--host 0.0.0.0` to the same command binds 0.0.0.0 and the published port answers 200.

Since the option already reads `HOST`, one line in the Dockerfile fixes it without giving the default back:

    ENV HOST=0.0.0.0

Verified on the same image: `-e HOST=0.0.0.0` with no flag binds 0.0.0.0 and the published port answers 200, and `-e HOST=0.0.0.0 --host 127.0.0.1` still binds loopback, so the flag wins. A pip install keeps the loopback default.

I did not test MacOS or Windows Docker Desktop, only Linux bridge networking.

kta1kri:
You're right, and thanks for the precise repro — that's my regression.

I reproduced both halves independently before changing anything:

```
bind 127.0.0.1 -> published port: 000 (connection reset)
bind 0.0.0.0   -> published port: 200
```

and the typer precedence you relied on:

```
no env, no flag                 -> 127.0.0.1
HOST=0.0.0.0, no flag           -> 0.0.0.0
HOST=0.0.0.0 + --host 127.0.0.1 -> 127.0.0.1
```

So `ENV HOST=0.0.0.0` is the right shape: the image binds every interface inside its own network namespace (which still requires the operator to publish the port), a pip install keeps loopback, and an explicit `--host` overrides. Pushed in fae758f with a comment on the line so it doesn't get 'fixed' back later.

I also added a line under the pip recipe in `streamable-http/index.mdx` noting the loopback default and how to opt out — the docs didn't say it anywhere, which is part of why the Docker recipe broke quietly rather than loudly.

I only changed the Dockerfile and that doc note; the MacOS/Windows and Linux commands are unchanged and should both work as written now.

ebarkhordar:
That matches what I measured, so I think the Dockerfile env var is the right lever.

My case C was `-e HOST=0.0.0.0` with no flag: bound `0.0.0.0`, published port answered 200. Case D was `-e HOST=0.0.0.0 --host 127.0.0.1`: bound `127.0.0.1`, so the flag still wins over the env var and anyone who wants loopback inside the image can still ask for it.

I read fae758f rather than rebuilding the image against it. The doc note under the pip recipe is the part I could not have suggested from the diff, thanks for adding it.

echarles:
OK to move to 127.0.0.1 - The docker instructions and default configuration should be updated as part of this PR to avoid breaking Docker users.

echarles:
@kta1kri Gentle pling, this is nearly ready for merge.

kta1kri:
Both of those are already in the branch, at `fae758f`:

- **Docker instructions / default configuration**: the Dockerfile now sets `ENV HOST=0.0.0.0`, so `-p 4040:4040` keeps working unchanged for existing Docker users. `--host` on the command line still overrides it. @ebarkhordar measured the precedence independently (env alone binds `0.0.0.0`; env plus `--host 127.0.0.1` binds loopback).
- **Docs**: a note under the pip recipe in `streamable-http/index.mdx` states the loopback default, how to opt out, and why the image differs.

So the loopback default applies to a pip install, and the Docker recipe is unaffected. Nothing outstanding from my side.

echarles (approved):
LGTM Thx @kta1kri

PR #455, opened 2026-09-14
fix(cell-ids): a move names the cell that moved

### LINKED ISSUE #454: move_cell reports the id of the cell it moved onto, not the cell that moved
`move_cell` addressed by id hands back the wrong id. Given `source_cell_id` and `target_cell_id`, the result's `io.jupyter-mcp/cell_id` is the target's, so an agent that moves a cell and then addresses "the cell I just moved" acts on the cell it moved onto.

Against `dev/content/new.ipynb` (nbformat 4.5) at a259155, jupyter extension mode:

```
cell 0 id: 3c3442e4   print("Hello, World!")
cell 1 id: a70f6bd9   # A New Notebook

move_cell(source_cell_id="3c3442e4", target_cell_id="a70f6bd9", notebook_name="moverepro")
  -> "Cell moved successfully from index 0 to 1 (code)."
  -> meta {'io.jupyter-mcp/cell_id': 'a70f6bd9'}

after the move: cell 0 is a70f6bd9, cell 1 is 3c3442e4
```

The cell that moved is `3c3442e4`. The result names `a70f6bd9`, which is the one that stayed where it was.

`move_cell` resolves twice (`server.py:1557` for the source, `:1566` for the target) and `cell_ids.resolve` attaches its id through `add_meta(cell_id=...)`. That is one key, so the second call overwrites the first.

The same key feeds the `resources/updated` cell frame. `notifications.changed_cells` reads `cell_id` and `cell_ids` together and dedupes them, and the comment there says a move has a source and a target, so it is written expecting two ids. It gets one, and it is the wrong one: a client subscribed to `notebook://<nb>/cells/3c3442e4` never hears that its cell moved.

Only fires when `target_cell_id` is passed. With a target index there is a single resolve and the id is the source's, which is right.

PR coming.

ebarkhordar:
Fixes #454.

`cell_ids.resolve` attaches the id it landed on with `add_meta(cell_id=...)`, and `move_cell` calls it twice, so the target's resolve overwrote the source's and the result named the cell that stayed put.

`note_cell` keeps the first id named as the primary one and puts a later one in `cell_ids`, so a move now answers `cell_id: <the cell that moved>` plus `cell_ids: [<the other end>]`. `changed_cells` already merges and dedupes those two keys, so the cell frame names both ends, which is what its comment says it is written for.

One thing I ran into: the new tests are in `test_cell_ids.py` and not in `test_cell_id_addressing.py`, because everything in that file skips. `dev/content/notebook.ipynb` is nbformat 4.2, its cells have no ids, and `_id_or_skip` catches every test. I ran the move by hand against `new.ipynb`, which is 4.5, and the meta comes back `{'io.jupyter-mcp/cell_id': 'a70f6bd9', 'io.jupyter-mcp/cell_ids': ['3c3442e4']}` with `a70f6bd9` the cell that moved. Repointing that file at a 4.5 notebook is its own change, tell me if you want it.

The new tests fail on main except the one that moves a cell onto itself.


echarles (approved):
LGTM Thx @ebarkhordar

PR #457, opened 2026-09-15
fix(tasks): tasks/list hands back a cursor when a page was truncated

### LINKED ISSUE #456: tasks/list caps at 50 and never sets next_cursor, so tasks past the first page cannot be listed
`_handle_list` asks the store for a page and returns it without a cursor:

```python
async def _handle_list(self, ctx: Any, params: Any) -> ListTasksResult:
    records = await self.store.list()
    return ListTasksResult(tasks=[record.public() for record in records])
```

`MemoryTaskStore.list` defaults to `LIST_PAGE` (50), and `ListTasksResult` is a `PaginatedResult`, where an absent `next_cursor` means there are no more results. So with more than 50 live tasks the client is handed 50 and told that is all of them. `params.cursor` is not read either, so asking again with one gives back the same page.

The tasks after the 50th are then unreachable: their ids only ever appear in `tasks/list`, and `tasks/get` needs an id. The architecture page says `tasks/list` is "the tasks this server holds, newest first".

60 tasks in a `MemoryTaskStore`, called through the real binding at a4808782:

```
page 1 returned         : 50
page 1 next_cursor      : None
page 2 (cursor='50')    : 50 tasks, first=tsk_0000
page 2 == page 1        : True
distinct tasks reachable: 50 of 60
```

50 live tasks is not far off for an agent running cells as tasks with the default 15 minute retention.

One more thing that matters for fixing it: the sort key in `MemoryTaskStore.list` is `created_at`, which is ISO seconds, so tasks made in the same second tie and their order is whatever the dict held. In the run above all 60 shared a second, and "newest first" came back oldest first. A page boundary lands inside a tie like that.

I have a fix ready and will send it as a PR: honour the cursor, set `next_cursor` when a page was truncated, and make the sort key a total order so the boundary is stable. It leaves `TaskStore.list` alone since a deployment can supply its own through `JUPYTER_MCP_TASK_STORE_CLASS`.

ebarkhordar:
Fixes #456.

`_handle_list` returned one page and no `next_cursor`, and on a `PaginatedResult` an absent cursor means there is nothing more. So a client holding more than 50 live tasks got 50 and was told that was all of them, and since ids only appear in the listing, `tasks/get` could not reach the rest either. `params.cursor` was not read, so asking again with one gave the same page back.

It now reads the cursor, fetches one record past the page so "is there more" comes from what came back, and sets `next_cursor` when the page was truncated. A cursor this server did not issue is refused rather than read as zero, which would hand back page one and turn a client's walk into a loop.

One thing I changed beyond that, because pagination needs it: the sort key in `MemoryTaskStore.list` is now `(created_at, task_id)`. `created_at` is ISO seconds, so tasks made in the same second tie, and a page boundary inside a tie can repeat a task or skip one. In the repro all 60 shared a second and "newest first" came back oldest first.

`TaskStore.list` keeps its signature, since a deployment can point `JUPYTER_MCP_TASK_STORE_CLASS` at its own store and I did not want to break one. The cost is that page N asks the store for N pages worth, which seemed the right trade for a list bounded by the retention window, but say the word if you would rather the protocol grew a cursor.

Two tests: walking the cursor over `LIST_PAGE + 10` tasks that all share a second, and a bad cursor. Both fail on main.


echarles (approved):
LGTM Thx @ebarkhordar

PR #458, opened 2026-09-16
fix(restart_notebook): report a restarted kernel to the hooks only when one restarted

serhiizghama:
`restart_notebook` fired `KERNEL_LIFECYCLE` / `restarted` after the tool returned, whatever it had returned. So a restart that failed — the sandbox's `restart()` raised, `restart_kernel` rejected the kernel, reprovisioning didn't come through, or the notebook wasn't connected at all — still landed in every audit sink and in the OTel span file as a restart (with `kernel_id="unknown"` in the last case). The tool answer said "Failed to restart …" while the record next to it said a kernel restarted, which isn't what the event is documented to mean ("on kernel start, restart, or shutdown").

Moved the event into `RestartNotebookTool`, where the outcome is known: it fires right after `restart_kernel` returns, after a reprovision succeeds (under the new kernel id), and when the sandbox restart went through. Nothing fires on a failure path. Same placement `execute_cell` / `execute_code` already use for `BEFORE_EXECUTE` / `AFTER_EXECUTE`, and the id reported is the one the server used to read back anyway.

Tests cover both modes and the reprovision path, plus one through the registered tool so a failed restart leaves no event and a successful one leaves exactly one — that last one fails on `main`.

`use_notebook` / `unuse_notebook` fire `started` / `stopped` the same unconditional way. Left them alone here: what counts as a start on a `connect`, or a stop when the kernel keeps running, is a question of its own.


echarles (approved):
LGTM Thx @serhiizghama

PR #460, opened 2026-09-16
fix(list_files): make the pattern filter match the way glob does

### LINKED ISSUE #459: list_files with pattern='**/*.ipynb' skips the notebooks in the directory it was given
`list_files(path="", max_depth=1, pattern="**/*.ipynb")` gives back the notebooks in subdirectories and leaves out the ones sitting in the directory itself.

The filter is `fnmatch.fnmatch(f["path"], pattern)` at list_files_tool.py:225. fnmatch has no idea what a `/` is, so `**` is only "any characters" twice and the `/` after it is a literal one the path has to supply. `Untitled.ipynb` has no slash in it, so it never matches. The same blind spot runs the other way for `*`: `*.py` matches `work/helper.py`, which glob would not.

Repro at 7465525 in python:3.11-slim, through `ListFilesTool.execute` against a `LargeFileManager` over a temp root holding `Untitled.ipynb`, `top.py`, `work/analysis.ipynb` and `work/helper.py`:

```
pattern='**/*.ipynb'  ->  ['work/analysis.ipynb']
pattern='*.py'        ->  ['top.py', 'work/helper.py']
```

`glob.glob('**/*.ipynb', recursive=True)` over the same tree returns both notebooks, so the expected first line is `['Untitled.ipynb', 'work/analysis.ipynb']` and the expected second is `['top.py']`.

`'*.py', '**/*.ipynb'` are the examples in the tool's own docstring (list_files_tool.py:196), so those are the patterns a client is going to send. Sending a PR for it now.

ebarkhordar:
Fixes #459.

Replaced the `fnmatch` call with a small glob-to-regex translation so the pattern knows where the directory separators are: `*` and `?` stop at a `/`, `**/` covers zero or more leading directories, `**` takes the rest of the path.

One behaviour change beyond the reported bug, worth a look before merging. `*.py` used to match `work/helper.py` and now matches only `top.py`. That is what glob means by `*` and it is the other half of the same fnmatch blind spot, but it is still a change for anyone leaning on the loose form. If you want `*` to keep crossing `/`, that is one line in `compile_glob`.

A bad pattern behaves as before: `[z-a]` raises out of `re.compile` the way it did out of fnmatch, so it lands in the existing `except` and prints the same warning.

Added tests for the two documented examples, a literal directory prefix, and the no-match message. The first two fail on main.


echarles (approved):
LGTM Thx @ebarkhordar

PR #461, opened 2026-09-16
fix(capabilities): ask each extension once, at startup

ebarkhordar:
`collect_capabilities()` is called from one place, `capabilities_resource()`, and `declare()` re-applies the extension's own `enabled` and `source` to a name that already exists. So a read of `capabilities://` reconfigures the server: an operator turns a plugin capability off, the next client looks at the list, and it is back on with the plugin's source.

The same lazy collection means a plugin's capability is not in the registry at all until somebody reads that resource, so a tool consulting `enabled(name)` on the request path gets False until then. `register_extension_tools()` now collects as well, which is where the rest of an extension's registration already happens.

Asking each extension once is what stops the clobber. An extension registered after the first collection is still asked, there is a test for that.

Left alone on purpose: `--capability plugin.name` at startup still raises `UnknownCapability`, since serve.py applies the flag before `do_start` registers anything. Fixing that means deciding whether an operator's flag outranks a plugin's own declaration, and that is your call rather than mine.

This needs a plugin that declares a capability and nothing in the tree does yet, so no shipped deployment is affected. The tests drive a plugin double through the documented `capabilities()` API. The clobber one fails on main.


echarles (approved):
LGTM Thx @ebarkhordar

PR #465, opened 2026-09-17
fix(healthz): report the installed package version

serhiizghama:
`GET /mcp/healthz` has been answering `"version": "0.20.0"` since #170 — the string is hard-coded in `MCPHealthHandler`, while the package is at 2.1.16 and the MCP `initialize` handshake a few hundred lines up in the same file already reports `serverInfo.version` from `__version__`. So the two ways of asking the extension which version it is disagree, and the health check is the one that's wrong. Switched it to `__version__` (already imported there) and added a check next to `test_extension_health` that pins the healthz version to the package version; it fails on `main` with `'0.20.0' == '2.1.16'`. Same family as the mcpb drift #236 fixed, just on the runtime side.


echarles:
Strange, all tests failing the same way although this PR does not touch the failing parts.

```
FAILED tests/test_notebook_update_notifications.py::TestThePageSaysWhatHappens::test_it_still_says_another_party_is_not_covered - assert 'A change by somebody else is not covered' in '# Resources\n\nTools **push**. Everything a call produces comes back whether the agent wanted\nit or not, so a cell that printed a megabyte spends a megabyte of the client\'s\ncontext every time anybody reads it.\n\nResources **pull**. The agent is told what exists and reads the one thing it\nneeds.\n\nThat is what these three are for.\n\n| Resource | What it is |\n|---|---|\n| `notebook://{name}` | A notebook in use, as nbformat JSON |\n| `notebook://{name}/cells/{cell_id}` | One cell, with its outputs *listed* |\n| `notebook://{name}/cells/{cell_id}/outputs/{index}` | One output, on its own |\n\n## Addressed by name and cell id\n\nA notebook path contains slashes. It cannot sit in one URI template segment\nwithout being encoded into something nobody can read, and the **name** is what\nevery tool here already takes — the one `use_notebook` registered.\n\nA cell is addressed by its nbformat 4.5 **id**, not its index. An index is a\nposition in a document somebody else is editing: between reading a notebook\nand reading "cell 4", cell 4 may be a different cell. The id is not.\n\n:::note The scheme is provider-neutral\n\nThis server talks to a Jupyter server. A `datalayer://` URI ...e.\n\n```python\nclass MyPublisher:\n    async def publish(self, uri: str) -> bool:\n        ...\n```\n\n### Why a gate\n\nA deployment that addresses notebooks by its own identifiers serves its own\nresources and does not want these. `notebook://{name}` is the name a *worker*\nknows, and such a deployment\'s clients have never seen one.\n\n```python\nclass MyGate:\n    def serves(self, uri_template: str) -> bool:\n        ...\n```\n\nA withheld resource raises `ResourceWithheld`, not `ResourceNotFound`. *We do\nnot answer this here* and *there is no such cell* are different answers, and a\nclient told the second when the first is true goes looking for a cell it will\nnever find.\n\n:::note A misconfigured seam is fatal; a broken one is not\n\nA class that cannot be imported or built stops the server, for the reason the\naudit sink gives: a deployment that configured delivery and got a server\nrunning without it believes subscribers are being told, and they are not.\n\nA gate that *raises when asked* serves the resource instead. That is the safe\ndirection for a scaffold — the alternative is a server that silently answers\nnothing and looks like a server with no resources.\n:::\n'
 +  where '# Resources\n\nTools **push**. Everything a call produces comes back whether the agent wanted\nit or not, so a cell that printed a megabyte spends a megabyte of the client\'s\ncontext every time anybody reads it.\n\nResources **pull**. The agent is told what exists and reads the one thing it\nneeds.\n\nThat is what these three are for.\n\n| Resource | What it is |\n|---|---|\n| `notebook://{name}` | A notebook in use, as nbformat JSON |\n| `notebook://{name}/cells/{cell_id}` | One cell, with its outputs *listed* |\n| `notebook://{name}/cells/{cell_id}/outputs/{index}` | One output, on its own |\n\n## Addressed by name and cell id\n\nA notebook path contains slashes. It cannot sit in one URI template segment\nwithout being encoded into something nobody can read, and the **name** is what\nevery tool here already takes — the one `use_notebook` registered.\n\nA cell is addressed by its nbformat 4.5 **id**, not its index. An index is a\nposition in a document somebody else is editing: between reading a notebook\nand reading "cell 4", cell 4 may be a different cell. The id is not.\n\n:::note The scheme is provider-neutral\n\nThis server talks to a Jupyter server. A `datalayer://` URI ...e.\n\n```python\nclass MyPublisher:\n    async def publish(self, uri: str) -> bool:\n        ...\n```\n\n### Why a gate\n\nA deployment that addresses notebooks by its own identifiers serves its own\nresources and does not want these. `notebook://{name}` is the name a *worker*\nknows, and such a deployment\'s clients have never seen one.\n\n```python\nclass MyGate:\n    def serves(self, uri_template: str) -> bool:\n        ...\n```\n\nA withheld resource raises `ResourceWithheld`, not `ResourceNotFound`. *We do\nnot answer this here* and *there is no such cell* are different answers, and a\nclient told the second when the first is true goes looking for a cell it will\nnever find.\n\n:::note A misconfigured seam is fatal; a broken one is not\n\nA class that cannot be imported or built stops the server, for the reason the\naudit sink gives: a deployment that configured delivery and got a server\nrunning without it believes subscribers are being told, and they are not.\n\nA gate that *raises when asked* serves the resource instead. That is the safe\ndirection for a scaffold — the alternative is a server that silently answers\nnothing and looks like a server with no resources.\n:::\n' = <function TestThePageSaysWhatHappens._page at 0x10ca45760>()
```

serhiizghama:
It's not this branch — CI runs on the merge ref, so it picked up 2e42e5a from main. `test_it_still_says_another_party_is_not_covered` pins the sentence "A change by somebody else is not covered" in `docs/docs/architecture/resources/index.mdx`, and that's exactly the paragraph 2e42e5a rewrote into "A change by somebody else *is* covered, by a watcher". The tripwire fired the way its docstring says it should. main's own Test run on that commit is red the same way: https://github.com/datalayer/jupyter-mcp-server/actions/runs/35182964987 (and it fails locally on a clean checkout of main, 1 failed / 31 passed in that file).

Once the test is flipped or dropped I'll rebase — or I can fold that one line in here if you'd rather not do a separate commit.

serhiizghama:
Merged main now that the tripwire test was updated in #466 — CI should be green again.

echarles (approved):
LGTM Thx @serhiizghama

